PR #2, opened 2019-03-15
Initial readme

claudiulodro:
An initial readme for the theme repo so visitors have useful info. Most of this is [the same as the plugin repo readme](https://github.com/Automattic/newspack-plugin), and can be changed to be theme-repo-specific if needed.


laurelfulford:
LGTM! Thanks @claudiulodro!

PR #10, opened 2019-03-25
Remove non-latin font fallbacks

### LINKED ISSUE #9: Remove font fallbacks for different languages
Twenty Nineteen includes font fallbacks for a number of non-latin languages, which is important for a default theme.

For our purposes, however, it's not necessary to cover all of those possibilities in the theme's default settings. These font fallbacks should be removed from the theme.

laurelfulford:
### All Submissions:

This PR removes the non-latin font fallbacks from the theme. Though they make sense in a default theme, like Twenty Nineteen, Newspack's own customization options should make these fallbacks unnecessary. It also shaves 2000+ lines of CSS from the theme.

Closes #9.

1. Apply the PR. 
2. Run the build script (`npm run build`) to make sure the CSS compiles without errors.
3. View the front-end and editor, to confirm the typography is still styled correctly.

### Other information:


laurelfulford:
Good question! We'll definitely need a way to allow for non-latin fonts. I'm hoping we can get around using _as_ much CSS in a couple ways: 

* We're looking at including the ability to customize fonts, which should make it possible to switch to different faces that support different non-latin characters on a site-by-site basis.
* We can make the default simpler, by using system fonts that support more non-latin languages, and even by just using one font for everything -- Twenty Nineteen uses two throughout the theme, so the overrides have to be a lot more specific than in a theme like Twenty Seventeen, which only uses one. So, if we do need to re-add something like this back, it won't need to be as much CSS to do it.

claudiulodro:
Sounds good. Thanks for clarifying.

PR #13, opened 2019-03-29
Add option to hide the page title on the static front page

### LINKED ISSUE #12: Add ability to hide front page title
Depending on the layout, you may not always want to display the page title on the static front page.

Adding an option to the Customizer should make doing this easier, without having to actually remove a page's title, or remove the output fully from the page template.

laurelfulford:
### All Submissions:

This PR adds an option to the Customizer to hide the page title on the static front page. 

It may not always be needed in layouts; this makes sure it can be hidden without a work-around (like leaving the 'title' field empty on the page). 

Closes #12.

1. Apply the patch. 
2. Navigate to Customize > Homepage Settings.
3. If it's not already, switch the test site to use a static front page.
4. Check the 'Hide Homepage Title' checkbox.
5. Make sure the preview works as expected, and the title hides and re-shows.
6. Make sure the changes publish correctly.
7. Make sure the setting doesn't affect other pages.
8. Make sure the setting doesn't affect the blog posts page when set to the front page. 

### Other information:

### DIFF
--- a/inc/customizer.php
+++ b/inc/customizer.php
@@ -80,6 +80,29 @@ function newspack_customize_register( $wp_customize ) {
 			)
 		)
 	);
+
+	// Add option to hide page title on static front page.
+	$wp_customize->add_setting(
+		'hide_front_page_title',
+		array(
+			'default'           => false,
+			'type'              => 'theme_mod',
+			'transport'         => 'postMessage',
+			'sanitize_callback' => 'newspack_sanitize_checkbox',
+		)
+	);
+
+	$wp_customize->add_control(
+		'hide_front_page_title',
+		array(
+			'label'       => esc_html__( 'Hide Homepage Title', 'newspack' ),
+			'description' => esc_html__( 'Check to hide the page title, if your homepage is set to display a static page.', 'newspack' ),
+			'section'     => 'static_front_page',
+			'priority'    => 10,
+			'type'        => 'checkbox',
+			'settings'    => 'hide_front_page_title',
+		)
+	);
 }
 add_action( 'customize_register', 'newspack_customize_register' );
 
@@ -142,4 +165,19 @@ function newspack_sanitize_color_option( $choice ) {
 	}
 
 	return 'default';
+}
+
+/**
+ * Sanitize the checkbox.
+ *
+ * @param boolean $input Value of checkbox.
+ *
+ * @return boolean true if is 1 or '1', false if anything else
+ */
+function newspack_sanitize_checkbox( $input ) {
+	if ( 1 == $input ) {
+		return true;
+	} else {
+		return false;
+	}
 }
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -26,6 +26,11 @@ function newspack_body_classes( $classes ) {
 		$classes[] = 'newspack-customizer';
 	endif;
 
+	$hide_title = get_theme_mod( 'hide_front_page_title', false );
+	if ( true === $hide_title ) {
+		$classes[] = 'hide-homepage-title';
+	}
+
 	return $classes;
 }
 add_filter( 'body_class', 'newspack_body_classes' );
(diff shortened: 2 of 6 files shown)


inc/customizer.php
```diff
@@ -143,3 +166,18 @@ function newspack_sanitize_color_option( $choice ) {
...
 }
+
+/**
+ * Sanitize the checkbox.
+ *
+ * @param boolean $input Value of checkbox.
+ *
+ * @return boolean true if is 1 or '1', false if anything else
+ */
+function newspack_sanitize_checkbox( $input ) {
```
jeffersonrabb: This looks a little suspicious to me. When I tested, I was getting a value of `true|false` for `$input`, in which case this function isn't necessary at all. I'm curious why we need?

inc/template-functions.php
```diff
@@ -26,6 +26,11 @@ function newspack_body_classes( $classes ) {
 		$classes[] = 'newspack-customizer';
 	endif;
 
+	$hide_title = get_theme_mod( 'hide_front_page_title', false );
+	if ( true === $hide_title ) {
+		$classes[] = 'hide-homepage-title';
```
jeffersonrabb: Legitimate theming best-practices question: should we prefix the classes `newspack_`?

sass/site/primary/_posts-and-pages.scss
```diff
@@ -222,6 +222,14 @@
 	}
 }
 
+/* Hide page title on the homepage */
+
+.page.home.hide-homepage-title {
+	.entry-header {
+		display: none;
```
jeffersonrabb: From a screen reader perspective is there any danger in hiding this, as opposed to taking approaches laid out in https://css-tricks.com/places-its-tempting-to-use-display-none-but-dont/?

inc/template-functions.php
laurelfulford: It's a good question -- it's something that comes up a lot! :) The theme mods are automatically stored as `theme_mods_[themeslug]`, so it's not required to prefix the classes in theming best practices (going off of WP.org theme requirements, and default themes). The one super-odd case where you could have a conflict is if there's a plugin installed that's storing something as a `theme_mod` for some reason, and it has the same name.

inc/customizer.php
laurelfulford: It's a best practice to sanitize all the things in the Customizer, but yeah, it can definitely look funny checking against two possible values. What it's suppose to prevent is any kind of bad actors somehow injecting other values.

It looks like you can also use an existing function, `sanitize_text_field()`, for sanitizing checkboxes and radio buttons ([link](https://core.trac.wordpress.org/ticket/41810#comment:9) - TIL!) so switching to that could get rid of an extra function, and remove some confusion.

sass/site/primary/_posts-and-pages.scss
laurelfulford: In most cases, yeah, that's definitely the better approach! 

In this case, I'm using it as a touch-less-files way to removing the element entirely. But I can swap it to check the `theme_mod` before writing the page title to the page instead, so the element just isn't there -- it would make what's happening less confusing (the theme wouldn't re-show the title unless the option is changed, so it's not like it needs to be hiding there, waiting).

inc/template-functions.php
laurelfulford: OH I just realized what this comment was about _facepalm_. It's a CSS class, definitely no need to add `newpack_` in front.

sass/site/primary/_posts-and-pages.scss
jeffersonrabb: Thinking on this a bit more, I'm sure this is fine. The page already has a title which screen readers will pick up, so this is neither unique nor critical information.

inc/customizer.php
jeffersonrabb: Looking closer I think you have this correct. It was the user of the Equal operator (`==`) rather than Identical (`===`) that threw me, but in the case of checkbox sanitization we do want anything that isn't obviously false, so looks like the right logic.

laurelfulford:
Thanks @jeffersonrabb!

PR #15, opened 2019-04-23
Correct `$content_width` value for the theme

laurelfulford:
### All Submissions:

This pull request just sets a higher `$content_width` value for the theme, which is used in embeds and for some content image sizing. 

While working on blocks, I noticed the current value was causing the default 'large' image sizes not to be available.

This might not be the final width (it will depend on how wide the content area of the theme ends up being) but it will make it easier to work with in the short-term. 

As an aside, the `$content_width` is coming from Twenty Nineteen, where [it's also incorrect based on the theme's actual width](https://core.trac.wordpress.org/ticket/46262).

1. Apply the patch.
2. Make sure the 'Large' image size is at least 960px wide (you can check under Settings > Media).
3. Add an image that's at least 960px wide to the editor.
4. Publish the post.
5. Make sure the image being loaded is 960px wide in the editor and on the front-end (as opposed to 640px, the original `$content_width` size).

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -213,7 +213,7 @@ function newspack_content_width() {
 	// This variable is intended to be overruled from themes.
 	// Open WPCS issue: {@link https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1043}.
 	// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedVariableFound
-	$GLOBALS['content_width'] = apply_filters( 'newspack_content_width', 640 );
+	$GLOBALS['content_width'] = apply_filters( 'newspack_content_width', 960 );
 }
 add_action( 'after_setup_theme', 'newspack_content_width', 0 );
 


jeffersonrabb:
I think I may be testing this incorrectly. Here's what I tried:

- Branch switch to `master`
- Go to Settings->Media, and set "Large Size" Max Width to 960. I also set Max Height to 960 just to be safe.
- Create new post, add Image block, pick large image from Media Library (the raw image is 2000x1125)
- Initially I have resize handles on the image, so I change the block alignment to Wide.
- In sidebar set Image Size to Large (it's the default). 
- At this point, the image in the editor is 400x225
- Branch switch to `fixes/content-width-value`
- Repeat all the above steps

At this point, the image in the editor is still 400x225, although I expected it to be larger. Am I doing something wrong?

laurelfulford:
Nope, my steps were terrible -- sorry about that! I was testing with a block I was working on, forgetting that it works a bit differently with images inserted in content... and that Gutenberg image blocks don't use `$content_width` yet 🤦‍♀️

Here are revised -- and double-checked! -- testing steps: 

1. Make sure your 'Large' image size is at least 960px wide (under Settings > Media). 
1. In a classic block, add an image. Set the image to the size 'Large'. The pixel size next to the label in the size dropdown should say 640 wide.
2. Save and publish; observe the size of the image. It should be displaying at 640px wide.
3. Apply the patch.
4. Add the another image, with the same settings. The pixel size next to the label in the size dropdown should now say 960 wide.
5. Save and publish. 
6. The image will display at narrower than 960px (the theme's content area will also need adjusting), but it will fill the content area; the width attribute will also say 960px. 

I think this should work, but just let me know if you run into any issues with the above!

laurelfulford:
Thanks @jeffersonrabb!

PR #18, opened 2019-05-23
Feature/amp native

philipjohn:
### All Submissions:

Makes the minor changes necessary to make the theme AMP Native

1. Apply the patch
2. Activate the AMP plugin and choose Native mode
3. Observe that the theme works as expected
4. Observe that the AMP validation tool shows all pages as valid AMP

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -262,6 +262,24 @@ function newspack_scripts() {
 add_action( 'wp_enqueue_scripts', 'newspack_scripts' );
 
 /**
+ * Fix skip link focus in IE11.
+ *
+ * This does not enqueue the script because it is tiny and because it is only for IE11,
+ * thus it does not warrant having an entire dedicated blocking script being loaded.
+ *
+ * @link https://git.io/vWdr2
+ */
+function newspack_skip_link_focus_fix() {
+	// The following is minified via `terser --compress --mangle -- js/skip-link-focus-fix.js`.
+	?>
+	<script>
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -238,24 +241,6 @@ function newspack_scripts() {
...
 
-/**
- * Fix skip link focus in IE11.
- *
- * This does not enqueue the script because it is tiny and because it is only for IE11,
- * thus it does not warrant having an entire dedicated blocking script being loaded.
- *
- * @link https://git.io/vWdr2
- */
-function newspack_skip_link_focus_fix() {
```
laurelfulford: This script is indented to fix a keyboard navigation issue in IE11, where the visual focus will move to the content, but the input focus won't (so if you hit enter on 'Skip to Content' then tab again, rather than passing you on to the next link, it kicks you back up the page to the 'Skip to Content' link). 

Assuming we'll be supporting IE11, could we switch it to a standard enqueue instead and still be AMP-native?

laurelfulford: Looking at some AMP work the theme team was testing out with Twenty Nineteen, this could be another approach: adding a check to the theme for an amp endpoint:

https://github.com/Automattic/twenty-nineteen-amp-prototype/blob/fe8c1cd6665f2b4701fb44e1cce59bed5810d740/functions.php#L196-L205

... and that wrapping that script -- and others -- in it:

https://github.com/Automattic/twenty-nineteen-amp-prototype/blob/fe8c1cd6665f2b4701fb44e1cce59bed5810d740/functions.php#L302-L312

(Also just let me know if I'm missing something silly here AMP-wise -- I wouldn't rule it out! :) I'm mostly thinking in terms of making sure things still work, with or without AMP).

philipjohn: It's already handled by AMP :) https://github.com/ampproject/amphtml/issues/18671

laurelfulford: Right! I was thinking approaching this to a similar way as https://github.com/Automattic/newspack-blocks/pull/20 -- so the theme will still load this JavaScript in cases where it's being used without AMP :)

laurelfulford:
🚢 Looks good to me!

PR #26, opened 2019-05-28
Preliminary work to update the theme's width

laurelfulford:
### All Submissions:

This PR starts the process of fixing the theme's width to better match the in-progress mockups. It sets the header and footer width to 1200px, as well as the front page when set to a static page. It sets the content area of single posts to approximately 780px, and slightly offset. 

There's a number of things that still need to be done, like adding a sidebar, and deciding how to handle centring the content, and in what cases (like will it be always centred when there's no sidebar, or just for special features?). The editor styles will also need to be updated to match; these will all be handled in separate PRs.

1. Apply the PR.
2. Make sure the header and footer are 1200px wide and centred.
3. Make sure the content of single posts and pages is ~780px wide, and set to the left. 
4. Set a static front page; make sure the content area is 1200px in that case. 

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -5,12 +5,6 @@
 	margin: 32px 0;
 	max-width: 100%;
 
-	@include postContentMaxWidth();
-
-	@include media(tablet) {
-		margin: 32px auto;
-	}
-
 	> *:first-child {
 		margin-top: 0;
 	}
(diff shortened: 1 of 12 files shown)


sass/blocks/_blocks.scss
```diff
@@ -20,29 +14,23 @@
...
 	&.alignwide {
-		margin-left: auto;
-		margin-right: auto;
-		clear: both;
-
 		@include media(tablet) {
-			width: 100%;
-			max-width: 100%;
+			margin-right: -53%;
+			max-width: 1000%;
```
claudiulodro: 1000%? Is that correct?

It doesn't seem to cause any issues, though because the style gets overrode by a more specific one.

laurelfulford: Good question! It's correct in that it's meant to allow those blocks to be wider than their container, but it's kind of hacky :) In 0a402a2, I've swapped it for `100vw`, which does the same thing, but also a) makes sure it doesn't exceed the width of the browser window, and b) looks less weird.

Issue #20, opened 2019-05-24
Build out theme header


laurelfulford:
Though the eventual plan for the theme is to have a header that can be customized, let's start by building out a basic fallback that allows a few different looks, just by how many menus you assign.

Here's a rough placement for the different elements (though the actual appearance will also depend on the current style). 

![header-wireframe](https://user-images.githubusercontent.com/177561/58359506-aa2f3680-7e38-11e9-81db-778b0fc0237c.png)

laurelfulford:
Closing as done; also in favour of further refinement in #96.

PR #28, opened 2019-05-29
Add 'secondary' menu space to the header

laurelfulford:
### All Submissions:

This PR adds the secondary menu space outlined in #20. 

It adds a one-level menu that will sit to the left of the social menu when added. 

It does not:

1. Do anything 'special' on mobile -- the links currently all remain visible, though around the tablet breakpoint this menu and the social menu will become stacked. 
2. Add more than one level to the menu.

Mobile improvements will definitely be handled in a future PR; adding the ability to have more than one level of menu can also be added pretty painlessly, if it's deemed necessary. Either of these should use the styles and functionality already built out for the primary menu, to reduce duplication of code, so it will need to be reworked a bit. 

1. Apply the PR.
2. Add a menu to the 'Secondary Menu' space. 
3. Test with and without a social menu, and with a short and longer menu, to make sure there is no major breakage. 

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -54,10 +54,10 @@ if ( ! function_exists( 'newspack_setup' ) ) :
 		// This theme uses wp_nav_menu() in two locations.
 		register_nav_menus(
 			array(
-				'menu-1' => __( 'Primary Menu', 'newspack' ),
-				'menu-2' => __( 'Secondary Menu', 'newspack' ),
-				'footer' => __( 'Footer Menu', 'newspack' ),
-				'social' => __( 'Social Links Menu', 'newspack' ),
+				'primary-menu'   => __( 'Primary Menu', 'newspack' ),
+				'secondary-menu' => __( 'Secondary Menu', 'newspack' ),
+				'footer'         => __( 'Footer Menu', 'newspack' ),
+				'social'         => __( 'Social Links Menu', 'newspack' ),
 			)
 		);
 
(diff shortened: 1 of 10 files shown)


laurelfulford:
I should also note that this PR fixes an issue where the menus and site title don't sit flush on the left side -- this came out of committing a few parallel PRs that didn't end up playing well together once they were all in `master`:

image

functions.php
```diff
@@ -54,7 +54,8 @@ function newspack_setup() {
 		// This theme uses wp_nav_menu() in two locations.
 		register_nav_menus(
 			array(
-				'menu-1' => __( 'Primary', 'newspack' ),
+				'menu-1' => __( 'Primary Menu', 'newspack' ),
```
philipjohn: Nitpick: Descriptive names are always better IMO, and we could match the human readable name here.

laurelfulford: Ah, that's a good call, especially as we add more menu items! 

The `menu-1`, `menu-2`... thing is kind of a throwback from when I was on the theme team. By giving menus and widgets consistent IDs on WordPress.com, we could help make sure they weren't unassigned when someone switched themes. This is less of a concern in this theme -- folks are less likely to be cycling through multiple themes to check them out -- and it could help us keep the specific locations straight. 

I've renamed these two spaces to `primary-menu` and `secondary-menu` in 6831ce6 , to match their visible names.

Issue #24, opened 2019-05-27
Custom Colours: Add secondary colour option


laurelfulford:
The theme should have at least one more custom colour option in the customizer.

This colour should behave the same as the existing colour: it should show up in the editor, and a variation with a darker shade should be automatically generated. 

This colour should also be applied to some existing elements (links, menu, or buttons).

laurelfulford:
The initial work for this has been committed in #29.

PR #29, opened 2019-05-29
Add second custom color option

laurelfulford:
### All Submissions:

The secondary colour has only been added to the content links so far (which alone, needed a lot of variables to be changed). We still need to finalize exactly where it should be applied otherwise in the theme, and added there in a future PR. 

This PR also:

* Cleans up comments in color-patterns.php, to reference the colour being applied rather than what it's being applied to. 
* Removes theme's Dark Grey and Light Grey from the editor colour palette; we will want to have some default colours that are always there, but we'll need to figure out what they should be once the default colour palettes are finalized. As is, with a default secondary grey colour, you end up with a lot of different greys. 
* Removes the `postMessage` transport from the Customizer colour options, and the JavaScript preview. Unfortunately, incorporating multiple colours makes previewing like this more complex -- it's something to revisit in the future, once the colour behaviour is finalized, but is not a breaking issue (the Customizer will just do a full refresh when the colour is changed, rather than previewing without). 

See #24.

1. Apply the PR.
2. Confirm that there is a mid-range and dark grey added to the editor as colour options.
3. In the Customizer, confirm that there is an option for a second custom colour. 
4. Change the colour; confirm it changes links in the content body, and the colour options in the editor. 

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -137,7 +137,7 @@ if ( ! function_exists( 'newspack_setup' ) ) :
 			)
 		);
 
-		$primary_color = newspack_get_primary_color();
+		$primary_color   = newspack_get_primary_color();
 		$secondary_color = newspack_get_secondary_color();
 
 		// Editor color palette.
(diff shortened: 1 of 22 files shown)


functions.php
```diff
@@ -154,14 +155,14 @@ function newspack_setup() {
 					'color' => 'default' === get_theme_mod( 'theme_colors' ) ? newspack_adjust_brightness( $primary_color, -40 ) : newspack_adjust_brightness( get_theme_mod( 'primary_color_hex', $primary_color ), -40 ),
 				),
 				array(
-					'name'  => __( 'Dark Gray', 'newspack' ),
-					'slug'  => 'dark-gray',
-					'color' => '#111',
+					'name'  => __( 'Secondary', 'newspack' ),
```
philipjohn: Minor: It's always safer to use `esc_html__()` and adjust later if required for translations with special characters.

functions.php
```diff
@@ -154,14 +155,14 @@ function newspack_setup() {
...
+					'slug'  => 'secondary',
+					'color' => 'default' === get_theme_mod( 'theme_colors' ) ? $secondary_color : get_theme_mod( 'secondary_color_hex', $secondary_color ),
 				),
 				array(
-					'name'  => __( 'Light Gray', 'newspack' ),
-					'slug'  => 'light-gray',
-					'color' => '#767676',
+					'name'  => __( 'Secondary Variation', 'newspack' ),
```
philipjohn: Minor: It's always safer to use `esc_html__()` and adjust later if required for translations with special characters.

functions.php
```diff
@@ -154,14 +155,14 @@ function newspack_setup() {
 					'color' => 'default' === get_theme_mod( 'theme_colors' ) ? newspack_adjust_brightness( $primary_color, -40 ) : newspack_adjust_brightness( get_theme_mod( 'primary_color_hex', $primary_color ), -40 ),
 				),
 				array(
-					'name'  => __( 'Dark Gray', 'newspack' ),
-					'slug'  => 'dark-gray',
-					'color' => '#111',
+					'name'  => __( 'Secondary', 'newspack' ),
+					'slug'  => 'secondary',
+					'color' => 'default' === get_theme_mod( 'theme_colors' ) ? $secondary_color : get_theme_mod( 'secondary_color_hex', $secondary_color ),
```
philipjohn: Nitpick: It'd be more readable to have this broken up across multiple lines.

inc/color-filters.php
```diff
@@ -21,10 +22,30 @@ function newspack_get_primary_color() {
 }
 
 /**
- * Tests the current default hue against NEWSPACK_DEFAULT_HEX.
+ * The default color used for the secondary color throughout this theme
+ *
+ * @return string the default hexidecimal color.
+ */
+function newspack_get_secondary_color() {
+	return apply_filters( 'newspack_secondary_color', NEWSPACK_DEFAULT_SECONDARY );
```
philipjohn: We should document this hook according to the [inline docs standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters).

functions.php
laurelfulford: Ah, good call! This is throughout the theme -- strings marked for translation aren't escaped in default themes, but it is standard practice on WP.com. I've created a separate issue to tackle this (#30), and am going to mark this comment as resolved.

functions.php
laurelfulford: Same as the previous comment -- let's fix this up throughout in #30!

functions.php
laurelfulford: Good call! Tidied up in f484753.

inc/color-filters.php
laurelfulford: Good catch - I've added comment for both this new colour filter and the one that was already there.

PR #33, opened 2019-05-30
Update the theme font faces and sizes

laurelfulford:
### All Submissions:

This looks like a lot, but it's just a couple font updates to the theme:

* Swapping out Bakerville for Georgia in the theme's body font. 
* Adjusting the font sizes throughout -- reduce the size of the base and the headers, and add an additional larger font size for post titles.
* Add font family to Newspack block article meta (the class will probably need to be updated if/when the block is broken into multiple blocks.

Probably the quickest thing to check would be:

1. Apply the PR.
2. Confirm that the body font has changed to Georgia.
3. Confirm that the headers and body is a bit smaller (the quickest thing to check is the widget title size, since it's pretty noticeable). 

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -798,6 +798,14 @@
 		}
 	}
 
+	//! Newspack Block
+
+	.wp-block-newspack-blocks-homepage-articles {
+		.article-meta {
+			font-family: $font__heading;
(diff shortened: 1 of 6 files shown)


laurelfulford:
Great - thanks for the reviews! 🎉

PR #34, opened 2019-05-30
Update how the primary content width is handled

laurelfulford:
### All Submissions:

This is some additional reworking of the theme's widths. It shouldn't actually _look_ different, but it should make adding a sidebar easier than the original styles. Instead of setting widths on individual items, like the post title, content, meta and previous/next navigation, the width has been moved to the containing element. 

It also has the added benefit of removing a bunch of CSS (that I'm sure will be re-added with the sidebar!). 

1. Apply the PR.
2. Not to oversimplify things, but give things a quick once-over and make sure they don't really look any different :) 

### Other information:


laurelfulford:
Thanks for the review, @claudiulodro! I did a pretty poor job explaining these changes -- both of these things are intentional, at least for now:

For the homepage, it's wider than the rest of the pages, giving you 1200px to lay out blocks in different columns. It should have been this way in an earlier commit, but I left some styles on the `.aligncenter` class that were causing the blocks to still display narrowly (that class is always being applied to the blocks for some reason I haven't figure out yet -- possibly because it doesn't support alignments? I filed an issue for that here: https://github.com/Automattic/newspack-blocks/issues/26). 

The image sizing issue should be fixed when https://github.com/Automattic/newspack-blocks/issues/15  is fixed-- the problem is that the block right now is using the 'large' thumbnail size, which by default is only 1024 wide. Increasing that in the media library settings will temporarily fix that for now, but the block should really set its own custom image sizes.

For the single posts, the wide and full blocks do ooze all over the place at the moment -- my short-term plan for once the sidebar is in place is to include a body class when the sidebar is in place, and override those styles so everything fits in the content area. Ideally, we could do something like use JavaScript to see when we're "past" the sidebar, to let wide blocks be wide when they won't run into it, but that'll probably be down the road :) 

Anyway -- all that to say, I'm going to merge this for now :) thank you for checking it out!

PR #35, opened 2019-05-30
Update article post meta

laurelfulford:
### All Submissions:

This PR tidies up the post meta for each post:

1. It makes it consistent on the single posts and blog posts page (displays the author and date information at the top, and the category, tag and edit link at the bottom).
2. Removes the comments link from the top for now.
3. Adds the author avatar, and aligns the post author's name and date next to it. 
4. Removes the icons from the post author and date. 
5. Updates the post header template part to be used on single and archive pages. 
6. Removes the 'Featured' text from sticky posts, and the associated styles.

1. Apply the PR.
2. Confirm that only the author avatar, name and date appear at the top of posts, without icons. The author name and date should sit next to the author avatar.
3. Confirm that the categories, tags and edit link appear at the bottom of posts.
4. Create a sticky post (or check an existing sticky post) on the block posts page and confirm that it does not say 'Featured' at the top. 

### Other information:


inc/template-tags.php
```diff
@@ -24,8 +24,7 @@ function newspack_posted_on() {
 		);
 
 		printf(
-			'<span class="posted-on">%1$s<a href="%2$s" rel="bookmark">%3$s</a></span>',
-			newspack_get_icon_svg( 'watch', 16 ),
+			'<span class="posted-on"><a href="%1$s" rel="bookmark">%2$s</a></span>',
 			esc_url( get_permalink() ),
 			$time_string
```
philipjohn: All output should be secured by appropriate escaping [as late as possible](https://vip.wordpress.com/documentation/best-practices/security/validating-sanitizing-escaping/#always-escape-late), even if we can see where the string is created in the code, to protect against future changes.

inc/template-tags.php
```diff
@@ -38,10 +37,10 @@ function newspack_posted_on() {
 	 */
 	function newspack_posted_by() {
 		printf(
-			/* translators: 1: SVG icon. 2: post author, only visible to screen readers. 3: author link. */
-			'<span class="byline">%1$s<span class="screen-reader-text">%2$s</span><span class="author vcard"><a class="url fn n" href="%3$s">%4$s</a></span></span>',
-			newspack_get_icon_svg( 'person', 16 ),
-			__( 'Posted by', 'newspack' ),
+			/* translators: 1: Author avatar. 2: post author, only visible to screen readers. 3: author link. */
+			'<span class="author-avatar">%1$s</span><span class="byline"><span>%2$s</span> <span class="author vcard"><a class="url fn n" href="%3$s">%4$s</a></span></span>',
+			get_avatar( get_the_author_meta( 'ID' ) ),
```
philipjohn: We shouldn't trust that core is going to provide safe output, so we should escape this string before printing it out. `wp_kses_post()` should be fine, given we expect an `img` tag here.

template-parts/header/entry-header.php
```diff
@@ -7,33 +7,18 @@
 
 $discussion = ! is_page() && newspack_can_show_post_thumbnail() ? newspack_get_discussion_data() : null; ?>
 
-<?php the_title( '<h1 class="entry-title">', '</h1>' ); ?>
+
+<?php
+if ( is_singular() ) :
+	the_title( '<h1 class="entry-title">', '</h1>' );
```
philipjohn: My recommendation here would be to use `get_the_title()` inside `echo wp_kses_post()` to protect against the possibility of bad input in the title field.

philipjohn:
> Confirm that the categories, tags and edit link appear at the bottom of posts.

Looking at the standard "Hello World!" post, which has a tag of "content" I don't see that tag on the bottom of the post, only "Uncategorized" and the edit link.

> Create a sticky post (or check an existing sticky post) on the block posts page and confirm that it does not say 'Featured' at the top.

I'm not sure what "block posts page" refers to here - is it just a page using the homepage articles block? If so, I'm struggling to see stickies in there per my comment on [Automattic/newspack-blocks#28].

laurelfulford:
Thanks for the feedback here, @philipjohn!

> Looking at the standard "Hello World!" post, which has a tag of "content" I don't see that tag on the bottom of the post, only "Uncategorized" and the edit link.

Strange! My 'Hello World!' post does not have that tag out of the box, but when I manually add it, I get:

image

That's on master and this branch; I didn't actually change the output for the entry footer, so I'm not sure why this isn't working -- could you please double-check when you have a chance? 

> I'm not sure what "block posts page" refers to here - is it just a page using the homepage articles block?

Gah, I meant **_blog_** posts page, sorry! You don't want to know how many times I've autotyped this exact same thing since Gutenberg was launched 😐 

It would be whatever page you've picked to show your blog posts, whether it's the homepage or another page on the site -- it'll look something like this in the master branch: 

image

inc/template-tags.php
laurelfulford: Ah thanks! Wrapped `$time_string` in `wp_kses` and passed the expected `time` tag and attributes in e94b7de.

inc/template-tags.php
laurelfulford: 👍 Went with `wp_kses` again with specific tags, since the output here should have consistent markup -- added in 27e0829!

template-parts/header/entry-header.php
laurelfulford: TIL! I thought the `the_whatever` functions were all considered 'safe', but `the_title` looks to be a weird exception; flipped it around to use `get_the_title()` and added `wp_kses_post()` in 
151374a!

laurelfulford:
> must have been PEBKAC!

I run into a lot of those kinds of issues myself 😆 Thanks for testing this, @philipjohn!

PR #36, opened 2019-05-31
Add core functionality for the style packs

laurelfulford:
### All Submissions:

This PR adds existing core functionality to add style packs to the Newspack theme. 

As is, the code doesn't add anything to the interface yet (it will need an additional file like https://github.com/Automattic/themes/blob/master/radcliffe-2/inc/style-packs.php to define each style pack before they're added to the Customizer). So right now it mostly needs a code review. 

### Other information:

### DIFF
--- a/classes/class-newspack-style-packs-core.php
+++ b/classes/class-newspack-style-packs-core.php
@@ -2,6 +2,9 @@
 /**
  * Style Packs core functionality.
  *
+ * Based on functionality in Radcliffe 2:
+ * https://github.com/Automattic/themes/blob/master/radcliffe-2/inc/style-packs-core.php
+ *
  * @package Newspack
  */
 /**
--- a/classes/class-newspack-style-packs-core.php
+++ b/classes/class-newspack-style-packs-core.php
@@ -1,0 +1,406 @@
+<?php
+/**
+ * Style Packs core functionality.
+ *
+ * @package Newspack
+ */
+/**
+ * This class sets up a styles selector for the theme.
+ *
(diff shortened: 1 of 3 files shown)


laurelfulford:
> Before landing this PR could you leave a note about where the code comes from, just so we have a record if someone is trying to make sense of this in the future.

Sure thing! I can reference the first theme to have it added (Radcliffe 2) -- will that work? 

> I found this code duplicated in a number of WP.com and VIP themes. Are there any plans to either bring the feature into core or make it available for inclusion in themes without committing duplicated code?

I don't think so, no. It was kind of a theming experiment, to explore changing styles without actually changing a theme, and see how far it could be pushed, in anticipation of Gutenberg. I'm sure it would've been made more central if it was something we'd decided to add to more themes (it's only in three live themes, Photo Blog, Small Business and Radcliffe 2). 

> If we find there are aspects of this implementation that we don't end up needing (thumbnail support? print stylesheet?) would we remove? Or better to leave these files intact forever?

I think it makes sense to remove anything we're don't end up using, to keep things simplified. I have made a couple updates to the file compared to the other themes, to fully replace the stylesheet instead of adding it on top of the original, and to do a full refresh in the Customizer rather than trying update via JavaScript (we ran into issues with the preview working correctly with this, it would lose the style update when you made other changes). 

> In the Javascript file the syntax doesn't seem to match WordPress coding standards (e.g. no spaces around arguments). Any idea why?

I have JSPrettier set to fix too many files types on save 🤦‍♀ Will clean that up before committing.

PR #37, opened 2019-06-02
Fix full and wide aligned blocks on the static front page

laurelfulford:
### All Submissions:

The static front page template in the theme has a different width than the rest of the templates, and because of this, the full and wide alignment styles don't look great.

This PR makes sure they're aligned and sized correctly.

(There's some extra RTL styles in this PR that catches style-rtl.css up with the style.css from previous commits; sometimes the build script doesn't regenerate the RTL styles and needs to be re-run -- unfortunately, I didn't catch it when it happened last time!). 

1. On a test site, add a full and wide aligned block on the front page. 
2. Publish and view on the front-end. Both are likely pushed too far to the right, and hanging out of view.
3. Apply the PR.
4. Confirm that they now appear to be aligned correctly, and aren't cut off on the right. 

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -102,6 +102,7 @@
  */
 .entry .entry-content,
 .entry .entry-summary {
+
 	.entry-content,
 	.entry-summary,
 	.entry {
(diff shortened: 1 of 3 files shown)


laurelfulford:
Thanks @claudiulodro!

PR #41, opened 2019-06-04
Add bare-bones for first style pack option

laurelfulford:
### All Submissions:

This PR adds an option for a style pack, creatively named Style 1.

As part of it, it creates some new scss variable files. It reorganizes the existing Sass a bit, and it adds a new CSS file to the theme's build process. Ultimately, it enables us to change a few Sass variables, add additional styles, and regenerate a new CSS file for the theme to switch to.

Unfortunately that last bit -- regenerating the new CSS file -- makes this PR pretty large, but it's just basically adding a copy style.css with a few of the values changed.

**Still to do:**

* Set it up to create editor styles for this style pack (right now when you view the editor with Style 1 enabled, it will show Gutenberg's default styles). 
* Set it up to create RTL styles for this style pack (right now if you try to view the RTL styles, no stylesheet will load).
* Set it up to accept custom colours in the style pack -- to do this it will also be necessary to change the custom colour default (note that it may be overkill to do this, as colours can be customized individually already). 

1. Apply the PR to your test site.
2. Navigate to Customizer > Style Pack
3. Switch the style pack from 'Default' to 'Style 1'. 
4. Confirm that things look (a bit) different in the Customizer -- right now, it will only change the background colour to a light grey, change the fonts, and add some borders to the widget titles in the footer. 
5. Make sure you can switch back to the default style pack, and the styles update again in the preview.
6. Switch back to Style 1, Save and Publish, and view on the front end. 

It's not super exciting yet, but once the other styles are in place we can start exploring changing around the header :) 

### Other information:


laurelfulford:
> If time permits it would be great to articulate (on P2 or elsewhere) a grand vision for style packs in the Newspack theme.

That's a great idea! Added to my list :)

PR #40, opened 2019-06-03
Fix widths in editor styles

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR is meant to correct the content width in the editor, so for standard posts and pages, it should be about 780px, and for the static home page, it should be around 1200px.

Since Gutenberg doesn't include body classes in the editor outside of whether it's a post or page, the approach for adding a class to the homepage is a bit hacky; I'm definitely open to other ways to tackle that!

### How to test the changes in this Pull Request:

1. Apply the PR.
2. Set up the site with a static homepage, and verify the content width is about 1200px.
3. Check a single post in the editor, and verify the content width is about 780px.
4. Check a regular page in the editor, and verify the content width is about 780px.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #44, opened 2019-06-05
Alternate: Fix widths in editor styles

jeffersonrabb:
### All Submissions:

This PR is an alternate approach to https://github.com/Automattic/newspack-theme/pull/40. In this one instead of applying the `newspack-static-front-page` class to `editor-writing-flow` via Javascript (which feels brittle) this uses the `admin_body_class` filter to apply the class to the page body. The static front page styles are moved to a dedicated stylesheet (`style-editor-static-front-page.css`) which is enqueued conditionally if the page being edited is the static front page. Even though this adds a bit more code than the original PR I think it's cleaner since the Javascript hack is avoided and the styles are only enqueued when needed.

Closes # .

1. Follow procedures detailed in https://github.com/Automattic/newspack-theme/pull/40

### Other information:


js/homepage-editor-class.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * File homepage-editor-class.js
```
laurelfulford: Not a blocker, but this file can be removed (it's not being enqueued anymore, so it's not causing any issues).

jeffersonrabb: Good catch! Removed in https://github.com/Automattic/newspack-theme/pull/44/commits/a4becaa77421e0a3a0351973475386eeb93a3667

PR #42, opened 2019-06-05
Add default theme styles for the article blocks

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR adds some default styles for the Newspack article blocks, mostly around giving the section headers a distinct look. 

### How to test the changes in this Pull Request:

1. Add the homepage article blocks to a page, including the section header.
2. Apply the PR.
3. Confirm that the block section headers are blue (or using your primary accent colour), all-caps, and with a border underneath.
4. Confirm that the blocks look the same in the editor (note that the border detail won't be exactly the same; see the discussion in this block PR here for more info: https://github.com/Automattic/newspack-blocks/pull/30#pullrequestreview-245499130

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #49, opened 2019-06-05
Add editor styles for style packs

### LINKED ISSUE #47: Style pack: Add editor styles
The bare-bones style pack example added in #41 should also generate an enqueue editor styles, to give you a true WYSIWYG experience.

laurelfulford:
### All Submissions:

This PR adds editor styles for Style Pack 1.

Changes include adding a build for the style pack editor styles, and splitting the original styles out into a 'base' (styles that are built into every editor style sheet) and a default scss file (which mixes the default variables with the editor style base). 

It also adds the body background style to the editor styles, so it can be changed.

Closes #47.

1. Apply the PR.
2. Navigate to Customizer > Style Pack, and switch to Style 1.
3. Compare the front-end to the editor -- confirm that the fonts match (all serif), and that the background is a light grey (the two main differences between Style 1 and the default at the moment). 

### Other information:


jeffersonrabb:
I'm a little confused on this one—do I need to have the Newspack Homepage Articles block in a post to test?

laurelfulford:
You should have to -- it's just a little subtler without it. If you check out a regular post with the default styles, you'd have: 

image

And with style pack 1 assign plus the patch you should have a light grey background, and a serif header:

image

Prior to the patch, the editor won't be loading any styles at all when you pick style pack one -- so the content will be narrower, you'll have Gutenberg's default serif header, and a white background:

image

Just let me know if you run into any issues with this!

laurelfulford:
> Ideally we wouldn't have to muddy the theme with styling for one specific block... Could there be a more general way to do this, where the styles are applied based on a semantic tag, or maybe an additional style (.use-body-font or something?) is defined and then applied in the block?

For the font styles, absolutely: I overthought the header structure of those blocks. I was worried if the section header used an actual header tag, we could set it up to create a proper header structure if it was used (say, `<h2>` on the section headings, and `<h3>` on the articles), but an incorrect one if it was not (jumping right to the `<h3>` on articles). I think the quickest fix would be to get a semantic header tag in there -- at the risk of overthinking this even more, it should be possible to switch the markup used on the article title based on whether there's a section header, to use `<h3>` if there is one, and `<h2>` if there's not.

(I can get working on separate PRs for the theme and block plugin to fix up that markup and minimize styling). 

For other kinds of styles -- like the underline in #42 -- that might be better handled with a special class, but we still might end up with block-specific styles in the theme for weird cases, like to get it displaying correctly in the editor 😕 I'm not sure the best approach there -- maybe a separate stylesheet? Or turn it into a block style, instead of something that would be in the theme? 

> I'm wondering if we can cook up a build process that automatically iterates over all directories in /sass/styles and processes all based on the same rule.

I love this idea! Especially since you have to add not one, but three lines to build everything for a single style pack -- it would cut out so much opportunity for forgetting something (my specialty!)

PR #55, opened 2019-06-05
Fix issue with style pack preview overriding custom colours and fonts

### LINKED ISSUE #52: Style Pack: Custom fonts, colours overridden in the preview
In the preview, if the styles define different fonts or colours, they override the custom fonts and colours in the Customizer preview.

This is because, in the preview, the style pack styles are enqueued via JavaScript, and at the very end of the header. Any styles that come before it are overridden. 

I'm not totally convinced we _need_ to include colours/fonts in the style packs, but this would be a nice thing to fix. 

Steps to reproduce:

1. Navigate to Customizer > Style Packs, and pick Style Pack 1.
2. Navigate to Typography, and set up a custom typeface. 
3. Note it's not previewing.
4. Navigate to front end; note the style pack and typeface is correct.
5. Navigate back to the customizer - the font will not be wrong.

laurelfulford:
### All Submissions:

The custom colours and typography styles get over-riden in the Customizer preview by the Style pack styles. This is because, in the preview, they are appended as the last item in the `<head>`, so they are the last styles loaded.

I thought I could rework how the style pack preview works to remove the JavaScript all together, but it's turning out to be a little trickier than I thought. So instead I created an optional value that could be set for the style pack settings -- the main CSS file's ID -- and if populated, used it to insert the style pack preview styles right after the theme's default style.css. That way, the custom font and colour styles come after it, and successfully override the values. 

If the value is left empty, the styles are added to the bottom of the `<head>` in the preview. 

Closes #52.

1. Navigate to Customizer > Typography, and set a custom font.
2. Navigate to Customizer > Style Packs, and pick Style Pack 1. (If you've already changed the style pack and changed back to the default at any point, you'll still get this issue). 
3. Save and Publish.
4. Return to the Customizer; confirm that your fonts are not previewing. 
5. Apply the PR.
6. Refresh the Customizer. Confirm that the correct font is now displaying. Try changing the style packs/fonts again, and confirm they continue to be correct. 

### Other information:

### DIFF
--- a/classes/class-newspack-style-packs-core.php
+++ b/classes/class-newspack-style-packs-core.php
@@ -54,6 +54,7 @@ class Newspack_Style_Packs_Core {
 					'body_class_format'  => 'style-pack-%s',
 					'styles_directory'   => 'styles',
 					'js_directory'       => 'js',
+					'default_css_id'     => '',
 					'style_thumbs'       => array(),
 					'style_descriptions' => array(),
 					'default_headers'    => array(),
@@ -277,6 +278,7 @@ class Newspack_Style_Packs_Core {
 			'preview_style'     => $this->get_preview_style(),
 			'styles'            => $style_pack_stylesheets,
 			'fonts'             => $style_pack_fonts,
+			'default_css_id'    => $this->config['default_css_id'],
 		);
 		wp_localize_script( 'style-packs-customizer', 'stylePacksData', $style_packs_data );
 	}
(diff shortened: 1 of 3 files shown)


jeffersonrabb:
I propose we hold on this one for a minute and see if we can find a cleaner solution. Since this bug only affects preview, it won't block broader theme/typography efforts so getting this landed quickly isn't super necessary. Let's discuss some alternate approaches.

laurelfulford:
@jeffersonrabb I've double-checked this PR, and it does still work: it fixes the issue with the custom typography and colours not previewing properly in the customizer.

The style packs themselves still have some preview issues (the default style pack always appears to be applied, with any others applied 'on top' of it); this is something separate we'll need to dig into, but getting this one landed will help.

laurelfulford:
Thanks @jeffersonrabb!

PR #61, opened 2019-06-06
Try out an approach for customizing headers

laurelfulford:
### All Submissions:

I suggested this as one way to allow different headers layouts in the theme prior to full-page editing: 
This PR adds a 'Header Layout' dropdown to the Site Identity section of the customizer, and has two options on top of the default: one to centre the logo, and one to use the primary colour as a background. These aren't finalized, more of a way to test how easy it would be to change layout and how colours are assigned. 

I had originally considered including the headers as part of the style packs, but it seems like it would make more sense to keep it independent; I'm interested to know if this seems like a logical approach, and if there are any concerns about it. 

1. Apply the PR.
2. Navigate to Customizer > Site Identity, and try switching between different header options (note that #52 will cause issues with the 'Solid' option). 

### Other information:

### DIFF
--- a/header.php
+++ b/header.php
@@ -38,7 +38,7 @@
 						?>
 					</nav>
 				<?php endif; ?>
-				<?php if ( has_nav_menu( 'social' ) ) : ?>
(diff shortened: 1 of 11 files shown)


laurelfulford:
Closing this in favour of the approach tried in #72

PR #68, opened 2019-07-04
Add styles for single article author bio. 

laurelfulford:
### All Submissions:

This PR roughs in the styles for the author bios at the bottom of each story.

Still outstanding:

* Additional fields -- right now, the job title ("Staff Writer") is hard coded. We need to finalize what fields we might need (title, additional social media like Twitter), and add dynamic fields. I'm also wondering if displaying email should be optional (since it's a required field in WordPress).
* Should showing the bio at all be an option (say, for shorter stories). 
* Co-Authors Plus -- assuming we go with this plugin, this code will need to be updated to include its custom functions, with the standard ones as a fallback.

1. Apply the PR and run `npm run build` (edited to add the build).
2. View a story where the assigned author has a description (right now, the theme copies Twenty Nineteen and requires the author to have a description before the profile is displayed at the bottom). 
3. Confirm that the styles match those mocked up in 'Style A': https://cloudup.com/ilw9quyPhje

### Other information:


laurelfulford:
> In my testing in MS Edge the Avatar doesn't show at all.

What version was this (of Windows & Edge)? I'm not seeing the issue using Browserstack, but I probably just don't have the right combo. Right now the image is hidden on mobile to save space -- but I'm double-guessing that as an approach if that's why it's not showing for you (don't want the fallbacks to just be confusing!) 

> Doing a strict comparison to the mockup, it looks like name and title ("Staff Writer") are lighter font weights than the mockup. 

To me they look to be the right weight, but not quite the right size (fixing that now). That might be why they don't appear right -- the mockups also use the Mac default sans-serif font, and it might not line up quite right on Windows -- I can double-check the font stack for that to see if it can be improved, if that's the case!

laurelfulford:
@jeffersonrabb I've updated the font sizing -- I had set a smaller base size for the whole profile, messing up what was being inherited (everything was about 2px too small). Let me know if that fixes whether the weights look correct!

laurelfulford:
I've also updated the mobile styles, so the avatar is no longer hidden there; not sure if that was the issue with Edge but it resolves my double-guessing around that choice :)

laurelfulford:
> Ignore my comments about font weight. I had custom typography defined from back when we were working on that feature, and this was to blame for the visual differences. 

Phew! I couldn't tell looking at it, either :| Glad you spotted that!

> However, when I view on mobile I don't see the Avatar.

I _think_ this might be the same missing build step issue as #76 -- can you confirm? I'll make sure to include that in the PR testing going forward, since it's a shift from previous testing steps.

laurelfulford:
@jeffersonrabb Okay I think I have the avatar issue solved now -- sorry about that! I know I wrote the styles but I missed some important step, I'm declaring it a general #gitfail on my part. 

Just let me know if there are any other issues!

PR #76, opened 2019-07-09
Reorder single post and update styles

laurelfulford:
### All Submissions:

This PR is a version rescuing changes I made, and tried to recover, from #69. I originally thought I accidentally merged some of that PR, but on inspection it looks like I just blew it up without affecting the master branch (CC @jeffersonrabb, since i mislead you on this point earlier). 

This PR makes updates to the single article's appearance and order to match the mockups; specific changes include:

1. Moving Jetpack's sharing icon output to site up next to the title.
2. Moving the category output above the title.
3. Adjusts general styles and spacing.

1. Apply the PR and run `npm run build` (edited to add the build).
2. Activate Jetpack's share buttons; set to just show icons.
3. Review single stories and compare against the mockups; make sure they generally look correct. Note: the sharing icons display different with AMP enabled (large and rectangular, not smaller and rounded) - I'm less familiar with the styling there, so have opted not to try to override it for now.

### Other information:

### DIFF
--- a/inc/jetpack.php
+++ b/inc/jetpack.php
@@ -72,6 +72,15 @@ function newspack_infinite_scroll_render() {
 }
 
 /**
+ * Remove Jetpack Share icons from standard location so they can be moved.
+ */
+function newspack_remove_jetpack_share() {
+	remove_filter( 'the_content', 'sharing_display', 19 );
+	remove_filter( 'the_excerpt', 'sharing_display', 19 );
+}
+add_action( 'loop_start', 'newspack_remove_jetpack_share' );
+
+/**
  * Alter gallery widget default width.
  */
 function newspack_gallery_widget_content_width( $width ) {
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -41,6 +41,11 @@ function newspack_body_classes( $classes ) {
 		$classes[] = 'has-sidebar';
 	}
 
+	// Adds class if singular post or page has a featured image.
+	if ( is_singular() && has_post_thumbnail() ) {
+		$classes[] = 'has-featured-image';
+	}
+
 	return $classes;
 }
 add_filter( 'body_class', 'newspack_body_classes' );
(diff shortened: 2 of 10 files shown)


jeffersonrabb:
@laurelfulford pointed out quite correctly that I had failed to execute the build script. My apologies! After doing this, here's what the same region looks like:

Screen Shot 2019-07-10 at 11 24 10 AM

- Category styling is resolved.
- Gray rule is resolved (although collision with the Search widget is worth a look)

This leaves only these questions open:

- The Jetpack share buttons are very different from the mockups, in style and placement. Maybe there's a Jetpack setting I'm supposed to change?
- Where is the dek text drawing from, and how can I get set up to view that?

laurelfulford:
Phew! I was worried I'd somehow messed up these PRs :) 

> The Jetpack share buttons are very different from the mockups, in style and placement. Maybe there's a Jetpack setting I'm supposed to change?

Yeah, it was kind of buried in the steps: you need to set the button style to 'Icon only':

image

The placement looks correct to me, though -- can you add some more details about that if they still look off when it's switched to icons? 

If you run with AMP enabled, this is the 'squared' design I mentioned above which I need to dig into: 

image

> Where is the dek text drawing from, and how can I get set up to view that?

Ah crud, I meant to add a note about that 😬 I still need to figure that piece out. If we can turn the "post meta" (byline, date and shares) into a block, it can be a paragraph, with a paragraph style we set. If not, we might have to go a custom field route. 

I've also seen the Custom Excerpt used to do something similar ([Twenty Sixteen](https://twentysixteendemo.wordpress.com/) being an example), but that would show up on the homepage and archives, and the same content probably wouldn't make sense in the two different contexts.

laurelfulford:
Oh! And one more thing I didn't comment on because it had to do with the missing build, but is still a detail I should've included:

> Is the light gray horizontal rule expected to appear?

That border is only there when there isn't a featured image; if you add a featured image, it kind of acts like the 'barrier' between the header and content.

laurelfulford:
For completeness, I moved the dek question here: #98

PR #100, opened 2019-07-11
Add separate front-page.php template to theme

### LINKED ISSUE #99: Sidebar is showing up on static front page
In #71, a sidebar section was added to page.php without my considering that that file is also used for the front page. So if you add a widget to the sidebar, it also displays below the front-page content at full width:

image

laurelfulford:
### All Submissions:

This PR adds a front-page.php template to the theme. There's also a filter to make sure this template is only used on the static front page and not when the front page is set to a blog posts page, just in case -- the filter is borrowed from the same approach used in Twenty Seventeen, and saves having to add a home.php as well to avoid that issue.

I opted to add a new template instead of a conditional check to page.php so the sidebar wouldn't show up, in case we need additional changes to the front page. 

Closes #99 .

1. Set up your test site with a sidebar. 
2. View the static front page; confirm that the sidebar content displays below the page content.
4. Refresh the static front page; confirm that the sidebar content is gone. 
5. Navigate to Customizer > Homepage Settings, and set the homepage to display the blog posts 
6. View the front page; confirm that the sidebar is visible (confirming that it's not using the new front-page.php). 

### Other information:

### DIFF
--- a/front-page.php
+++ b/front-page.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * The template for displaying all single posts
+ * The template for displaying the static front page.
  *
  * @link https://developer.wordpress.org/themes/basics/template-hierarchy/#single-post
  *
--- a/front-page.php
+++ b/front-page.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * The template for displaying all single posts
+ *
+ * @link https://developer.wordpress.org/themes/basics/template-hierarchy/#single-post
+ *
+ * @package Newspack
+ */
+
+get_header();
+?>
+	<section id="primary" class="content-area">
+		<main id="main" class="site-main">
+
+			<?php
+
+			/* Start the Loop */
+			while ( have_posts() ) :
+				the_post();
+
+				get_template_part( 'template-parts/content/content', 'page' );
+
+				// If comments are open or we have at least one comment, load up the comment template.
+				if ( comments_open() || get_comments_number() ) {
+					comments_template();
+				}
+
+			endwhile; // End of the loop.
+			?>
+
+		</main><!-- #main -->
+	</section><!-- #primary -->
+
+<?php
+get_footer();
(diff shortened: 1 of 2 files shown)


front-page.php
```diff
@@ -0,0 +1,35 @@
+<?php
+/**
+ * The template for displaying all single posts
```
jeffersonrabb: We should update this file doc to describe the front page rather than single posts.

functions.php
```diff
@@ -326,6 +326,17 @@ function newspack_enqueue_editor_static_front_page_assets( $classes ) {
...
+ * Use front-page.php when Front page displays is set to a static page.
+ *
+ * @param string $template front-page.php.
+ *
+ * @return string The template to be used: blank if is_home() is true (defaults to index.php), else $template.
+ */
+function newspack_front_page_template( $template ) {
+	return is_home() ? '' : $template;
+}
+add_filter( 'frontpage_template', 'newspack_front_page_template' );
```
jeffersonrabb: This is probably just me being dumb, but I don't understand what this filter accomplishes? I seem to be getting the same results with or without it. (I see that [it's in TwentySeventeen](https://github.com/WordPress/twentyseventeen/blob/master/functions.php#L372-L384) but I'm hazy on what it's doing there too...)

laurelfulford: It has to do with WordPress's template hierarchy:

If you have a static front page, it will look for front-page.php, then page.php, then index.php (excluding templates that aren't relevant to our theme). 

The blog posts page, if set to the front page, will look for home.php, then front-page.php, then index.php.

Since we don't have a home.php, if you set the blog posts page as the homepage, it will use the front-page.php template unless that filter is there. Which is not a case we expect with this theme, but you never know! Things should look mostly the same, except there is no sidebar; if we make further changes to the front-page.php, those would show up, too.

We could also work around this by adding a home.php.

In Twenty Seventeen's case, things only get hairy if you've set up the 'front page panels' in the Customizer -- the way you set up the [different panels](https://twentyseventeendemo.wordpress.com/) for the static front page. Without the filter, if you switched a site to use the blog posts on the front page, those panels would show up on the blog posts page. 

TL;DR: it makes sure you have have sidebar widgets if you opt to set the blog posts page as the front page :)

front-page.php
laurelfulford: Added in d7d3890!

laurelfulford:
I went ahead and merged this one, but let me know if you have any further questions about that front page filter!

Issue #102, opened 2019-07-12
Update VIP Coding Standard


GaryJones:
**Describe the bug**
The [currently used version of VIP Coding Standards](https://github.com/Automattic/newspack-theme/blob/0a05698993dcd454865fd07040a20fafa427a3ea/composer.json#L16) is `0.4.0`. This is out of date, and is not compatible with the latest versions of WPCS.

**Expected behavior**
Please use [VIPCS 2.0](https://github.com/Automattic/VIP-Coding-Standards/releases/tag/2.0.0), which was released a few minutes ago :-)

laurelfulford:
It looks like this has been closed with #105!

PR #105, opened 2019-07-15
Upgrade to v2.0.0 of the VIP Coding Standards

philipjohn:
Simply upgrades our use of the VIP Coding Standards so we're not out of date!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,6 @@
     "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
     "brainmaestro/composer-git-hooks": "^2.6",
     "automattic/vipwpcs": "^2.0.0",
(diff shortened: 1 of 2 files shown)


GaryJones:
Fixes https://github.com/Automattic/newspack-theme/issues/102.

GaryJones:
`composer uninstall wptrt/wpthemereview wp-coding-standards/wpcs` and then do the `composer install`.

https://github.com/WPTRT/WPThemeReview/blob/develop/composer.json#L42 shows that the theme review package does support WPCS 2.1, so it's just what's in your `composer.lock` that is causing the conflict.

GaryJones:
You may still want the wpthemereview package, but you'd need to use the `dev-develop` branch for it to work with VIPCS.

Issue #96, opened 2019-07-10
Build theme header


laurelfulford:
The Newspack theme's header is kind of complex, so I wanted to list out the requirements in an issue so I can keep track of it. The tasks listed here will likely grow longer, and may at some point be broken out into separate issues where it makes sense.

- [x] Build base styles for desktop
- [x] Build base styles for AMP (mobile and desktop) - started in #120
- [x] Build non-AMP styles and behaviour for mobile menu
- [ ] Make sure non-mobile menu works well on touch devices - AMP
- [ ] Make sure non-mobile menu works well on touch devices - non-AMP
- [x] Add 'header background' toggle and add styles
- [x] Add 'logo center' toggle and add styles (related exploration #72)
- [x] Add 'shorter' toggle and add styles (related exploration #72) - PR pending in #119
- [ ] ~~Add 'overlap' toggle and add styles (related exploration #72)~~ have opted to move this to the style pack -- see discussion in #112 
- [ ] Add 'overflow' offscreen menu, both for sections and for menus not shown in some cases (like the 'shorter' simplified menu). 
- [x] Add search icon and ability to toggle the search - PR pending in #116
- [x] Test/fix/add custom colour support
- [x] Test/fix/add custom typography support

Some related issues:

Needs Design: #89, #85 

Enhancements: #90, #88, #84

laurelfulford:
Though the exact tasks aren't completed as written, I would consider the theme header complete; we can handle specific requests in different issues as they come up.

PR #109, opened 2019-07-18
Add options to customize the header (solid background and centred logo)

laurelfulford:
### All Submissions:

This PR starts the process of building out the header options for the theme. 

It adds two checkboxes -- one to switch the header to use a solid background colour, and one to center the logo.

Two additional options -- one to make a 'short' header, and one to make the content overlap the header -- still need to be added. 

The aim is to make it possible to mix-and-match any of these options.

Mockup of default (no background, logo on the left):

image

Mockup of both background colour and centred logo:

image

Plus my own screenshots of how the other variations should look -- just background colour: 

image

... and just centred logo:

image

This is part of a larger ticket to build out the full header functionality: #96. So while things should not look terrible on small screens or in certain circumstances, they will be progressively improved as we work down that list. 

To start:
2. Set up the header -- I'd recommend adding at least a logo and tagline, and primary, social and tertiary menu; you can also add a secondary one. 

Testing the solid background colour:
1.  Under Customizer > Site Identity, and check 'Solid Background', then click Publish.
2. Confirm that the background colour of the header changes to the blue, and the text and links to the white. 
3. Test changing the custom background colour to a light colour, and make sure the text turns dark (note that this won't preview correctly due to #52, so the changes will need to be saved and published).
4. Uncheck 'Solid Background'

Testing the centred logo:
1. Under Customizer > Site Identity, check 'Center Logo', then click Publish.
2. Confirm that the logo and tagline is now centred, and the social links (if assigned) appear to the left. 
3. Confirm that the 'primary' navigation is now centred. 

Note: With the centred logo, if you don't have a social menu assigned, what you see will be misaligned, like:
image

We have a couple options here -- one would be to add something else, like the search icon, to that location in that case to even things out. Another would be not to align the logo centre unless there's something to the left of it... unfortunately that one could be confusing as all get out.

I plan to explore a fix in a future PR.

Why not both?
1. Go back to Customizer > Site Identity and check 'Solid Background'.
2. Confirm that things look similar to the second screenshot above, with elements centred and with the primary colour used as the background.

### Other information:

### DIFF
--- a/header.php
+++ b/header.php
@@ -21,81 +21,98 @@
 <div id="page" class="site">
 	<a class="skip-link screen-reader-text" href="#content"><?php _e( 'Skip to content', 'newspack' ); ?></a>
 
-		<header id="masthead" class="site-header">
+	<header id="masthead" class="site-header">
 
-			<div class="top-header-contain">
-				<div class="wrapper">
-					<?php if ( has_nav_menu( 'secondary-menu' ) ) : ?>
-						<nav class="secondary-menu" aria-label="<?php esc_attr_e( 'Secondary Menu', 'newspack' ); ?>">
-							<?php
-							wp_nav_menu(
-								array(
-									'theme_location' => 'secondary-menu',
-									'menu_class'     => 'secondary-menu',
-									'items_wrap'     => '<ul id="%1$s" class="%2$s">%3$s</ul>',
-									'depth'          => 1,
-								)
-							);
-							?>
-						</nav>
-					<?php endif; ?>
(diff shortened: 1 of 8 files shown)


laurelfulford:
Attaching a couple logos to make testing easier:

All white (I promise, there's an image here): 
newspack-logo-white

Full-colour: 
newspack-logo-blue

laurelfulford:
Thanks for digging into this one, @claudiulodro! It was a big of a big one :)

PR #112, opened 2019-07-18
Add option to have the header overlap the content, and related styles.

laurelfulford:
### All Submissions:

This PR adds an option to have the content area overlap the header, similar to what's mocked up in the Style C mockups:

image

There's still one outstanding header customization option -- the ability to switch to a simpler header -- so it's not possible to create an identical header yet.

This style is ideally paired with a solid background colour, but without there is a light grey background to help distinguish the header from the content area: 

image

And here's how it will look with a solid background on the header, and no primary menu:

image

Part of #96.

1. Apply the PR, and run `npm run build`. 
2. Navigate to Customize > Site Identity, and check 'Content Overlap' 
3. Verify that the content overlaps the header, and that it looks okay.
4. Test with the other two header customization options -- background and centred logo.

Smaller screens overall for the header still need to be finalized, but things should still look okay on mobile, just in need of some scaling/spacing finalizing. 

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -191,6 +191,17 @@ function newspack_custom_colors_css() {
 		';
 	}
 
+	if ( true === get_theme_mod( 'header_overlap', false ) ) {
+		$theme_css .= '
+			.header-overlap.header-solid-background .bottom-header-contain {
+				background-color: ' . $primary_color . ';
+			}
+			.header-overlap .site-content #primary {
+				border-color: ' . newspack_adjust_brightness( $primary_color, -40 ) . ';
+			}
+		';
+	}
+
 	$editor_css = '
 		/*
 		 * Set colors for:
--- a/inc/customizer.php
+++ b/inc/customizer.php
@@ -134,6 +134,24 @@ function newspack_customize_register( $wp_customize ) {
 		)
 	);
 
+	// Header - add option for header overlap.
+	$wp_customize->add_setting(
+		'header_overlap',
+		array(
+			'default'           => false,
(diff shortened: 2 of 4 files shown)


laurelfulford:
> When studying the mockups, I had imagined this setting was a style pack, not a separate setting. This may reveal my misunderstanding of the concept of style packs, but I thought this was exactly the type of change they would be responsible for.

That's a really good point! I think as soon as I started thinking about the header being customizable, I just mentally filed this away in that pile of settings, since it was something that changed. But there's a few reasons why it'd probably work better in the style pack -- a big one being that it is what makes the styles in 'C' distinct, everything else is kind of downplayed to work with such a bold header. It's also the one setting that affects more than just the header -- the content becomes narrower, too. 

> If this is going to be discreet, settable customizer feature, do you think we should allow control over how far down the header sits?

It'd be technical possible, and pretty easy to adjust; I guess it would depend on where the setting ends up. I could see it being helpful in some cases, like if someone needs a ton of menus, the same spacing you'd want for a very simple header isn't going to look quite right. But if it's style-pack based, I'm not sure where the setting should be -- maybe something that gets shown/hidden in the site identity with the other header options, but only if you pick that style pack? 

TL;DR: I think moving this into the one style pack makes sense; I'm also open to exploring some kind of height option -- I think it could be helpful though I'm not really sure where it would go. :)

laurelfulford:
I'm going to close out this PR for now and rework the styles into the related style pack.

jeffersonrabb:
As noted, my two comments were essentially contradictory, an either/or situation. So, if you're going the Style Pack route there's no need to worry about a height control!

PR #116, opened 2019-07-23
Adds search toggle and search box to header

laurelfulford:
### All Submissions:

This PR adds a search box and toggle to the theme's header.

It builds out the functionality using AMP, and includes a JavaScript fallback that should only be loaded on non-AMP pages. 

Right now it's pretty basic visually; it displays a search form that floats below the toggle. I didn't want to get too far into the weeds for how it should function with the different header layouts, so it remains next to the primary menu, whether or not the header is centred.

Search icon in line with the primary menu:  
image

Search opened:
image

Related to #96

2. Make sure AMP is set to 'standard', or 'transitional', and that you are viewing an AMP-enabled version of a page.
3. Click on the search icon, and confirm that the search opens; toggle it a couple times to check the following:
    * Confirm that the search icon turns into an 'X', and back to a magnifying glass.
    * Inspect the element and check that the screenreader text label changes from 'Open Search' to 'Close Search'
    * Confirm that the `aria-expanded` attributes on both the button and search container toggle between 'true' (opened) and 'false' (closed). 
4. Navigate to a non-AMP page, or switch AMP to Transitional or Reader mode, and test the the points in step three again to make sure they work the same. 

I welcome any pointers for the AMP implementation, too -- I mostly relied on toggling classes to show/hide things, but that might not be the best approach. 

### Other information:

### DIFF
--- a/classes/class-newspack-svg-icons.php
+++ b/classes/class-newspack-svg-icons.php
@@ -166,7 +166,7 @@ class Newspack_SVG_Icons {
     </g>
 </svg>',
 
-		'search'                   => /* martial-design - search */ '
+		'search'                   => /* material-design - search */ '
 <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
     <path d="M15.5 14h-.79l-.28-.27C15.41 12.59 16 11.11 16 9.5 16 5.91 13.09 3 9.5 3S3 5.91 3 9.5 5.91 16 9.5 16c1.61 0 3.09-.59 4.23-1.57l.27.28v.79l5 4.99L20.49 19l-4.99-5zm-6 0C7.01 14 5 11.99 5 9.5S7.01 5 9.5 5 14 7.01 14 9.5 11.99 14 9.5 14z"/>
     <path d="M0 0h24v24H0z" fill="none"/>
--- a/classes/class-newspack-svg-icons.php
+++ b/classes/class-newspack-svg-icons.php
@@ -165,9 +165,16 @@ class Newspack_SVG_Icons {
         <path fill="currentColor" fill-rule="nonzero" d="M12 2c5.52 0 10 4.48 10 10s-4.48 10-10 10S2 17.52 2 12 6.48 2 12 2zM6 14a2 2 0 1 0 0-4 2 2 0 0 0 0 4zm6 0a2 2 0 1 0 0-4 2 2 0 0 0 0 4zm6 0a2 2 0 1 0 0-4 2 2 0 0 0 0 4z"/>
     </g>
 </svg>',
+
 		'search'                   => /* martial-design - search */ '
 <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
     <path d="M15.5 14h-.79l-.28-.27C15.41 12.59 16 11.11 16 9.5 16 5.91 13.09 3 9.5 3S3 5.91 3 9.5 5.91 16 9.5 16c1.61 0 3.09-.59 4.23-1.57l.27.28v.79l5 4.99L20.49 19l-4.99-5zm-6 0C7.01 14 5 11.99 5 9.5S7.01 5 9.5 5 14 7.01 14 9.5 11.99 14 9.5 14z"/>
+    <path d="M0 0h24v24H0z" fill="none"/>
+</svg>',
+
+		'close'                    => /* material-design - close */ '
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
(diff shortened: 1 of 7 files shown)


sass/site/header/_site-header.scss
```diff
@@ -92,6 +88,46 @@
...
+		display: none;
+	}
+}
+
+.header-search-contain {
+	position: relative;
+
+	#header-search {
+		position: absolute;
+		top: calc( 100% + 2px );
```
jeffersonrabb: What would you think about giving this a tiny bit more air? I'm seeing the search form overlap the bar ever so slightly, and it tracks as an error:

Screen Shot 2019-07-25 at 12 29 03 PM

sass/site/header/_site-header.scss
```diff
@@ -92,6 +88,46 @@
...
+		width: 300px;
+
+		@include media (tablet) {
+			right: 0;
+		}
+	}
+}
+
+.hide-header-search {
+	#search-toggle {
```
jeffersonrabb: When the search field is open there's an outline around the toggle button. Unless you're aware of an accessibility argument for keeping it, what would you think of setting `outline: none;` for the `:focus` state?

header.php
```diff
@@ -21,7 +21,7 @@
 <div id="page" class="site">
 	<a class="skip-link screen-reader-text" href="#content"><?php _e( 'Skip to content', 'newspack' ); ?></a>
 
-	<header id="masthead" class="site-header">
+	<header id="masthead" class="site-header hide-header-search" [class]="searchVisible ? 'show-header-search' : 'hide-header-search'">
```
jeffersonrabb: When the toggle occurs this element loses the `site-header` class. To see this in action, use a header with a background and then toggle the search. You should lose the background color the first time you click. I'm not sure this proposal is the most elegant solution, but it will work for now and we can audit all unusual AMP bits later:

```suggestion
	<header id="masthead" class="site-header hide-header-search" [class]="searchVisible ? 'site-header show-header-search' : 'site-header hide-header-search'">
```

template-parts/header/header-search.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Displays the search form for the header.
+ *
+ * @package Newspack
+ */
+?>
+
+<div class="header-search-contain">
+	<button id="search-toggle" on="tap:AMP.setState({searchVisible: !searchVisible})" aria-controls="search-menu" [aria-expanded]="searchVisible ? 'true' : 'false'" aria-expanded="false">
```
jeffersonrabb: Unless you're aware of a reason why this might be an AMP problem, let's space this out according to WordPress coding convention:

```suggestion
	<button id="search-toggle" on="tap:AMP.setState( { searchVisible: ! searchVisible } )" aria-controls="search-menu" [aria-expanded]="searchVisible ? 'true' : 'false'" aria-expanded="false">
```

classes/class-newspack-svg-icons.php
```diff
@@ -165,12 +165,19 @@ public static function get_social_link_svg( $uri, $size ) {
         <path fill="currentColor" fill-rule="nonzero" d="M12 2c5.52 0 10 4.48 10 10s-4.48 10-10 10S2 17.52 2 12 6.48 2 12 2zM6 14a2 2 0 1 0 0-4 2 2 0 0 0 0 4zm6 0a2 2 0 1 0 0-4 2 2 0 0 0 0 4zm6 0a2 2 0 1 0 0-4 2 2 0 0 0 0 4z"/>
     </g>
 </svg>',
+
 		'search'                   => /* martial-design - search */ '
```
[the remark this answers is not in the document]
laurelfulford: Ha! Yes - fixed!

sass/site/header/_site-header.scss
laurelfulford: Absolutely -- I bumped it down a couple more pixels. I'd like to experiment a bit more with how it looks, but I think it's a lower priority right now.

sass/site/header/_site-header.scss
laurelfulford: Unfortunately, it needs to stay -- that button still has the focus, and it's necessary to have some kind of visual feedback for what's in focus, for when you're navigating with a keyboard.

However, I've updated the JS and AMP to move the focus to the search field when it's open -- that's better for accessibility anyway. The one downside is I couldn't figure out how to remove focus in AMP when the search is closed -- it does return to the 'close' button when you tab back to it, but if you just click it, it doesn't. Given that this is for keyboard navigation I don't think that will actually be an issue, but I can double-check.

header.php
laurelfulford: Oh shoot, I totally missed that! Good work around -- another option may be to move the toggled class to another element.

template-parts/header/header-search.php
laurelfulford: Updated - it doesn't seem to cause issues!

laurelfulford:
Thanks for the review, Jeff! All the issues should be resolved.

laurelfulford:
Thanks Jeff!

PR #120, opened 2019-07-24
WIP: Adds mobile menu using AMP

laurelfulford:
### All Submissions:

This PR is a WIP of adding a mobile-style menu using AMP functionality. 

I wanted to share it to get some feedback on the approach -- it does duplicate quite a bit of code, and essentially hides the dupes when not in use (so they're ignored by screen readers, but still there). 

AMP does have an approach ([toolbar-target](https://amp.dev/documentation/components/amp-sidebar/#toolbar-target)) where you can have a `nav` in the sidebar automatically copied into a specific ID at a certain breakpoint, which would reduce the duplication. However:

* We'd need to add the duplication back for the non-AMP fallback. So it wouldn't be in the output of an AMP site, but it would be in the PHP files in some form.
* I'm a little worried the `toolbar-target` approach would get confusing with the complexity already in the header (it will involve empty divs). 
* Out of the box I got some weird styling trying the `toolbar-target` approach (a menu displaying in a very small space with scrollbars); but I didn't dig to far into how much work it would be to override. 
* It might also make sense to stick with this duplication approach for now, and optimize it later on. 

I'm not against trying the `toolbar-target` approach, but I didn't want to get too far into the weeds before getting a second opinion. 

See #96.

2. Add all the menus; primary, secondary, tertiary. 
3. Make sure AMP is set to 'standard'.
4. Shrink down the browser window until the mobile styles kick in (the regular menus will disappear and a 'Menu' toggle will display). 
5. Open close the menu; review the appearance. 
6. Check over the code (the bulk of the duplication is in the new mobile-sidebar.php file) and weigh in.

### Other information:

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -226,6 +226,42 @@ function newspack_get_discussion_data() {
 }
 
 /**
+ * Add an extra menu to our nav for our priority+ navigation to use
+ *
+ * @param object $nav_menu  Nav menu.
+ * @param object $args      Nav menu args.
+ * @return string More link for hidden menu items.
+ */
+function newspack_add_ellipses_to_nav( $nav_menu, $args ) {
+
+	if ( 'primary-menu' === $args->theme_location ) :
+
+		$nav_menu .= '<div class="main-menu-more">';
+		$nav_menu .= '<ul class="main-menu">';
+		$nav_menu .= '<li class="menu-item menu-item-has-children">';
+		$nav_menu .= '<button class="submenu-expand main-menu-more-toggle is-empty" tabindex="-1" aria-label="More" aria-haspopup="true" aria-expanded="false">';
+		$nav_menu .= '<span class="screen-reader-text">' . esc_html__( 'More', 'newspack' ) . '</span>';
+		$nav_menu .= newspack_get_icon_svg( 'arrow_drop_down_ellipsis' );
+		$nav_menu .= '</button>';
+		$nav_menu .= '<ul class="sub-menu hidden-links">';
+		$nav_menu .= '<li id="menu-item--1" class="mobile-parent-nav-menu-item menu-item--1">';
+		$nav_menu .= '<button class="menu-item-link-return">';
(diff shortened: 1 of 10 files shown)


laurelfulford:
I created a second example PR in #129 that shows the `toolbar-target` approach -- the menu code only appears once (in the mobile-sidebar.php file, but for a non-AMP fallback it will need to be repeated.

inc/template-functions.php
```diff
@@ -225,6 +225,42 @@ function newspack_get_discussion_data() {
...
+ *
+ * @param object $nav_menu  Nav menu.
+ * @param object $args      Nav menu args.
+ * @return string More link for hidden menu items.
+ */
+function newspack_add_ellipses_to_nav( $nav_menu, $args ) {
+
+	if ( 'primary-menu' === $args->theme_location ) :
+
+		$nav_menu .= '<div class="main-menu-more">';
```
westonruter: This `div` is causing a validation error here. The AMP plugin is reporting the element as being invalid here because it is violating this constraint in the [`amp-sidebar > nav` tag spec](https://github.com/ampproject/amphtml/blob/e9f3857405939612d3a84a7afd10f59acd504b85/extensions/amp-sidebar/validator-amp-sidebar.protoascii#L75-L97):

```
  child_tags: {
    mandatory_num_child_tags: 1
    child_tag_name_oneof: "OL"
    child_tag_name_oneof: "UL"
  }
```

Nevertheless, even though the AMP plugin is reporting it as a validation error, it is not removing it. Also, the markup is not causing validation errors in the AMP Validator. So I'm confused.

westonruter: I think what you have is correct here, and there is a bug with the AMP plugin. Namely, because the parent `nav` does not have a `toolbar` attribute, this constraint should not be applying.

westonruter: I've opened a fix: https://github.com/ampproject/amp-wp/pull/2926

laurelfulford:
Thanks for catching and reporting that, @westonruter!

A cleaned up version of the AMP menu was added in #158, so I'm going to close this PR.

PR #124, opened 2019-07-24
Style Newspack homepage block headers in theme

laurelfulford:
### All Submissions:

I thought I'd merged something like this ages ago! 

This PR makes sure the section headers in the Newspack block use the styles from the theme, both on the front-end and in the editor. 

@jeffersonrabb and I talked about not using the block's classes in the theme if at all possible and going with something generic, but I've been stuck on finding something that made sense in this case for so long, I've put off getting this style in. 

Right now, the section title does use an `h2` and will inherit the theme's styles/fonts for that. But the way we styled it in the mockups -- with a special alternative header style -- needed a special class. I wanted to leave the block class in the theme more related to its containing block than its appearance (it may not always use a 'special header style') so the best approach seemed to be to use the block's existing class in the theme. 

Another approach would be to remove the class altogether and just use the HTML hierarchy (`.wp-block-newspack-blocks-homepage-articles > h2`) but I think this is less clunky, and will make it easier to customize the blocks on a theme-by-theme (or style-variation-by-style-variation) basis.

1. Set up a homepage block, and add a section title.
2. Confirm it's appearance -- it should be scaled correctly, but just black sans-serif text.
3. Apply PR and run `npm run build`.
4. Confirm that the block is now using the mocked up styles, both on the front-end and in the editor:

image

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -50,6 +50,7 @@ function newspack_custom_colors_css() {
 		.comment .comment-metadata .comment-edit-link:hover,
 		#colophon .site-info a:hover,
 		.widget a, .widget a:visited,
+		.accent-header, .article-section-title,
 		.entry .entry-content .wp-block-button.is-style-outline .wp-block-button__link:not(.has-text-color),
 		.entry .entry-content > .has-primary-color,
 		.entry .entry-content > *[class^="wp-block-"] .has-primary-color,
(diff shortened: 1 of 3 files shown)


laurelfulford:
D'oh! Yes, good catch!

There's some other issues in the custom colours in the editor (the block post title will inherit them when they shouldn't; the author name won't when they should...); I'll tackle that separately, but the section header is fixed in c5a38cf.

laurelfulford:
Thanks for the review and for the reminder about the colours! I'll aim to do a solid sweep for that next week and catch all the bits I've missed :)

PR #134, opened 2019-07-29
Fix up a couple visual issues with Infinite Scroll

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Infinite scroll isn't available in a lot of places in the theme -- basically, the blog posts page (which isn't a huge focal point when paired with custom blocks for the static front page), and the archive pages. 

But right now, it has a couple minor visual hiccups:

* On the archive pages, if you load most posts, they're formatted incorrectly (they're using the wrong template part). 
* On each new 'page' of posts, there's more space than normal (mostly noticeable on mobile). 

### How to test the changes in this Pull Request:

1. Turn on infinite scroll (Dashboard > Jetpack > Settings > Writing, and under "Infinite Scroll" select "Load more posts as the reader scrolls down" or "Load more posts in page with a button" -- the theme has been set to always load new posts on click, I'll also be updating that in a future PR).
2. Visit on the archive pages that has a lot of posts (if you have a lot of test posts, there's a good chance that `www.website.com/category/uncategorized/` will meet the needs; otherwise, `www.website.com/author/yourusername` is another good bet). 
3. Scroll to the bottom and click 'Older Posts'
4. Note that the next set of posts loads each post's full content (in the "Archive page before" screenshot, this starts at "Template: Excerpt (Generated)").
5. Also note the space before the newly loaded page worth of posts is wider than usual.
6. Apply the PR. 
7. Confirm that the post content is no longer loading on the second 'page' of posts, and the spacing is more consistent. 
8. Double-check the blog posts page (the page you've set to display the blog posts) still displays full posts when you click 'Older Posts' (I'll be iterating on this page in the future).

**Archive page before:**

image

**Archive page after:**
Screen Shot 2019-07-29 at 2 33 16 PM

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #135, opened 2019-07-29
Update Infinite Scroll to use load on scroll when available

laurelfulford:
### All Submissions:

This theme doesn't showcase infinite scroll, but it should work in a predictable way. 

The settings originally set it up to always use click to scroll (so the older posts button is there), but anecdotally this confuses folks, especially since the Jetpack settings always offer both options when a theme supports infinite scroll, regardless what 'type' it sets:

image

However, this update also makes sure infinite scroll is switched to 'click' when a page has footer widgets -- otherwise, it's impossible to get to those widgets, and that can be frustrating.

It also switches to click to scroll on small screens if you have a sidebar widget for similar reasons to the above -- on smaller screens the sidebar will wrap below the content and also be difficult to get to if posts are loading on scroll. 

The theme doesn't 'show off' infinite scroll in any way, and it's unlikely that the scroll-to-load functionality will actually be available on many sites due to footer widgets, but I want to make sure it works in a correct, not-confusing way.

This PR compliments work in #134, which is to make sure infinite scroll also displays correctly. 

2. Remove all widgets. 
3. Turn on infinite scroll (Dashboard > Jetpack > Settings > Writing, and click 'Load more posts as the reader scrolls down'). 
4. Navigate to the blog posts page and confirm that more posts load as you scroll down the page.
6. Return to the Customizer and add a footer widget.
7. Navigate to the blog posts page and confirm that it now displays an 'Older Posts' button.
8. Return to the Customizer; remove the footer widget and add a sidebar widget. 
9. On a laptop/desktop computer, navigate to the blog posts page and confirm that it loads more posts as you scroll. 
10. On a mobile device, navigate to the blog posts page and confirm it displays the 'Older Posts' button. (Note: this must be a real mobile device or Mac's iOS simulator and not just a narrow browser window; it uses Jetpack's [`jetpack_is_mobile()`](https://github.com/Automattic/jetpack/blob/3699f34bb90a2b0a2af221042dacacf777238d2b/class.jetpack-user-agent.php#L13) function, which checks for actual devices). 

### DIFF
--- a/inc/jetpack.php
+++ b/inc/jetpack.php
@@ -16,7 +16,6 @@ function newspack_jetpack_setup() {
 	add_theme_support(
 		'infinite-scroll',
 		array(
-			'type'      => 'click',
 			'container' => 'main',
 			'render'    => 'newspack_infinite_scroll_render',
 			'footer'    => 'page',
@@ -70,6 +69,18 @@ function newspack_infinite_scroll_render() {
 		get_template_part( 'template-parts/content/content' );
 	}
 }
+
+/**
+ * Toggle between click to load or scroll to load for infinite scroll.
+ */
+function newspack_toggle_infinite_scroll_type() {
+	if ( is_active_sidebar( 'footer-1' ) || ( ( jetpack_is_mobile( '', true ) && is_active_sidebar( 'sidebar-1' ) ) ) ) {
+		return true;
+	}
+	return false;
+}
+add_filter( 'infinite_scroll_has_footer_widgets', 'newspack_toggle_infinite_scroll_type' );
+
 
 /**
  * Remove Jetpack Share icons from standard location so they can be moved.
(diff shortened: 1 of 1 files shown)


inc/jetpack.php
```diff
@@ -71,6 +70,18 @@ function newspack_infinite_scroll_render() {
 	}
 }
 
+/**
+ * Toggle between click to load or scroll to load for infinite scroll.
+ */
+function newspack_toggle_infinite_scroll_type() {
+	if ( is_active_sidebar( 'footer-1' ) || ( ( jetpack_is_mobile( '', true ) && is_active_sidebar( 'sidebar-1' ) ) ) ) {
```
jeffersonrabb: Non-blocking suggestion: the logic in this line is a bit difficult to parse. I would suggest breaking it out to separate lines for each case:

```
if ( is_active_sidebar( 'footer-1' ) ) {
  return true;
} 
if ( jetpack_is_mobile( '', true ) && is_active_sidebar( 'sidebar-1' ) ) {
  return true;
}
```

laurelfulford: Ah, good call, Jeff! I made a quick update for that in 14ce829.

yogeshbeniwal:
@laurelfulford May i know if it still work? Checked on singe post with no widget in Footer, seems to be not working. Is it because of copyright text and privacy policy link.

PR #126, opened 2019-07-25
Fixes sidebar spacing

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

The updates made for the archive or search styles caused a regression in the content/sidebar spacing on the single posts. This PR:

* Fixes the spacing issue
* Makes the Sass a bit more concise (originally the widths were repeated twice because the markup isn't the same; this PR puts them together)
* Reduces the spacing width a bit for tablet-sized screens. 

Before:

image

After: 

image

### How to test the changes in this Pull Request:

1. Apply the PR and run `npm run build`.
2. Confirm that the spacing between content and sidebar on single posts is improved.
3. Confirm that there isn't a regression/change in this spacing on the archive/search results pages. 

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #136, opened 2019-07-29
Fix search sidebar styles

laurelfulford:
### All Submissions:

I thought it was so careful not to introduce any regressions in #126, but I still broke the search results page! 🤦‍♀ 

This PR makes sure the sidebar goes back to acting like a sidebar again on the search results page. 

**Before:**

image

**After:**

image

1. Make sure you have a sidebar widget.
2. Run a search and view the search results page.
3. See that huge wide single column of content, and a sidebar hidden far below, and despair. 
5. View the search results page again; be filled with joy at the sight of it displaying correctly again. 

### Other information:

### DIFF
--- a/sass/layout/_layout.scss
+++ b/sass/layout/_layout.scss
@@ -54,6 +54,7 @@
 
 .archive #main,
 .blog #main,
+.search #main,
 .page:not(.newspack-front-page) #main,
 .single-post .main-content {
 	@include media(tablet) {
@@ -63,6 +64,7 @@
 
 .archive #secondary,
 .blog #secondary,
+.search #secondary,
 .page:not(.newspack-front-page) #secondary,
 .single-post #secondary {
 	@include media(tablet) {


laurelfulford:
Thanks Jeff! 

> This is unrelated to this PR but when I add Search to the top of the sidebar, this is the result:

Is the issue the doubled search icon, or the vertical spacing?

For the icon: I'm betting with the search always available in the header, it's probably not something many sites will also add to the sidebar. If a site chooses to, that's something entirely under their control (for good or bad!). The only way I can think to avoid any kind of duplication is to remove the search from the header.

If it's the vertical spacing, that should be fixed with this PR: #137.

jeffersonrabb:
> If it's the vertical spacing, that should be fixed with this PR: #137.

My question was about the vertical spacing not the duplication (agreed it's a choice most will avoid), so 137 is the answer (my fault for looking at the PRs out of order!)

PR #133, opened 2019-07-29
Updates post title style in editor to match the front-end

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR updates the title in the editor to match the front end.

For the pages, this is done simply in the style-editor-base.scss.

For the posts, it's a bit more complicated: there is a `.post-type-post` class on the body in the editor that can be used, but to target the body class, the editor styles have to be enqueued in a different way. 

For the post title, I renamed and reused the existing style-editor-static-front-page.scss file, which also needed to target a body class, and enqueued it on all pages rather than just the static front page. That way it can be used for the post title sizing, as well as any exceptions we run into down the road. 

Post - front end:

image

Page - front-end:

image

Post - editor (before):

image

Post - editor (after):

image

Page - editor (before):

image

Page - editor (after):

image

### How to test the changes in this Pull Request:

1. Apply the PR and run `npm run build`.
2. Compare a page title on the front end to how it's sized in the editor.
3. Compare a post title on the front end to how it's sized in the editor. 

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #139, opened 2019-07-30
Clean up image block styles

laurelfulford:
### All Submissions:

This PR fixes up and adds some styles for the image blocks, in the editor and on the front-end.

2. Copy-paste [this test data](https://cloudup.com/cT3iW9y49cq) into a post, in the Code Editor view -- it includes images with captions, with no alignment, aligned left, right, centred, wide and full.
3. Do a visual review on the front-end and check for any weirdness.
4. Do a visual review in the editor and check for any weirdness. 
5. Switch templates to the 'Article feature' and do a visual review on the front-end.

**General notes:**

* Wide and full images should only display as such using the 'Article feature'.
* Right-aligned images should only pull out of the content area using the 'Article feature'.
* Left-aligned images should always pull out of the content area to the left, but less so with the default article template.

**Default template - images aligned left and right**
image

**Default template - images aligned wide and full**
image

**Article feature - images aligned left and right:**
image

**Article feature - images aligned wide and full:**
image

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -572,7 +572,7 @@
 		}
 
 		figcaption {
-			border-bottom: 1px solid #ccc;
+			border-bottom: 1px solid $color__border;
 			margin: auto;
 			max-width: 780px;
 		}
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -539,22 +539,7 @@
 		}
 
 		.aligncenter {
-
-			@include media(tablet) {
-				margin: 0;
-				width: 100%;
-
-				img {
-					margin: 0 auto;
-				}
-			}
-
-			@include media(desktop) {
-
-				img {
-					margin: 0 auto;
-				}
-			}
+			margin: 0 auto;
 		}
 
 		&.alignfull img {
@@ -564,6 +549,10 @@
 
 		&.alignwide {
 			max-width: 100vw;
+		}
+
+		figcaption {
+			border-bottom: 1px solid #ccc;
 		}
 	}
 
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -152,6 +152,20 @@
 		padding: 20px 30px;
 	}
 
+	//! Captions
+	.wp-block-audio figcaption,
+	.wp-block-video figcaption,
+	.wp-block-image figcaption,
+	.wp-block-gallery .blocks-gallery-image figcaption,
+	.wp-block-gallery .blocks-gallery-item figcaption {
+		font-family: $font__heading;
+		font-size: $font__size-xs;
+		line-height: $font__line-height-pre;
+		margin: 0;
+		padding: ( $size__spacing-unit * .5 );
+		text-align: center;
+	}
+
 	//! Audio
 	.wp-block-audio {
 
@@ -555,6 +569,8 @@
 
 		figcaption {
 			border-bottom: 1px solid #ccc;
+			margin: auto;
+			max-width: 780px;
 		}
 	}
 
@@ -630,20 +646,6 @@
 		figcaption a {
 			color: #fff;
 		}
-	}
-
-	//! Captions
-	.wp-block-audio figcaption,
-	.wp-block-video figcaption,
-	.wp-block-image figcaption,
-	.wp-block-gallery .blocks-gallery-image figcaption,
-	.wp-block-gallery .blocks-gallery-item figcaption {
-		font-family: $font__heading;
-		font-size: $font__size-xs;
-		line-height: $font__line-height-pre;
-		margin: 0;
-		padding: ( $size__spacing-unit * .5 );
-		text-align: center;
 	}
 
 	//! Separator
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -29,6 +29,7 @@
 		width: 100vw;
 	}
 
+	&.wp-block-image .alignleft, // Gah image block, why are you marked up like this??
 	&.alignleft {
 		/*rtl:ignore*/
 		float: left;
@@ -53,6 +54,7 @@
 		}
 	}
 
+	&.wp-block-image .alignright,
 	&.alignright {
 		/*rtl:ignore*/
 		float: right;
@@ -540,6 +542,7 @@
 
 			@include media(tablet) {
 				margin: 0;
+				width: 100%;
 
 				img {
 					margin: 0 auto;
@@ -943,6 +946,7 @@
 .single.post-template-default.has-sidebar {
 	.entry .entry-content > *,
 	.entry .entry-summary > * {
+		&.wp-block-image .alignleft,
 		&.alignleft {
 			margin-left: 0;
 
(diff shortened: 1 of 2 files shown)


laurelfulford:
Made updates to add borders to captions.

sass/style-editor-base.scss
```diff
@@ -329,6 +329,15 @@ figcaption,
 	}
 }
 
+/** === Image === */
+
+.wp-block-image {
```
jeffersonrabb: [Question] Is there any way to adjust Editor styling based on the selected Template? I suspect Full Width image blocks in the default template will be a source of confusion—users will see the full bleed style in the editor and be confused about why the same does not occur in the front end.

sass/blocks/_blocks.scss
```diff
@@ -955,6 +947,7 @@
 				margin-left: #{ -4 * $size__spacing-unit };
 			}
 		}
+		&.wp-block-image .alignright,
```
jeffersonrabb: [Question] Do you think the mismatch of left/right -aligned images in the default template might track as an error? Seems clear you did this intentionally so I'm ready to follow your lead but I just wanted to confirm.

sass/blocks/_blocks.scss
```diff
@@ -563,6 +551,10 @@
 		&.alignwide {
 			max-width: 100vw;
 		}
+
+		figcaption {
```
jeffersonrabb: I like the bottom-border in all but the full-width alignments. Something about the rule stretching from edge to edge bothers me. How would you feel about removing the rule in that alignment, or limiting the caption width?

sass/blocks/_blocks.scss
```diff
@@ -29,6 +29,7 @@
 		width: 100vw;
 	}
 
+	&.wp-block-image .alignleft, // Gah image block, why are you marked up like this??
```
jeffersonrabb: At some widths I get very narrow text columns around the floated alignments. Could we apply a percentage `max-width` to left/right to avoid this?

Screen Shot 2019-07-31 at 12 43 11 PM

sass/blocks/_blocks.scss
```diff
@@ -563,6 +551,10 @@
 		&.alignwide {
 			max-width: 100vw;
 		}
+
+		figcaption {
+			border-bottom: 1px solid #ccc;
```
jeffersonrabb: I found a bunch of `#ccc` in the codebase — good candidate for a sass variable?

sass/style-editor-base.scss
laurelfulford: There isn't a class added based on template automatically, but we might be able to riff on the approach [used for the front page](https://github.com/Automattic/newspack-theme/blob/master/functions.php#L313-L324) and add our own based on template. It's something I can dig into. 

Do you think it will be confusing that those options -- wide and full -- are still there?

laurelfulford: Adding a follow up that it does look like it will be doable, but I'd like to handle it in a separate PR -- #133 clashes a lot with the same files I'll need to edit, and I'm anticipating a mess of conflicts if I do it now. 

Let me know if you think it'll be an issue!

sass/blocks/_blocks.scss
laurelfulford: I included it to create some visual interest, but I was wondering about it being confusing, too -- your initial reaction doesn't sell keeping it to me. 😅 I've removed it in 0a83526, and we can do the same in the editor once we have the ability to differentiate between templates.

sass/style-editor-base.scss
jeffersonrabb: Sounds good, but your point about wide/full both being available options is a good one. Let's discuss this before you get too far into it, but definitely agreed it doesn't belong in this PR.

sass/blocks/_blocks.scss
jeffersonrabb: Another alternative would be to extend the right-aligned blocks just as far to the right. Is that also a viable option or is there a good reason not to go that way?

sass/blocks/_blocks.scss
laurelfulford: I tried adding a max-width to the caption in 29082ae, so it's never wider than the content area width -- can you give that a peek and let me know what you think!

I'm a little worried removing it entirely will look too inconsistent, but I'd be open to it if this also looks weird.

sass/blocks/_blocks.scss
laurelfulford: I've added a max-width in efbbdd2; the one downside is that it was pretty hacky/annoying to pull of in the editor.

sass/blocks/_blocks.scss
laurelfulford: Yeah, it's getting that way! I'll handle that in a separate PR, just to keep this one from getting too muddled.

laurelfulford: Oh, there was already a variable using this colour! Have updated these changes to use `$color__border` in b772c64, and tackled the others in a separate PR.

sass/blocks/_blocks.scss
jeffersonrabb: I don't think we need to worry about parity on this point in the editor.

sass/blocks/_blocks.scss
jeffersonrabb: Personally I think this looks great! I also like that long captions will be constrained to the column width and will break to multiple lines if needed. Super-wide captions beneath full-bleed images look junky IMO. If this change makes you uneasy, though, we could solicit other opinions.

sass/style-editor-base.scss
laurelfulford: > Let's discuss this before you get too far into it, but definitely agreed it doesn't belong in this PR.

Sounds good!

sass/blocks/_blocks.scss
laurelfulford: The issue is the right sidebar 😕 In other themes, we've used JavaScript to check where a sidebar ends and have pulled stuff out only when it was underneath it (Twenty Seventeen is a good example of this, like the right-aligned quote [here](https://twentyseventeendemo.wordpress.com/2017/01/02/welcome-to-our-blog/)), but I'm leery of going that route and having to figure out how to get it working with AMP.

It also has the issue of not previewing nicely in the editor -- there's no way to figure out what will and won't 'pull out' of the content from there. I think just keeping them both within the content width is the safest approach.

sass/blocks/_blocks.scss
laurelfulford: Awesome! I'm good with this approach :) I wanted to avoid creating inconsistencies (some with borders, some without) because that kind of stuff can be read as 'bug'.

> I also like that long captions will be constrained to the column width and will break to multiple lines if needed. 

Excellent point!

sass/blocks/_blocks.scss
laurelfulford: It works _pretty_ closely, but yeah -- I find the editor can be kind of an inconsistent experience for breakpoints, since it's using the whole window, including both editor sidebars. So the content can be displaying at what you'd think was a mobile width, but the browser window is at a tablet -- or even desktop -- breakpoint, so so are the styles.

I was mostly excusing my terrible, terrible CSS 😄

laurelfulford:
Thanks Jeff!

sass/style-editor-base.scss
laurelfulford: I made an issue for this in #147.

PR #149, opened 2019-08-01
Add custom typography to editor

### LINKED ISSUE #45: Create editor styles for custom typography
The custom fonts added in #39 should also be included in the preview editor. So both the font CSS and styles to enqueue the fonts themselves need to be added there.

laurelfulford:
### All Submissions:

This PR enqueues the custom typography styles and fonts in to the editor, to match the behaviour on the front-end. 

It also reworks how the fonts were originally set up a bit to mirror how the colours are set up, to try to reduce repetition in the editor code. The big changes were:

* Switching the variable that holds the font styles from an array to a string.
* Removing the `<style></style>` tags in the function and using `wp_add_inline_style()` to add the block of CSS to the editor.
* Removing the `<link...>` from the fonts, and just returning the URL and using `wp_enqueue_style()` to add them to the front-end and the editor.

Fixes #45 .

1. Apply the PR.
2. In the Customizer, add at least one outrageous font from a third-party source to the Font Provider URL (so it's easier to tell if it's working).  I used //fonts.googleapis.com/css?family=Beth+Ellen|Saira+Stencil+One&display=swap.
3. Assign a Header and Body font -- from the above, you could use "Beth Ellen" and "Saira Stencil One".
4. Publish changes.
5. Confirm custom fonts are displaying on the front end. 
6. Edit a post and confirm they're being used in the editor.
7. Try a few variations of the above -- I'd recommend using both "Font Provider" fields, or only the second one; only having a header or body font, etc. 

Note: There will definitely be elements that the fonts aren't being applied to due to errors in the selectors -- I personally think we should handle those in a separate PR, but please feel free to note them here. Both the fonts and colours functionality need a careful sweep one the theme development slows down a bit, to catch all the stuff that feel through the cracks. 

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -410,7 +410,7 @@ function newspack_typography_css_wrap() {
 	?>
 
 	<style type="text/css" id="custom-theme-fonts">
-		<?php echo esc_html( newspack_custom_typography_css() ); ?>
+		<?php echo wp_kses( newspack_custom_typography_css(), '' ); ?>
 	</style>
 
 <?php
--- a/functions.php
+++ b/functions.php
@@ -266,6 +266,15 @@ function newspack_scripts() {
 		wp_enqueue_script( 'newspack-amp-fallback', get_theme_file_uri( '/js/amp-fallback.js' ), array(), '1.0', true );
 		wp_localize_script( 'newspack-amp-fallback', 'newspackScreenReaderText', $newspack_l10n );
 	}
+	// Load custom fonts, if any.
+	if ( get_theme_mod( 'custom_font_import_code', '' ) ) {
+		wp_enqueue_style( 'newspack-font-import', newspack_custom_typography_link( 'custom_font_import_code' ), array(), null );
+	}
+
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks for the review!

> I don't fully understand the reasoning behind the change of $css_blocks from an array to string concatenation—was this just a style preference or did it resolve a problem? It looks like it works equally well either way, so I have no problem with it - just curious.

I mostly did it because it was easier to wrap my head around getting it to work, to mimic how it was set up for the custom colours :) So it'd be fine to switch it back -- it was just the quickest way I could get it functional.

jeffersonrabb:
> I mostly did it because it was easier to wrap my head around getting it to work, to mimic how it was set up for the custom colours :) So it'd be fine to switch it back -- it was just the quickest way I could get it functional.

Nope, I'm fully comfortable with the way you went. I was just curious if you had uncovered a flaw in the approach, since I use it often elsewhere.

PR #151, opened 2019-08-01
Update approach wide and full styles

### LINKED ISSUE #150: Wide and full blocks go offscreen with default template + no sidebar
**Describe the bug**
When using the single post's default template with no sidebar, wide and full-aligned blocks go off-screen. Left-aligned blocks -- which 'stick out' of the content area -- also get cut-off. 

Originally reported in #146; there are some more good screenshots of the issue there.

**To Reproduce**
1. Add a wide and full block to a post.
2. Make sure it's using the 'default' template.
3. Navigate to the customizer and remove all sidebar widgets.
4. View published version of your post with wide and full blocks.
5. Adjust the browser window size -- it gets better/worse at different breakpoints.

**Screenshots**

Wide and full:
image

Left aligned: 
image

**Expected behavior**

One of three options:
* The content should be centred when there are no sidebars (like the 'article feature' templtae)
* The content should remain to the left, and the wide and full blocks should not run outside of the content area (acting like they do when there is a sidebar).
* The content should remain to the left, and the wide and full block should display as wide and full, but adjust for the offset content area and not get cut off. 

**Edited to add**

Left-aligned blocks.

laurelfulford:
### All Submissions:

This PR reworks the wide and full styles a bit -- they were originally set up as the default, but then overridden on a template basis. But in reality, so far only the 'article feature' template uses the wide and full styles --  the default post and page styles don't, and the archive and search views only show excerpts, not full posts.

This PR inverts this, so the wide and full styles, and pulling the left and right alignments outside of the content area, is only applied to the article feature template.

Down the road, we could also add a 'page' template that would allow you to do the same thing there, but I think I'll hold off and see if we get a request for this.

Closes #150  

2. Copy and paste [this test content](https://cloudup.com/cnGd7pFmIXh) into the Code Editor view of the editor -- it include example of all the block variations that have changed: wide, full, left and right, plus image, cover, columns, pull quote and table, since their full-width styles have all be moved. 
3. Test a single post with the default template and a sidebar widget. All the content should stay 'inside' of the content area.
4. Test a single post with the default template and without a sidebar widget. All the content should still stay 'inside' of the content area.
5. Test a single post with the 'feature article' template. Left and right aligned blocks should pop outside of the content area a bit on larger browser windows (but move over as you scale the browser window down), and wide and full-width blocks should extend outside of the content area.

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -13,22 +13,6 @@
 		margin-bottom: 0;
 	}
 
-	&.alignwide {
-		@include media( tablet ) {
-			margin-left: calc(25% - 25vw);
-			margin-right: calc(25% - 25vw);
-			max-width: 100vw;
-		}
-	}
-
-	&.alignfull {
-		margin-left: calc(50% - 50vw);
-		margin-right: calc(50% - 50vw);
-		max-width: 100vw;
-		position: relative;
-		width: 100vw;
-	}
-
 	// When the image block is aligned left or right, the markup changes,
 	// making a slighly different selector necessary.
 	&.wp-block-image .alignleft,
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks!

PR #142, opened 2019-07-31
Move some of the more opinionated default styles out of the 'base' stylesheet

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

I noticed when starting to experiment with stylesheet styles that matched the mockups that there were a few things in the default styles that would have to be overridden in each and everyone one:

- the 'accent' headers
- the button-like styles on all the tertiary navigation
- the backgrounds on the category links

This PR starts the process of moving those styles into a separate SCSS file that isn't included in the other style packs, and it moves over the 'accent' header styles. 

It also removes some of the placeholder styles from the "Style 1" style pack, to make testing easier. 

I will circle back and tackle the other more opinionated styles in separate PRs.

**Accent header with the default stylepack:**
image

**Accent headers with Style Pack 1:** (which just changes the font and background colour now)
image

### How to test the changes in this Pull Request:

1. Apply the PR and run `npm run build`.
2. Confirm that the 'accent' headers continue to look correct. These can be tested on the homepage, in the 'section header' of the homepage block, and as the sidebar widget's titles.
3. Navigate to Customizer > Style Packs and switch to Style 1 (note this will not preview correctly due to #52). 
4. On the front end, confirm that this header is no longer all caps, blue, or has a bottom border. This gives us a cleaner slate with this element to customize for the other style packs. 

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #152, opened 2019-08-02
Move more opinionated 'tertiary menu' styles to the default style.scss.

laurelfulford:
### All Submissions:

This is a continuation of work started in #142: pulling some of the more opinionated default styles into a separate stylesheet, so they don't need to be overridden in every other style variation.

This PR moves the 'tertiary menu' styles (the button-like links on the right) into a separate stylesheet.

image

2. Visually confirm the tertiary menu looks correct -- it is only applied when the header does not use the solid background or the short styles.
3. Navigate to the Customizer and switch to Style 1, and confirm that the links no longer have backgrounds.

### Other information:

### DIFF
--- a/sass/navigation/_menu-tertiary-navigation.scss
+++ b/sass/navigation/_menu-tertiary-navigation.scss
@@ -42,25 +42,6 @@ nav.tertiary-menu {
 	}
 }
 
-body:not(.header-solid-background):not(.header-simplified) {
-	nav.tertiary-menu {
-		a {
-			background-color: lighten($color__text-light, 45%);
-			color: $color__text-main;
-			@include button-transition;
-			border-radius: 5px;
-			font-size: $font__size-sm;
-			font-weight: 700;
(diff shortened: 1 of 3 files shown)


laurelfulford:
Ah dang! It's not expected, but it looks to be a similar-ish issue to #52 -- the preview is loading both style.css (the default) and style-1.css (the style pack style) in the Customizer, though it only loads one on the front-end:

Screen Shot 2019-08-03 at 1 08 35 PM

Unfortunately it's a lower priority to fix, compared to finishing the theme's front end, but I'll try to prioritize it once the theme is more complete.

jeffersonrabb:
> Unfortunately it's a lower priority to fix, compared to finishing the theme's front end, but I'll try to prioritize it once the theme is more complete.

Agreed. A very minor problem, especially given our aspiration to configure the theme from the plugin.

laurelfulford:
Thanks for the review, Jeff!

PR #159, opened 2019-08-06
Make sure default style is properly dequeued

laurelfulford:
### All Submissions:

I noticed when working on the style variations that the default style.css wasn't properly being dequeued when you used a non-default style variation. 

This PR fixes that. 

1. Navigate to Customizer > Style Pack, and pick 'Style 1'.
2. Publish changes.
3. View source -- note you can still find both `newspack-theme/style.css` and `newspack-theme/styles/style-1.css` -- only the latter should be there.
4. Apply the PR.
5. Confirm that `newspack-theme/style.css` is no longer being loaded.
6. Navigate to Customize > Style Pack and change back to the default.
7. Confirm that the default styles load correctly.

### Other information:

### DIFF
--- a/classes/class-newspack-style-packs-core.php
+++ b/classes/class-newspack-style-packs-core.php
@@ -64,20 +64,27 @@ class Newspack_Style_Packs_Core {
 	 * Gets active Style Pack and loads appropriate stylesheet.
 	 */
 	public function init() {
-		$this->style = get_theme_mod( 'active_style_pack' );
+		$this->style = get_theme_mod( 'active_style_pack', 'default' );
 		if ( is_customize_preview() ) {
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks @claudiulodro!

> The new look is probably the correct version, but just a heads-up, the article headings look different after making this change:

Yup! The border is only used for the default style pack, and the all-caps not for all of them, so I figured it'd be easier to apply them when needed, than having to override them repeated :)

laurelfulford:
🤦‍♀ Just realized my comment was not for this change specifically. I should have said, the border and header are being inherited from style.css, instead of being in style-1.css, so it's correct that they're not displaying anymore.

PR #161, opened 2019-08-07
Remove 'smart' text colours with solid background colours

laurelfulford:
### All Submissions:

Originally Twenty Nineteen included styles so when you set a solid background on, say, a pullquote or paragraph block, it would automatically change the block's default text colour to white.

This worked well for that theme because it only allowed customizing the hue; the white background colour would always have sufficient contrast. 

In the Newspack theme, you can pick a lot more variety of colours to be used as those defaults, with varying saturation -- white may not always have sufficient contrast. I was also running into issues getting these styles to work with our theme's styles.

This PR removes those styles; every block that lets you pick a background colour also lets you change the foreground, so you can still make sure blocks with background colours have sufficient contrast. 

**Before:**

image

**After:**

image

1. Copy and paste this [test content](https://cloudup.com/cKLk8i6HS3D) into a post.
2. Save and publish and view both in the editor and on the front end; the blocks using a 'default' colour (one set by the theme) automatically use white text, even though they haven't been assigned white text.
4. Confirm that the text in all the blocks matches the rest of the body text. 

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -759,43 +759,6 @@
 		font-size: $font__size-xl;
 	}
 
-	//! Custom background colors
-	.has-primary-background-color,
-	.has-primary-variation-background-color,
-	.has-secondary-background-color,
-	.has-secondary-variation-background-color {
-
-		// Use white text against these backgrounds by default.
-		color: $color__background-body;
-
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks!

PR #163, opened 2019-08-07
Fix positioning issue for search icon + AMP

laurelfulford:
… turns out re-using the AMP ones won't work for getting the menu in the right location.

### All Submissions:

Our AMP mobile menu setup uses `toolbar-target`, which helps you to avoid duplicating menu code by having AMP copy it from its original location in the mobile sidebar, to somewhere else on the site. This new location is determined by ID, and requires you to add empty HTML elements to the page.

I had hoped we could use these empty `divs` to help position the search icon next to the menus, but didn't consider that the menu might not be copied exactly where I expected it to be.

In other words, when you enable AMP with certain header settings, the search icon appears in the wrong spot: 

**With AMP on (incorrect):**
image

image

**With AMP off (correct):**

image

image

This PR fixes this issue by (unfortunately!) adding yet another `div` and reusing the existing styles. 

1. Under Customizer > Site Identity, check only 'Short Header'.
2. With AMP enabled, view the site and see the search icon is between two menus (tertiary and primary).
3. Under Customizer > Site Identity, check both 'Short Header' and 'Center Logo'
4. With AMP enabled, view the site to see the search is in front of the primary navigation on the right.
5. Apply the patch and run `npm run build`. 
6. Repeat steps 1-4 and confirm the menu now appears on the end.
7. Confirm that each variation of the header (just 'Short', and 'Short' + 'Centered Logo') look correct with AMP turned off. 

### Other information:

### DIFF
--- a/header.php
+++ b/header.php
@@ -105,18 +105,23 @@
 				// Short header:
 				if ( true === $header_simplified ) :
 				?>
-					<div id="site-navigation">
-						<?php
-						if ( ! newspack_is_amp() ) {
-							newspack_primary_menu();
-						}
+					<div class="nav-wrapper">
+						<div id="site-navigation">
+							<?php
+							if ( ! newspack_is_amp() ) {
+								newspack_primary_menu();
+							}
+							?>
+						</div><!-- #site-navigation -->
 
+						<?php
 						// Centered logo:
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks Jeff!

PR #164, opened 2019-08-07
Move opinionated footer styles to the default style variation

laurelfulford:
### All Submissions:

This PR moves the light grey border at the top of the footer in the default style variation to a style-pack specific CSS file, so it doesn't need to be overridden in each style pack.

Edited to add: it also removes some unneeded specificity from the footer styles.

2. With the 'default' style pack selected, view the footer and confirm there is a light grey border above it.
3. Under Customizer > Style Pack, switch to 'Style 1'.
4. Confirm that there is no border above the footer.

**Default:**

image

**Style 1:**

image

### Other information:

### DIFF
--- a/sass/site/footer/_site-footer.scss
+++ b/sass/site/footer/_site-footer.scss
@@ -9,7 +9,6 @@
 
 	.footer-branding {
 		.wrapper {
-			border-top: 4px solid $color__border;
 			padding-top: $size__spacing-unit;
 		}
 
--- a/sass/site/footer/_site-footer.scss
+++ b/sass/site/footer/_site-footer.scss
@@ -24,29 +24,29 @@
 			flex-wrap: wrap;
 			justify-content: space-between;
 		}
+	}
 
-		.widget {
-			width: 100%;
+	.widget {
+		width: 100%;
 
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks for testing this, Jeff!

> Does this assume presence of Widgets? I tested Style 1 vs. Default with no Widgets and got the top rule in both cases:

There are actually two borders there -- once above the widgets, and one above the 'site info' (the copyright, etc) -- it's less noticeable, but you can see it in the top screenshots, too. 

The 'site info' border is in most of the style packs, so I left it as the 'default' -- but you're right, the border above the widgets will only appear when there are widgets.

Just let me know if any of this doesn't make sense!

laurelfulford:
> However, in the customizer the top rule is present in both styles—is this expected?

This is the issue that is kind-of-like-but-not-like #52: the default style isn't dequeued in the Customizer. So it's probably best not to worry about testing any style pack stuff in there, at least until I have a chance to dig into the code. 

I'll make a different issue for that since it's not quite like #52 and it'll need a different fix!

laurelfulford:
Thanks for the review!

PR #166, opened 2019-08-08
Style pack B: update 'accent' headers

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR starts the process of building out 'Style B' by focusing on the 'accent' headers. This style appear in the Newspack Homepage block, and on post categories:

**Homepage Blocks:**

image

**Single Posts:** 
image

It also 'unsets' this header style, for the sidebar widget titles, and the author bio, both of which use a subtler style in this style pack:

**Sidebar Widget:**

image

**Author Bio:** 
image

### How to test the changes in this Pull Request:

1. Apply the PR and run `npm run build`.
2. View a Homepage Block with a section header; confirm the styles match the above.
3. View the Homepage block in the editor, and confirm the same styles are applied there (note: the border displaying there is removed in #167). 
4. View a single post and confirm that the category style matches the above. 
5. View a single post and confirm that the author bio title matches the above.
6. Add a sidebar widget with a title; confirm that the title matches the above.
7. Navigate to Customize > Colors and change the primary colour.
8. Confirm the new primary colour is being used, both on the front-end and in the editor.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #168, opened 2019-08-08
Style B: Add Pullquote block styles to front-end.

laurelfulford:
### All Submissions:

This PR builds out the pullquote block style used in Style B for the front end (editor styles coming in a future PR). 

~It relies a bit on #166 to test, since that PR will adds the correct font to the style pack.~ (Now merged)

1. Apply this PR and run `npm run build`.
2. Navigate to Customize > Style Packs and switch to 'Style 1' 
3. Set up multiple pullquote blocks using the different variations, or copy-paste [this test content](https://cloudup.com/cVjw03uEOuL) into the code editor view of a post. The variations are:
    * Default style (centred text, quote icon above which uses the 'primary' colour)
    * With 'main' colour (adds a border to the top that matches this colour)
    * With text colour (changes the text colour)
    * With 'solid' style and 'main' colour changed (has background colour and top border; quote icon and border will inherit the text colour). 
    * Each of the above aligned left, right, wide and full.
     
4. View the pullquote styles on the front-end; confirm they match the appearance of the screenshots.
5. Try changing the primary colour in the Customizer -- it should change the quote colour in all of the blocks but the ones with the solid background.
6. Try changing the colour settings of the blocks that are there and verify that they update on the front-end in an expected way.

<details><summary>Screenshots</summary>

image

image

image

image

image
image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

</details>

**Note:** The original plan was to have the default pullquote style be just text, and the 'main' colour style to have the border and icon, with the icon using the main colour you've set. Unfortunately, this block is weird -- that main colour is applied by adding an inline `style="border-color: #ccc` to the markup, and no class. Worse, that inline style is stripped out and replaced by a dynamic class in AMP, so you can't even use the attribute to add the quote, because it's stripped out in AMP (it would be possible to use a partial match for AMP's dynamic class, but it appears to ignore styles set that way). I had assumed assigning a 'main' colour would add a class, like every other colour customization in Gutenberg seems to do. This is why all of the variations include the quote icon, and why the quote icon can't inherit the `main` colour set in the block.

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -194,6 +194,14 @@ function newspack_custom_colors_css() {
 		';
 	}
 
+	if ( 'style-1' === get_theme_mod( 'active_style_pack', 'default' ) ) {
+		$theme_css .= '
+			.entry .entry-content .wp-block-pullquote blockquote:before {
+				color: ' . $primary_color . ';
+			}
+		';
+	}
+
 	if ( true === get_theme_mod( 'header_solid_background', false ) ) {
 		$theme_css .= '
 			.header-solid-background .site-header {
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -326,9 +326,7 @@
 
 		blockquote {
 			border: none;
-			margin-top: $size__spacing-unit;
-			margin-bottom: calc(1.25 * #{ $size__spacing-unit});
-			margin-right: 0;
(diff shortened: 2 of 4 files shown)


laurelfulford:
Thanks Jeff -- good catch! It looks like I missed overriding the left margin. 9203183 should take care of it!

laurelfulford:
Thanks! 😄

PR #175, opened 2019-08-08
First pass at adding a non-AMP mobile menu fallback

laurelfulford:
### All Submissions:

This PR adds a non-AMP version of the mobile menu, to be used as a fallback when AMP is not enabled. 

It inherits the styles from the AMP menu, so it shares some of the visual roughness (odd vertical spacing; missing tertiary menu styles and toggles to dig into different levels). 

It will also need some accessibility improvements, to toggle aria labels, and make sure the `focus` is moved to the sidebar when it's opened. 

I'll follow up on these in future PRs. 

image
 
2. Navigate to Dashboard > AMP > General, and set it to 'Transitional' -- this will make it easier to compare the AMP version of the mobile menu to the non-AMP version by using the 'AMP' button in the Admin Toolbar. 
3. On the front-end, reduce the size of the browser window and toggle the mobile menu open and closed.
4. Switch to the AMP view (using the admin toolbar toggle or by adding `?amp` to the end of the URL), and confirm that the menus have the same appearance. 
6. If you don't have one already, add a really long menu and make sure it scrolls.
5. Test on a real mobile device. 

### Other information:

### DIFF
--- a/header.php
+++ b/header.php
@@ -24,6 +24,8 @@
 <?php
 	if ( newspack_is_amp() ) {
 		get_template_part( 'template-parts/header/mobile', 'sidebar' );
+	} else {
+		get_template_part( 'template-parts/header/mobile', 'sidebar-fallback' );
 	}
 ?>
 
@@ -41,7 +43,7 @@
 		// Header is NOT short:
 		if ( false === $header_simplified ) :
 		?>
-			<div class="top-header-contain">
+			<div class="top-header-contain desktop-only">
 				<div class="wrapper">
 					<div id="secondary-nav-contain">
 						<?php
@@ -67,16 +69,13 @@
 			</div><!-- .top-header-contain -->
 		<?php endif; ?>
 
-
-
-
 		<div class="middle-header-contain">
 			<div class="wrapper">
 				<?php
 				// Centered logo AND NOT short header.
 				if ( true === $header_center_logo && false === $header_simplified ) :
 				?>
-					<div id="social-nav-contain">
+					<div id="social-nav-contain" class="desktop-only">
 						<?php
 						if ( ! newspack_is_amp() ) {
 							newspack_social_menu_header();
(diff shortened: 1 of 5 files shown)


header.php
```diff
@@ -147,12 +147,10 @@
 					</div><!-- .nav-wrapper -->
 				<?php endif; ?>
 
-				<?php if ( newspack_is_amp() ) : ?>
-					<button class="mobile-menu-toggle" on='tap:mobile-sidebar.toggle'>
-						<?php echo wp_kses( newspack_get_icon_svg( 'menu', 20 ), newspack_sanitize_svgs() ); ?>
-						<?php esc_html_e( 'Menu', 'newspack' ); ?>
-					</button>
-				<?php endif; ?>
+				<button class="mobile-menu-toggle" on="tap:mobile-sidebar.toggle">
```
jeffersonrabb: Is there any risk to defining AMP events in non-AMP markup? I suspect not, but would be curious if you have a definitive answer?

laurelfulford: I don't really know. It doesn't seem to cause issues, but it would be something good to run by the AMP folks, since they're probably more familiar the gotchas. 

We are doing this in other places -- like where we toggle the class on the header for the search, the whole opening tag is:

`<header id="masthead" class="site-header hide-header-search" [class]="searchVisible ? 'show-header-search site-header ' : 'hide-header-search site-header'">`

It would be straight-forward to add some more`is-amp` checks around this kind of stuff, if we needed to!

laurelfulford:
Thanks for the review, Jeff!

> The only visual difference I noticed is that in AMP the Close button appears to initially be in focus state, with a light dotted border.

This is a good catch! It falls under my vague 'accessibility improvements' that I still need to add -- when the menu is opened, the focus will switch from the 'Menu' open button, to this close button -- so ultimately in the end this bit will be the same. :)

PR #176, opened 2019-08-08
Style C: Add the styles to make the header overlap.

laurelfulford:
### All Submissions:

This PR adds the styles for the 'overlap' header used in the [Style C](https://cloudup.com/iIgIC67AIBG) style pack.

Note that it doesn't update the menu styles, or anything inside of the content area (outside of adjusting the padding there a bit).

**Mockups:**

image

image

I've included screenshots of different menus with different settings under the cut:

<details><summary><strong>Screenshots</strong></summary>

**Default header settings:** 

image

**Centred Logo:**

image

**Solid Background:**

image

**Short Header:**

image

**Short Header + Solid Background:** 

image

**Short Header + Centred Logo:**

image

</details>

2. Navigate to Customizer > Style Packs, and set it to 'Style 2'. 
3. Test the different header variations - this style is meant to work best with the 'short' header, but it should work with them all:
     * Default (no settings checked)
     * Solid Background
     * Centred Logo
     * Short Header
     * Short Header + Solid Background
     * Short Header + Centred Logo
4. Test the above on smaller screens as well (note: the header overall transitions pretty poorly from desktop to tablet styles -- there are some behaviours in between the desktop and tablet breakpoints that don't display great -- I'll be improving this in a future PR). 

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -261,6 +261,14 @@ function newspack_custom_colors_css() {
 		';
 	}
 
+	if ( 'style-2' === get_theme_mod( 'active_style_pack', 'default' ) ) {
+		$theme_css .= '
+			.site-content #primary {
+				border-color: ' . newspack_adjust_brightness( $primary_color, -40 ) . ';
+			}
+		';
+	}
+
 	if ( true === get_theme_mod( 'header_solid_background', false ) ) {
 		$theme_css .= '
 			.header-solid-background .site-header {
--- a/sass/layout/_layout.scss
+++ b/sass/layout/_layout.scss
@@ -23,7 +23,7 @@
 	overflow: hidden;
 }
 
-#content {
+.site-content {
 	margin-top: #{ 1.5 * $size__spacing-unit };
 
 	@include media( tablet ) {
--- a/sass/site/header/_site-header.scss
+++ b/sass/site/header/_site-header.scss
@@ -88,14 +88,11 @@
 }
 
 // Middle bar
-.middle-header-contain {
-
-	.wrapper {
-		align-items: center;
-		padding: $size__spacing-unit 0;
-		@include media( mobile ) {
-			padding: #{ 2 * $size__spacing-unit } 0;
-		}
+.middle-header-contain .wrapper {
+	align-items: center;
+	padding: $size__spacing-unit 0;
+	@include media( mobile ) {
+		padding: #{ 1.5 * $size__spacing-unit } 0;
 	}
 }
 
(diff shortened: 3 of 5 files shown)


laurelfulford:
Thanks for the review, Jeff!

> When search icon is clicked, the search box overlaps the bottom rule of the header. Presumably we want it to appear fully below the header area?

I was treating it kind of like a dropdown menu and just letting it overlap; I'm okay with moving it, but my one worry is that it might be kind of far and disconnected-feeling (it should probably be moved to the left a bit, too):

image

I might be overthinking that though - what do you think?

It's also worth noting there's a bit of overlap with the short header, too -- so whatever we go with should probably be done for every case:

image

> In the mockups the styling of the Tertiary and Primary menus look identical or similar. I'm seeing a very different text style for Tertiary in testing.

In the mockups for Style C, it's meant to be only showing one menu (the primary menu) and I still need to get the font styles right (they're not uppercase). 

The headers in Style D and E are probably better representatives of what the two menus should look like: https://cloudup.com/iWwXMrBMFGP The primary menu should be aligned left when the tagline is hidden, which is not working right now, but I can tackle that in a future PR (I know I had it working at some point, I must've broken it with all the header changes). 

I'll update Style C's header fonts in a future PR; I think things will look less 'off' then, but I can definitely revisit it if it still seems weird! 

> At browser widths between 600-700px, the logo appears between social and tertiary menus (good) but the vertical spacing is uneven.

At this breakpoint, the mobile menu should be shown (it starts showing at windows 768px wide) -- with AMP enabled you should get it, or once #175 is merged, so that should resolve this.  

I do think there are some points between about 1000px and and 768px that aren't as refined as they could be, though -- it feels like each menu wraps to a new line a little too "soon", and I think it makes the header feel a bit fragile. I plan to explore that in a future PR -- I might just need to increase the tablet break point, and adjust spacing/flexbox settings.

laurelfulford:
> In the mockups the styling of the Tertiary and Primary menus look identical or similar. I'm seeing a very different text style for Tertiary in testing.

This should now be fixed in #190; I'm switching this back to 'Needs Review', since I have a question about the search, and the other remaining issue -- things looking weird on smaller screens -- should be resolved with #175 (non-AMP mobile menu). Just let me know if it's better to leave it as 'Needs changes...' in cases like this!

jeffersonrabb:
> I was treating it kind of like a dropdown menu and just letting it overlap; I'm okay with moving it, but my one worry is that it might be kind of far and disconnected-feeling (it should probably be moved to the left a bit, too):

I think you're right—it would be too far from the icon. Let's leave it and see if this gets any negative reactions.

laurelfulford:
Thanks Jeff! 

> This is definitely the trickiest style, but it don't want to pre-optimize too much. Let's see how it works out in real sites and if there are issues address them later.

That makes sense to me!

PR #192, opened 2019-08-11
Phpstan

szepeviktor:
* [x] I have followed everything :)

@laurelfulford Static analysis is like adding 100 👀

I've fixed all errors on Level 4. You can go higher!

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,0 +1,51 @@
+#$ composer require --dev phpstan/phpstan-shim szepeviktor/phpstan-wordpress
+#$ vendor/bin/phpstan analyse
+
+includes:
+    - phar://phpstan.phar/conf/bleedingEdge.neon
+    - vendor/szepeviktor/phpstan-wordpress/extension.neon
+parameters:
+# TODO    level: max
+    level: 4
+    inferPrivatePropertyTypeFromConstructor: true
+    paths:
+        - %currentWorkingDirectory%/
+    excludes_analyse:
+        - %currentWorkingDirectory%/woocommerce/checkout/
+        - %currentWorkingDirectory%/tests/
+        - %currentWorkingDirectory%/vendor/
+    autoload_files:
+        # Stubs
+        - %currentWorkingDirectory%/tests/bootstrap.php
+        - %currentWorkingDirectory%/tests/functions-stubs.php
+        # Files loaded from functions.php
+        - %currentWorkingDirectory%/inc/back-compat.php
+        - %currentWorkingDirectory%/inc/color-patterns.php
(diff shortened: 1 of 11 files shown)


szepeviktor:
I think `PHPStan\\WordPress\\Composer\\FixWpStubs::php73Polyfill` needs to be executed as core's `is_countable()` shim conflicts with a symfony polyfill.

```json
  "scripts": {
    "php73": "PHPStan\\WordPress\\Composer\\FixWpStubs::php73Polyfill",
```

and run `composer run-script php73`

phpstan.neon.dist
```diff
@@ -0,0 +1,51 @@
+#$ composer require --dev phpstan/phpstan-shim szepeviktor/phpstan-wordpress
+#$ vendor/bin/phpstan analyse
```
claudiulodro: Can you please add these scripts/dependencies and the `php73` script you proposed to `composer.json`? That would make it much easier to get set up and run these tests in the future.

tests/bootstrap.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+define( 'WP_DEFAULT_THEME', '' );
+
+function is_amp_endpoint() { return true; }
```
claudiulodro: Is there a way to do this without needing to redeclare a bunch of functions as stubs? It seems inelegant and difficult to maintain the tests if every function needs to be declared as a stub.

szepeviktor: Static analysis works by figuring out types, so we need all referenced source code:

- Jetpack
- WooCommerce

I though it more sustainable to include 6 function stubs than 2 whole plugin stubs.
What do you think?

claudiulodro: Yeah that makes sense. Thanks for clarifying. The stubs for references to other plugin's functions aren't super difficult to maintain, and we probably won't need to add too many of those. 

What about the `newspack_is_amp` and other `newspack_`-prefixed stubs? The static analyzer should be analyzing `functions.php` and analyze the real functions, shouldn't it? Can't we just load `functions.php`?

phpstan.neon.dist
szepeviktor: Done.

tests/bootstrap.php
szepeviktor: Currently we have these stubs:

- WP core stubs
- functions.php's stubs
- this file

claudiulodro: Ah OK I see now. The WP stubs are loaded from https://github.com/GiacoCorsiglia/wordpress-stubs and this doesn't actually use the real functions. 👍

laurelfulford:
Thank you, @szepeviktor and @claudiulodro!

szepeviktor:
You're welcome.

PR #203, opened 2019-08-11
Style D: Update solid header to use secondary colour

laurelfulford:
### All Submissions:

I've been noodling a bit on how best to handle the dark grey headers shown in the mockups for Style D and E.

One approach would simply be to always set the header to dark grey when the 'solid header' option is chosen for these two style packs, rather than the primary colour. 

The other would be to use the 'secondary' colour instead, so it could be set to a more neutral colour. I've explored that approach here.

I think this could be the most flexible way to do it (it would, for example, let you set the header to other colours, too) but I'd be interested in getting a second opinion as well as the code tested. 

Right now, the non-short headers looks a little weird; I need to explore that further, but wanted to get the approach sorted first.

Any suggestions welcome!

image

2. Navigate to Customize > Style Pack and select 'Style 3'.
3. Set the header to 'Solid Background'; test with both 'Short Header' and without.
4. Navigate to Customize > Colours and try changing the secondary colour; confirm the header changes as well.

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -218,6 +218,10 @@ function newspack_custom_colors_css() {
 
 	if ( 'style-2' === get_theme_mod( 'active_style_pack', 'default' ) ) {
 		$theme_css .= '
+			.site-content #primary {
+				border-color: ' . newspack_adjust_brightness( $primary_color, -40 ) . ';
+			}
+
 			.site-footer {
 				background-color: ' . $primary_color . ';
 				color: ' . $primary_color_contrast . ';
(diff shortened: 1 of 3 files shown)


laurelfulford:
> When I test style 3 the header is the primary color (blue). Is this correct or is it supposed to be gray?

It's suppose to be grey if the default colour is set, or whatever your secondary colour is if you're using custom colours.

It's working for me locally -- I think this might be another case of weird rebasing, so I fixed the conflicts and pushed them up. Can you confirm if it's still showing blue for you?

laurelfulford:
I've updated the header to use a hard-coded grey, and used a variable, `$color__background-dark`. There are a few instances that also use a dark grey background (including this style pack's footer, which I missed), so this will have other uses.

When the header is not set to short, the primary menu bar below it should be a bit darker than the dark grey, when using Style D:

image

Once this PR is merged, I can reproduce similar styles for Style E.

laurelfulford:
Thanks Jeff! Me too - going to spin up a PR for Style E next :)

PR #242, opened 2019-08-14
Custom Colours: Make sure essential items have sufficient contrast

laurelfulford:
### All Submissions:

Right now, it's possible to set any colour as your primary or secondary colour. However, if you pick a really light colour, there are cases where it's used -- like links -- where it won't have a sufficient contrast against the theme's white background.

This PR roughs in one way to approach this: using the existing `newspack_get_color_contrast()` function, check to see if the selected colour has enough contrast against white, and if not, use a mid-grey fallback in cases where it's essential (like text; it's not as dire if decorative elements are low contrast.

This PR in particular just adds this functionality for the accent headers and post meta in the default style pack. If it seems like a good approach, I will work through other areas where this is a problem with additional PRs.

1. Apply the PR.
2. Navigate to Customize > Style Packs and switch to the default.
3. Navigate to Customize > Colors and pick something light for the primary colour -- like yellow.
4. View the homepage blocks; confirm that the accent headers are using grey for the text, rather than yellow. The post meta on the homepage blocks single posts should be doing the same.
5. Switch to a darker custom colour.
6. View the homepage blocks; confirm that the accent headers are using the primary colour for the text.

**Lighter primary colour**

image

**Darker primary colour:**

image

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -18,8 +18,21 @@ function newspack_custom_colors_css() {
 		$secondary_color = get_theme_mod( 'secondary_color_hex', $secondary_color );
 	}
 
+	// Create some variables to use in cases where the colours are against a white background.
+	$primary_color_text   = $primary_color;
+	$secondary_color_text = $secondary_color;
+
 	$primary_color_contrast   = newspack_get_color_contrast( $primary_color );
 	$secondary_color_contrast = newspack_get_color_contrast( $secondary_color );
+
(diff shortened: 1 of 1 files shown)


laurelfulford:
Oh, I like that a lot more! I've made that update in 6877f4f.

Should I do something similar with the other two variables, `$primary_color_contrast` and `$secondary_color_contrast`?

laurelfulford:
Ok cool -- I leave it as is then :) Thanks Jeff!

PR #243, opened 2019-08-14
Block Styles: Add style to columns block to reorder columns

laurelfulford:
### All Submissions:

This PR is for [this issue](https://github.com/Automattic/newspack-blocks/issues/40) in the Newspack Block repository -- I think it makes more sense to have in the theme, or its own plugin; I've started with it in the theme for now, but would appreciate a second opinion on this!

Something that came up a few times during the design feedback was how the articles would be ordered on mobile. By default, the columns would stack left to right, meaning the most important story (in theory) would be on the left.

This isn't always the case, though -- there are layouts where the most prominent, important story would actually be placed in the middle on desktop, but should be at the top on mobile/to screenreaders.

I got into it in more detail, along with accessibility considerations, in the original ticket: https://github.com/Automattic/newspack-blocks/issues/40 But the short of it is, the styles this PR adds reorders the columns on desktop sized screens only, so you technically place your most important stories in the first column, but can style that column to appear as the second or third column, to centre it on larger screens.

1. Apply the patch and run `npm run build`.
2. Copy paste[ this test content](https://cloudup.com/cUZLDreQrup) into the code view of the editor. It has a columns block with three columns, each with an article block.

image

3. Click the 'Change Block Style or Type' button in the editor for the Columns block; you should now see three options for styles: Default, and the horribly named 'Move first column to second' and 'Move first column to third'; the previews are also not great.

image

4. Apply 'Move first column to second' -- the larger first column should now appear in the middle:

image

(Though in the source it's still first; if you shrink down the browser window, it will still be at the top:

image
)

5. 'Move first column to third' will move it to the end -- though this would be more realistically used in cases where you have more than three columns:

image

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -342,6 +342,14 @@ function newspack_editor_customizer_styles() {
 add_action( 'enqueue_block_editor_assets', 'newspack_editor_customizer_styles' );
 
 /**
+ * Enqueue Block Styles Script for the Columns block
+ */
+function newspack_columns_block_style_javascript() {
+	wp_enqueue_script( 'newspack-columns-block-styles-script', get_theme_file_uri( '/js/blocks-columns-style.js' ), array( 'wp-blocks' ), '1.0', true );
+}
+add_action( 'enqueue_block_editor_assets', 'newspack_columns_block_style_javascript' );
+
+/**
  * Determine if current editor page is the static front page.
  */
 function newspack_is_static_front_page() {
--- a/js/blocks-columns-style.js
+++ b/js/blocks-columns-style.js
@@ -1,0 +1,19 @@
+/**
+ * File block-columns-style.js
+ *
+ * Adds block styles to the Columns block, to allow you to reorder columns.
+ */
+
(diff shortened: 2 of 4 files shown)


laurelfulford:
> I suspect this belongs in the Newspack Blocks plugin, but we should dig in to the pros and cons.

I was thinking it could also just be a stand-alone plugin because it doesn't require the other elements of Newspack to work... though it probably doesn't have a ton of use cases outside of that.

I spent too much time waffling about this myself, it's why it ended up dumped in the theme for the short-term :)

laurelfulford:
@jeffersonrabb Looking at building one of the sites the way I imagine, it would also be handy to have a 'border' style for the columns block :) 

Do you have a preference for where these things live (Newspack blocks plugin vs their own plugin)? For the former, I'm just not sure how to set it up within the code that's there.

laurelfulford:
Taken care of in https://github.com/Automattic/newspack-blocks/pull/84 -- thanks @jeffersonrabb!

PR #245, opened 2019-08-14
Style C: Adjust content padding on the front page.

laurelfulford:
### All Submissions:

In the mockups for style C, only the first part of the page -- the first block of articles, or post header -- is narrower. The rest of the content is 1200px wide. 

This style corrects the spacing on the static front page: only the first article block, or first columns block (whatever is the first immediate child of the content area) will be narrower:

image

image

I'll circle back and address the other pages in separate PRs.

2. Navigate to Customize > Style Packs and switch to Style 2.
3. On the homepage, try a few different settings as the first item -- an article block with a couple settings, a category block, a group block. Make sure it's indented more than the rest of the page, and that the rest of the content is 1200px wide on larger screens.

### Other information:

### DIFF
--- a/sass/styles/style-2/style-2.scss
+++ b/sass/styles/style-2/style-2.scss
@@ -11,7 +11,7 @@ Newspack Theme Styles - Style Pack 2
 
 .site-header,
 .header-solid-background .site-header {
-	padding-bottom: $size__spacing-unit;
+	padding-bottom: #{ 0.5 * $size__spacing-unit };
 
 	@include media(tablet) {
 		padding-bottom: #{ 4 * $size__spacing-unit };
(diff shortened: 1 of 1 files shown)


laurelfulford:
> ... I'm worried it looks unintentional. On wider viewports the difference is dramatic and works really well. Does this concern resonate with you?

Yeah, it does look a bit 'off'. In the [mockups](https://cloudup.com/i5_kx4nSvQt/f), the spacing was larger on small screens, but it makes for a pretty narrow space for content in application (and I'm not sure it's enough to look intentional, either). 

I tried pushing up a change where the padding is removed from that first section altogether on smaller screens in 7556eaa -- do you think this would work? Or is it better to try increasing the padding instead?

image

laurelfulford:
> If anything, we might want to consider removing the dropdown edges in mobile since they're so small...

That was my initial fix, actually 😄 I'll merge this as an improvement for now, but I think that could help reduce some weird visual friction they introduce. You definitely can't get the same effect on such a small screen!

PR #246, opened 2019-08-14
Code simplification: helper function to check the current style pack.

jeffersonrabb:
### All Submissions:

Introduces a helper function `newspack_is_active_style_pack()` to check if the current style pack matches a slug (or list of slugs). Meant to replace `if ( 'default' !== get_theme_mod( 'active_style_pack', 'default' ) )` logic, which complicates some of the theme files. There should be no visible changes in this PR—everything should work exactly as before.

1. Verify that all Style Pack settings continue to function as before. 
2. Keep a particularly careful eye on the cases where multiple slugs are passed in to match, e.g.: https://github.com/Automattic/newspack-theme/blob/28ad20c264cc5cbf84d9f3b52bac33983eb2ced8/inc/color-patterns.php#L190-L198

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -292,7 +292,6 @@ function newspack_custom_colors_css() {
 	}
 
 	if ( true === get_theme_mod( 'header_solid_background', false ) && ! newspack_is_active_style_pack( 'style-3', 'style-4' ) ) {
-
 		$theme_css .= '
 			.header-solid-background .site-header {
 				background-color: ' . $primary_color . ';
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -192,7 +192,7 @@ function newspack_custom_colors_css() {
 			background-color: ' . newspack_adjust_brightness( $primary_color, 200 ) . '; /* base: #005177; */
 		}';
 
-	if ( 'default' === get_theme_mod( 'active_style_pack', 'default' ) ) {
+	if ( newspack_is_active_style_pack( 'default' ) ) {
 		$theme_css .= '
 			.cat-links a,
 			.cat-links a:visited,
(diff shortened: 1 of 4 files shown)


inc/color-patterns.php
```diff
@@ -271,7 +267,7 @@ function newspack_custom_colors_css() {
 		';
 	}
 
-	if ( true === get_theme_mod( 'header_solid_background', false ) && 'style-3' !== get_theme_mod( 'active_style_pack', 'default' ) ) {
+	if ( true === get_theme_mod( 'header_solid_background', false ) && newspack_is_active_style_pack( 'style-3' ) ) {
```
laurelfulford: This check is actually inverted -- it should be applied in every case but `style-3`, not only `style-3`.

If we wanted to avoid 'not' checks, what about this?

`if ( true === get_theme_mod( 'header_solid_background', false ) &&  newspack_is_active_style_pack( 'default', 'style-1', 'style-2', 'style-4' ) ) {`

template-parts/post/author-bio.php
```diff
@@ -9,7 +9,7 @@
 <div class="author-bio">
 
 	<?php
-	if ( 'style-4' !== get_theme_mod( 'active_style_pack', 'default' ) ) :
+	if ( newspack_is_active_style_pack( 'style-4' ) ) :
```
laurelfulford: This was also originally a check that it's not `style-4`, and now it's checking that it is `style-4` -- it makes all the style packs have two avatars, except Style 4, which has none.

inc/color-patterns.php
jeffersonrabb: Whoops! Fixed in https://github.com/Automattic/newspack-theme/pull/246/commits/5b95e4aeb715f10004deddd68477037831de7eb7. I don't mind using `not` checks.

template-parts/post/author-bio.php
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-theme/pull/246/commits/5b95e4aeb715f10004deddd68477037831de7eb7.

PR #255, opened 2019-08-15
Adds class to let you 'highlight' a tertiary menu item

### LINKED ISSUE #215: Menu: Add class to use to 'highlight' menu items
Add styles for a class to add to menu items.

This would then be used to highlight specific menu items, through borders, backgrounds, and text styles, in the different style packs. 

In the dropdowns, it could simply be used to add divider.

laurelfulford:
### All Submissions:

This PR replaces the original styles that 'highlighted' the first or last link in the tertiary menu.

They add styles for a class, `menu-highlight`, which can then be used to manually add a link on a specific link, so it makes more sense for the layout a specific site has chosen. 

The default style for this class is bolded with a border. 

If you're using the default header settings, it will add the primary colour as a background colour instead. 

If there's a call for it, we can add styles for this class for the other menus, too. 

Closes #215.

2. Navigate to the Customizer > Menus, and edit the menu you have assigned to the tertiary menu spot. 
3. Edit one of the menu items in the menu -- there is a field, 'CSS Classes' where you can add the class `menu-highlight`:
4. Cycle through the different header options.

* Default header - the menu item with the class should have a primary colour background.

image

image

* Solid background, or short header - the menu item with the class should be bolded and have a border:

image

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -38,6 +38,7 @@ function newspack_custom_colors_css() {
 		.main-navigation .sub-menu,
 		.mobile-sidebar,
 		.site-header .main-navigation .main-menu .sub-menu,
+		body.header-default-background.header-default-height .site-header .tertiary-menu .menu-highlight a,
 		.entry .entry-content .wp-block-button .wp-block-button__link:not(.has-background),
 		.entry .button, button, input[type="button"], input[type="reset"], input[type="submit"],
 		.entry .entry-content > .has-primary-background-color,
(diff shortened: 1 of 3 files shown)


laurelfulford:
> When the logo is centered, the highlighted tertiary menu item appears to the left of it. This seems odd - I feel like I almost always see this type of button on the far right of a header.

I totally didn't think of that, but I agree - it feels a bit weird. 

What do you think about switching the primary tertiary menu locations when the logo is centred? That would keep the highlighted button on the right half (and would put the main menu first, which is preferable anyway). 

I made a new issue for that here: #261 -- if you're thinking of a different way to fix it, can you please comment there? Thanks!

PR #265, opened 2019-08-15
Menus: Fixes multi-level dropdowns for the primary menu.

laurelfulford:
### All Submissions:

This PR makes it so multi-level menus (3+ levels) are distinguishable -- right now if you nest more than three levels deep, it just looks like one level of dropdowns.

It also simplifies the dropdown styles, removing some CSS that isn't needed.

**My menu settings:**

image

**Before:**

image

**After:** 
image

1. Set up a primary menu with three-plus levels.
2. Test on the front-end -- it will look like you just have one level of dropdowns.
4. View the menu again; your different levels should now be visible, and sub-menu items with a dropdown should show an arrow.

### Other information:

### DIFF
--- a/sass/navigation/_menu-main-navigation.scss
+++ b/sass/navigation/_menu-main-navigation.scss
@@ -193,7 +193,7 @@
 			bottom: auto;
 			height: auto;
 			transform: none;
-			min-width: 200px;
+			width: #{ 12.5 * $size__spacing-unit };
 		}
 
 		.main-menu .sub-menu .menu-item-has-children:not(.off-canvas):focus-within > .sub-menu {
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks Jeff!

> What's your plan for multi-level menus in the mobile menu? Leave as-is or add collapsible groupings?

The original plan was collapsible groupings -- I think it'll make it much easier to parse, you're not bombarded with a ton of different links of different priority all at once. I still have to dig into it from an AMP perspective, though -- I'm pretty sure we can at least re-use the toggle approach that we used for the search, I just have to wrap my head around how to approach it.

PR #268, opened 2019-08-15
Menus: Change flyout background colour.

### LINKED ISSUE #267: Dropdown menus: Let's work on some colours
Right now the dropdown menus use the primary colour as the background, which, depending on the other settings can be jarring (with the white background), or blend right in (some cases with the solid background, like the short header or Style C).

Let's try a light grey for the default (keeping the bolder hover colour), and some variation of the primary or secondary colour for the other cases.

laurelfulford:
### All Submissions:

This PR replaces the primary colour background used in the flyout backgrounds for the default header with a light grey:

image

If you've set a solid background, it uses a slightly darker version of the primary colour:

image

image

(It also does this for styles D and E, where the solid header is dark grey; I'm torn a bit between liking the pop of colour, but wondering if it's too bright - would appreciate feedback on that!). 

Closes #267 .

2. Test the header with the default background, with the default colour.
3. Test the header with the default background, with a custom colour.
4. Test the header with a solid background, with the default colour.
5. Test the header with a solid background, with a custom colour.
6. Test the header with a solid background + Style D (Style 3), and the default colour.
7. Test the header with a solid background + Style D (Style 3), and a custom colour.

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -73,7 +73,11 @@ function newspack_custom_colors_css() {
 		.mobile-sidebar a:visited,
 		.mobile-sidebar .main-navigation .sub-menu > li > a,
 		.mobile-sidebar .main-navigation ul.main-menu > li > a,
-		body.header-default-background.header-default-height .site-header .tertiary-menu .menu-highlight a {
+		body.header-default-background.header-default-height .site-header .tertiary-menu .menu-highlight a,
+		.site-header .main-navigation .main-menu .sub-menu > li > a:hover,
+		.site-header .main-navigation .main-menu .sub-menu > li > a:focus,
+		.site-header .main-navigation .main-menu .sub-menu > li > a:hover + button,
+		.site-header .main-navigation .main-menu .sub-menu > li > a:focus + button {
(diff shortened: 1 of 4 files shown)


laurelfulford:
> For styles D & E where the navigation bar is already a solid dark gray, I would expect the dropdown color to be the same gray. 

Good call -- this has been updated in ac73b05. 

> The top of the dropdown is a few pixels above the bottom of the bar, which means there's a tiny bit of overlap. Is it possible/wise to address this?

The short answer is... it's tricky. The dropdown's position has to do with the `<li>` it's nested in, and the only way to guarantee it's at the exact bottom of the header is to make the `<li>` the same height as the header... which means some pretty tall menu items, especially should they ever wrap to two lines. Because of that, I'd rather avoid this approach if we can. 

As a compromise, I adjusted the menu styles to drop it down a bit further in df5de9c when you're using the short header. It doesn't quite get past the bottom of the header, but it's close; we could also experiment with a little triangle, like on [this menu here](https://radcliffe2demo.wordpress.com/), to better visually connect them.

Let me know what you think!

laurelfulford:
Closing in favour of #429.

PR #293, opened 2019-08-20
Add hover styles to the top-level of menu.

### LINKED ISSUE #254: Menus: Make sure links have some kind of hover effect.
Right now in certain cases, the header menus don't have any kind of hover effect. 

This needs to be tested thoroughly to find out when this happens (what style pack? What custom colours), and a fix needs to be put together.

laurelfulford:
### All Submissions:

This PR adds a hover style to the top level of each of the menus (primary, secondary, tertiary), and the search toggle.

I was trying to find a way to style these that didn't require a ton of CSS to get around the different colour options, header options, and style packs.

This approach is to add a slight opacity change on hover (so the links become a bit lighter). 

If it doesn't seem noticeable enough, another option would be to use a bottom border, either a text decoration, or a box-shadow if we want a bolder line. 

Open to thoughts about how this looks, and if it would be better to go with something bolder.

Closes #254.

1. Apply PR and run `npm run build`
2. Hover over items from each menu; the top level of the primary menu, and links from the secondary, tertiary, social and search toggle should all get slightly lighter on hover.

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -169,8 +169,6 @@ function newspack_custom_colors_css() {
 
 		/* Set primary variation color */
 
-		.main-navigation .main-menu > li > a:hover,
-		.main-navigation .main-menu > li > a:hover + svg,
 		.author-bio .author-description .author-link:hover,
 		.entry .entry-content > .has-primary-variation-color,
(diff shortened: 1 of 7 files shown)


laurelfulford:
Thanks!

PR #302, opened 2019-08-22
Cover Blocks: Remove/override default cover block font styles

laurelfulford:
### All Submissions:

The cover block originally only let you nest a header, body text, and a button. Because of that, the theme's original styles (and the block's own default styles) reflect that -- and caused some issues.

This PR makes sure blocks inside of the cover block use standard font sizes, and make sure h2s don't have a max-width. 

1. Copy paste [this content](https://cloudup.com/c9Pq7o_IDhB) into the editor (a cover block, with one Newspack article block inside of it).
2. View in the editor and on the front end; note that the section header is narrower than the rest of the content, and weirdly large:

image

... and the editor, it's also not white and has a max-width of 610px:

image

4. Confirm that the section header is using the right font size, and in the editor uses the right colour and is the same width:

image

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -396,7 +396,8 @@ function newspack_custom_colors_css() {
 		.editor-block-list__layout .editor-block-list__block .wp-block-pullquote.is-style-solid-color a,
 		.editor-block-list__layout .editor-block-list__block .wp-block-cover a,
 		.editor-block-list__layout .editor-block-list__block .wp-block-newspack-blocks-homepage-articles .entry-title a,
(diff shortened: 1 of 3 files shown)


laurelfulford:
> The sample content looks pretty different in my testing. Do I need to set any Customizer options to test correctly?

Yes, sorry! I had the 'default' style pack on in my testing.

I've updated the CSS a bit to remove some more of the default h2 styles when inside of the cover block in 6ff8391 -- the centred text alignment, and the left padding.

laurelfulford:
Cool! Thanks for the review :)

PR #317, opened 2019-08-23
Set a specific size for the featured image on single posts

laurelfulford:
### All Submissions:

This PR adds a specific image size for the featured image; it's to fix an issue with images  not resizing properly on the staging sites.

Unfortunately I couldn't recreate this locally; the steps mostly focus on making sure nothing changes. I did confirm it works by editing the code on one of the staging sites with this problem though. 

1. Set up two posts with featured images; make sure one is larger than 1200 (so it appears above the content) and one is smaller than 1200 (so it uses the fallback behaviour, and appears in the content)
2. Apply the PR.
3. Confirm that images look the same. 

Once the theme is updated on the staging sites, then we can confirm that this is totally fixed.

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -60,6 +60,7 @@ if ( ! function_exists( 'newspack_setup' ) ) :
 		add_theme_support( 'post-thumbnails' );
 		set_post_thumbnail_size( 1568, 9999 );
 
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks for the testing!

PR #328, opened 2019-08-26
Update column spacing when the border style is used

laurelfulford:
### All Submissions:

This PR increases the spacing between columns when you have the border style turned on.

From the front-end styles I also removed some styles from the columns (the `&[class*='has-'] > *` stuff) -- the columns block used to add classes to determine how many columns  

Note: there's an issue if you have custom widths set for your columns, where the columns wrap oddly at the tablet breakpoint where it seems like they shouldn't. I could recreate it without this PR, and using Twenty Nineteen, so it's not related to this PR or anything in the theme. I need to follow up on it.

1. Copy-paste this [test code](https://cloudup.com/cEATJk284YX) into the editor.
2. You should see something like this on the front-end and in the editor:

image

3. Apply the PR and run `npm run build`; confirm the columns are more spaced out on the front-end and in the editor:

image

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -703,12 +703,21 @@
 					margin-bottom: 0;
 				}
 			}
+		}
 
-			&[class*='has-'] > * {
-				margin-right: $size__spacing-unit;
+		&.is-style-borders {
+			.wp-block-column {
 
-				&:last-child {
-					margin-right: 0;
+				@include media(tablet) {
+					flex-basis: calc(50% - 32px);
+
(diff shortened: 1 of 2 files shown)


laurelfulford:
I think the issue may be my poor testing steps: the border style is on the Columns block, not the Article block. Do you see the option on that block?

It was introduced in a pretty new PR (https://github.com/Automattic/newspack-blocks/pull/85) so the repo might need to be updated locally?

Let me know if either of those don't do the trick and we can dig into things more next week!

laurelfulford:
Great - thanks Jeff!

PR #334, opened 2019-08-27
Style 3: Update column border styles 

laurelfulford:
### All Submissions:

This PR updates the column border styles for Style 3 to match the other borders/dividers in that style pack -- rather than light grey and solid, they're dark grey and dotted:

image

2. Navigate to Customize > Style Packs and pick Style 3.
3. Copy-paste [this test content](https://cloudup.com/cIRU38mjKqN) into the code view of the editor.
4. View on the front-end and in the editor and confirm that the borders are dotted, like the screenshot above.
5. Scale down the browser window size -- confirm that around the tablet size, when the three columns wrap to two per line, that the middle column -- second in the first row -- no longer has a border to the right. 

image

6. Scale down the browser window size -- confirm on the mobile view, when the columns are stacked, that the dotted borders separate the columns at the bottom, except the last column, which shouldn't have a bottom border.

image

### Other information:

### DIFF
--- a/sass/styles/style-3/style-3-editor.scss
+++ b/sass/styles/style-3/style-3-editor.scss
@@ -152,3 +152,27 @@ figcaption,
 hr {
 	border-top: 1px dotted $color__text-main;
 }
+
+.wp-block-columns.is-style-borders {
+	[data-type="core/column"] {
+		border-bottom: 1px dotted $color__text-main;
+
+		@include media( mobile ) {
+			border-bottom: 0;
+
+			&:after {
+				border-right: 1px dotted transparent;
+			}
+
+			&:nth-child(odd):after {
+				border-color: $color__text-main;
+			}
+		}
+
+		@include media( tablet ) {
+			&:after {
+				border-color: $color__text-main;
+			}
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


laurelfulford:
> This works correctly up to a viewport width of 767px. 768px and wider, I get this result with the columns breaking correctly but the second column on the first row showing a right border:

Oh yikes 😬 I'm getting this on master, too. I'm not exactly sure what's happening -- it's like the columns are wrapping sooner than "expected" -- I might have the breakpoints wrong in the original column border styles.

I'm going to try to untangle where this is happening and follow up here.

> Also, the top border is present in view but not in the editor. Is this expected?

It is expected -- the border's used in all the style packs when you don't have a featured image, though it's a lot more noticeable in this one due to it being darker. 

I think it might be best to leave it out, since it doesn't always appear on the front end -- so it would either incorrectly always show in the editor, or show/hide in a way I can't imagine would be easy to figure out.

laurelfulford:
I think I have this resolved by updating the theme's tablet breakpoint to match Gutenberg's in 
2b82489; I'm going to do the same for our blocks, for consistency. 

Just give me a ping if you continue to spot weirdness!

laurelfulford:
Great!

PR #343, opened 2019-08-28
Centred Logo: Allow for larger logos when centred

laurelfulford:
### All Submissions:

Right now, when you add a wider logo to the header and centre it, it gets cut off (as @claudiulodro noted in #319.

This PR tries to improve that behaviour, to allow for wider, more banner-like logos. 

There are some limits -- larger logos will squish out the menu to the right, for example, and even when you use small logos, the menus only get so much space. So some thought needs to go in what header to use, how many links to put in the header, and how large the logo should be set to be, depending on what's needed. 

1. Upload the below logo as your site logo, and size it up to be larger.

WordPress-logotype-standard

2. Under Customize > Header Settings, pick 'Center Logo'

2. View on the front-end; note that the logo gets cut-off.

image

4. Confirm that the logo now doesn't get cut off (though if it was set quite large, it may be squishing out the tertiary menu now). 

image

This is partially because I removed the fixed logo size in the theme as part of these updates, since you can resize it. If it's too large now, it just needs to be scaled down again using the Logo Resizer:

image

5. Try resizing the browser window and confirm things behave fairly well -- this is approximately tablet-sized, before the mobile menu appears:

image

6. Try a smaller square-shaped logo as well (you can crop the one I attached above for this effect) -- when a smaller logo is in place, each section of the header will be the same width: 

image

It's only when you have a larger logo that it starts eating into the menu space.

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -96,8 +96,6 @@ if ( ! function_exists( 'newspack_setup' ) ) :
 		add_theme_support(
 			'custom-logo',
 			array(
-				'height'      => 400,
-				'width'       => 800,
 				'flex-width'  => true,
 				'flex-height' => true,
 				'header-text' => array( 'site-title' ),
--- a/inc/logo-resizer.php
+++ b/inc/logo-resizer.php
@@ -65,9 +65,11 @@ function newspack_customize_logo_resize( $html ) {
 		// get the logo support size
 		$sizes = get_theme_support( 'custom-logo' );
 
+		$logo_max_width = ( $logo['width'] > 600 ) ? 600 : $logo['width'];
+
 		// Check for max height and width, default to image sizes if none set in theme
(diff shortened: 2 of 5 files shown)


laurelfulford:
Thanks for catching this, Jeff -- I can see this too when I test in Chrome (for some reason it works in Firefox though!). Darn it -- I'll revisit and get it sorted.

laurelfulford:
The centred logo should now no longer be getting cut off as of ce9a093  -- now that I found the issue, I'm a little confused how it was actually working in the first place but maybe Firefox was being forgiving to me.

laurelfulford:
> I wonder if we should apply a sanity max-width to the logo at the widest sizes? 

I tried adding a max width of 600px to the code in b824a05; depending on how long the menus are you could still technically break it, but in that case I think we can safely recommend not using the centred logo. 

Let me know what you think!

laurelfulford:
Cool - thanks Jeff!

Issue #213, opened 2019-08-12
Menu: Add touchscreen logic for dropdown menu for AMP


laurelfulford:
On larger touchscreens, the desktop menu dropdowns are visible, but can't be used.

This can be fixed with JavaScript for the non-AMP version of the site. 

We'll need to find an approach to fix for AMP sites -- ultimately it could just be showing that mobile menu on larger screens than we do currently.

laurelfulford:
I'm going to close this issue for now, as it hasn't come up since I first created this ticket.

PR #352, opened 2019-08-29
Add touch-screen fallback javascript for the dropdown menu

laurelfulford:
### All Submissions:

This PR adds JavaScript to help the dropdown menu work on touch-screens when AMP is not enabled. 

This approach uses jQuery; I spent way too much time failing to get this to work with vanilla JavaScript, so I went this route to get a first pass at this fix done. I think it can (and should!) be optimized to work without, though, since I don't believe we're loading that library on the front-end for anything else.

As for the AMP enabled version I still need to dig into an approach there. For the short-term, you can use the down arrows to toggle the menus with AMP enabled on touch screens; at the very least I think we can increase the breakpoint for where the mobile menu displays, so you don't see this menu as often on tablets. 

See #213 .

1. Apply the PR.
2. Make sure your primary menu has at least one dropdown. 
3. Switch AMP to 'Reader' mode.
4. Test on a touchscreen device that's large enough to show the dropdown menu (should be most tablets at this point). Tapping on a menu item with a dropdown once should open it; tapping a second time should follow that link.
5. Switch AMP to Standard mode.
6. Confirm that just tapping the arrow part of the menu item with child pages will open the menu (just to confirm that this indeed works for someone other than me). 

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -286,7 +286,7 @@ function newspack_scripts() {
 			'close_search' => esc_html__( 'Close Search', 'newspack' ),
 		);
 
-		wp_enqueue_script( 'newspack-amp-fallback', get_theme_file_uri( '/js/amp-fallback.js' ), array(), '1.0', true );
+		wp_enqueue_script( 'newspack-amp-fallback', get_theme_file_uri( '/js/amp-fallback.js' ), array( 'jquery' ), '1.0', true );
 		wp_localize_script( 'newspack-amp-fallback', 'newspackScreenReaderText', $newspack_l10n );
 	}
 	// Load custom fonts, if any.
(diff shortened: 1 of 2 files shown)


jeffersonrabb (commented):
This one has a question for @claudiulodro: can we use `amp-sidebar` (https://amp.dev/documentation/components/amp-sidebar/) in non-AMP contexts producing identical UI without having to develop and support two implementations?

laurelfulford:
Closing this PR since I can't see us using this approach (jQuery) for something like this should we end up adding this; we should start with a jQuery-free fresh PR if we do.

mmprep:
Is there any progress in getting rid of jQuery dependency completely?  Still, it negatively affects the performance of the site as a whole. And keeping it active solely to ensure the dropdown menu functionality is not too optimal.

PR #377, opened 2019-09-06
Style 3: Remove 'overlap' style when the block is in a group or cover block

### LINKED ISSUE #301: Style 3: Block overlap styles don't work in some cases
The 'overlap' style used on the homepage blocks with style 3 don't look good in a few cases:

* When the block is nested in a cover block -- it's a white block, and it's white text in white block.
* When the block is using the upcoming 'behind' article style for the image -- again, it's a white block with white text in it.

In both cases, the block styles should 'fall back' to something more standard.

laurelfulford:
### All Submissions:

In Style 3, the article block titles are 'pulled up' over the featured images.

To do this, we have to add a white background to the titles, which doesn't work well when they're nested inside of the cover or group blocks.

This PR removes that particular style in those cases, so the titles sit below the images. 

**Before:** 

image

**After:** 

image

Closes #301  .

1. Navigate to Customizer > Style Packs, and change to Style 3.
2. Copy-paste [this test content](https://cloudup.com/cpl3NU08Mab) into the code editor.
3. Update the article block settings to make sure at least one article per block has an image.
4. View on the front end and in the editor; note the odd white block.
5. Apply the PR and run `npm run build`
6. Confirm that the titles now sit below the images, and don't have white backgrounds.

### Other information:

### DIFF
--- a/sass/styles/style-3/style-3-editor.scss
+++ b/sass/styles/style-3/style-3-editor.scss
@@ -30,14 +30,22 @@ Newspack Theme Editor Styles - Style Pack 3
 	}
 }
 
-.wp-block-newspack-blocks-homepage-articles.show-image:not(.image-alignleft):not(.image-alignright) {
-	.post-has-image .entry-title {
-		background-color: $color__background-body;
-		margin-top: -1.75em;
-		padding: 0.5em 0.25em 0 0;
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks Jeff!

PR #381, opened 2019-09-09
Fix button padding when nested in a floated cover block

### LINKED ISSUE #369: Cover Block + Button Block + Floating = no padding
It looks like if you set up the following:

1. Add a cover block
2. Insert a button into it
3. Align the cover block left or right
4. View on front-end...

... it looks like the button loses its padding. Need to reproduce and then find a fix.

laurelfulford:
### All Submissions:

In an attempt to remove extra space from floated Cover Blocks, I removed the padding from the first and last item. But this caused some weirdness with the button block, so this PR removes it. 

Closes #369 .

1. Copy [this test content](https://cloudup.com/c616VScV4Ja) into a post.
2. View on the front end and in the editor; note the lack of padding on the button.

image

4. Confirm that the button now looks correct on the front-end and in the editor.

image

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -566,12 +566,10 @@
 
 				[class*="wp-block-"]:first-child {
 					margin-top: 0;
-					padding-top: 0;
 				}
 
 				[class*="wp-block-"]:last-child {
 					margin-bottom: 0;
-					padding-bottom: 0;
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks Jeff!

> What was the rationale for adding the 0 padding in 3810f9a, and will this change cause any regressions?

It was meant to dial down the vertical spacing on the block a bit when it was floated (and in turn, just make it a bit shorter). But the margin would be more important to remove, and that's still there -- I should've realized removing the padding would just cause weirdness, since it's inside of elements and likely visible in them (like with the button!). 

Gutenberg 6.4 also includes the ability to set a `min-height` on the cover block on a per-block basis, which kind of makes my adjustments on the theme level a bit redundant :)

PR #383, opened 2019-09-11
Fix donation block custom colours

### LINKED ISSUE #339: Donations Block: Make sure it's picking up custom colours
Make sure the donation block is picking up the custom colours set by the theme, both on the front-end and in the editor.

Right now only the button uses the custom colours on the front-end; in the editor, both the amount picker and button use the default colours, regardless of settings.

laurelfulford:
### All Submissions:

The donation block wasn't picking up the primary colour for the current selection, or the secondary custom colour for the button; this PR fixes that.
 
Closes #339.

1. Change your site to use two custom colours.
2. Add a donation block to a post or page.
3. Confirm its appearance in the editor and on the front end; it should be using all blue in the editor, and blue and your primary colour on the front end.

Front end: 
image

4. Apply the PR and run `npm run build`
5. Confirm that in the editor and on the front end, both your primary and secondary colours are used, not the Newspack blue:

image

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -44,7 +44,8 @@ function newspack_custom_colors_css() {
 		.entry .entry-content *[class^="wp-block-"] .has-primary-background-color,
 		.entry .entry-content *[class^="wp-block-"].is-style-solid-color,
 		.entry .entry-content *[class^="wp-block-"].is-style-solid-color.has-primary-background-color,
-		.entry .entry-content .wp-block-file .wp-block-file__button {
+		.entry .entry-content .wp-block-file .wp-block-file__button,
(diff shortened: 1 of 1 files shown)


laurelfulford:
> The only very subtle thing I noticed is that if you select Other, the outline of the text input is system blue.

It looks like it uses the blue border whether or not the input is focused, which isn't great -- I can tackle that in a separate PR though. I think it might be something we want to remove on a block level, not just on a theme level. 

> There may also be an accessibility argument for leaving it as a predictable, non-custom color, in which case this would be a negative change anyway.

This is a good point, and a wider issue through the theme (we're using the secondary colour for contrast all over the place, and it can be very light and hard to see). I made a separate issue for it here: #389.

PR #390, opened 2019-09-13
Single Post: First pass at adding featured image position option, to display behind title

laurelfulford:
### All Submissions:

This PR adds an option to the featured image area of posts to display the featured image behind the post title.

When select, the post title, post meta (author name and published date) and share buttons should display over the image. To reduce visual noise, I've also hidden the author avatar.

The styles managing this still need a bit of fine tuning -- right now the image is set to be 80% of the screen size on desktop and shorter on smaller screens, but I'd like to explore this a bit more in future commits to optimize behaviour for different image sizes (I need to do further testing with a full-sized monitor). 

In the meantime, I'd love some feedback on how it works/looks, and anything that seems weird/unexpected. 

2. Create a new post; confirm you have an option for featured image position, set to 'Default':

image

3. Add a featured image; right now the original 'too-small-image' fallback behaviour still applies, so images smaller than 1200px will be placed in the content area. Make sure to pick a larger image, and if this seems 'wrong', let me know!
4. Select 'Behind Article Title' for the image positioning.
5. Add a post title, and any other testing content you'd like.
6. Publish the post.
7. View on the front-end; make sure your featured image is now displaying behind the content, and looks roughly like this:

image

8. Try scaling down the screen and check for any 'gotchas' or weirdness. 

### Other information:


jeffersonrabb:
Putting this on hold until https://github.com/Automattic/newspack-theme/pulls ships.

laurelfulford:
Thanks Jeff! 

> Only issue I found, which is probably for another PR, is that when the header is tall it's common to get an initial display like this, with the title below the fold. Any ideas for how to resolve?

This is something in the past I would've resolved with JavaScript, but I'd like to avoid that (even minus the AMP factor, it always leaves you with a bit of a flash of the wrong height as it calculates things after load, which isn't ideal). 

I think we might be able to make some educated guesses about how tall the header is based on the settings, and set the image height from there. I can take a swing at that in a separate PR; at the very least, it should improve placement enough that the top of the title is visible, even if you don't get the full byline.

PR #396, opened 2019-09-13
Editor: Make sure wide template editor preview is correct

laurelfulford:
### All Submissions:

This PR adds the ability to add a class to the Gutenberg editor based on the current template, so the preview can be styled correctly.

It specific deals with making sure the 'One column wide' template content previews at 1200px wide; I'll build on it further to make sure wide and full align content doesn't display as such when you're using the default (two column) template.

2. Create a new post and add some content; click publish.
3. Confirm that the content doesn't exceed 780px wide in the editor.
4. Switch to the 'One col wide template', click 'Update', then refresh the page (unfortunately this approach requires a full page refresh, but if problematic we can look into adding the body tag via JS on template change).
5. Confirm that the post content in the editor now previews at 1200px wide.

### Other information:

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -410,6 +410,18 @@ function newspack_is_active_style_pack() {
 }
 
 /**
+ * Check for specific templates.
+ */
+function newspack_check_current_template() {
+	global $post;
+
+	$post_id       = $post->ID;
+	$template_file = $post_id ? get_post_meta( $post_id, '_wp_page_template', true ) : '';
+
+	return $template_file;
+}
+
+/**
  * Add body class on editor pages if editing the static front page.
  */
 function newspack_filter_admin_body_class( $classes ) {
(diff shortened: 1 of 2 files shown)


functions.php
```diff
@@ -409,6 +409,18 @@ function newspack_is_active_style_pack() {
...
 }
 
+/**
+ * Check for specific templates.
+ */
+function newspack_check_current_template() {
+	global $post;
+
+	$post_id       = $post->ID;
+	$template_file = $post_id ? get_post_meta( $post_id, '_wp_page_template', true ) : '';
```
jeffersonrabb: When on Dashboard page with no post (e.g. `/wp-admin`) I get this PHP Warning:

` Notice: Trying to get property 'ID' of non-object in /srv/www/newspack-repos/newspack-theme/functions.php on line 418`

This can be corrected by removing  line 418 and changing this line to:

```suggestion
	$template_file = ( $post && $post->ID ) ? get_post_meta( $post->ID, '_wp_page_template', true ) : '';
```

laurelfulford: Ah, good catch, Jeff! This has been fixed in 440fd06 .

laurelfulford:
Thanks Jeff! The error should now be fixed!

Created an issue here to improve the preview: https://github.com/Automattic/newspack-theme/issues/416

laurelfulford:
Thanks for catching it! :)

PR #407, opened 2019-09-20
Improve appearance of featured image when set as background

laurelfulford:
### All Submissions:

This PR improves the styles applied when the featured image is set as a background, to try to make sure the post title is not cut off. 

If you have a very tall logo you can still cut it off, but I would consider that an extreme case; most logo sizes should allow for the post title to appear, if not the byline and share icons.

2. Create a new post and add a featured image that's at least 1200px wide.
3. Under the featured image, pick the 'Behind Article Title' option.

image

4. View on front-end; confirm that the post title is visible on different screen sizes. 

image

5. Navigate to Customize > Header Settings and pick the 'Short Header'; repeat step 4. 
6. Edit the post again and pick a featured image that's narrower than 1200px wide. 
7. View on the front-end; regardless of option selection, the featured image should display in the content area.

### Other information:

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -82,7 +82,7 @@ function newspack_body_classes( $classes ) {
 
 	// Adds a class if singular post has a large featured image
 	$thumbnail_info = wp_get_attachment_metadata( get_post_thumbnail_id() );
-	if ( is_single() && has_post_thumbnail() && 1200 > $thumbnail_info['width'] ) {
+	if ( is_single() && has_post_thumbnail() && 1200 < (int) $thumbnail_info['width'] ) {
 		$classes[] = 'has-large-featured-image';
 	}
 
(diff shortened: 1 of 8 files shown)


laurelfulford:
Thanks Jeff!

> I tried all kinds of window sizing and the only size I found that forced the title out of view was height < 780 and width > 780. This is a pretty unusual size, but do you think we should address it anyway?

I think it would be nice to, even if it's odd -- I'll give it a shot! 

> Also, what's the reasoning for removing the border under the title if the image is large?

The border is meant to be a divider between the title and content area; when you have a large featured image, it kind of serves this purpose already -- here it is with:

image

And without:

image

I had the logic wrong originally, and the background image style made it a lot more obvious (the border just sat along the top of the image, so was invisible).

laurelfulford:
With c99e33a, you should now be able to squish up the browser window vertically and have the title always in view; if you make it really _really_ short, the title shouldn't cut off on the top, but you will have to scroll to see it all -- it's a very extreme case:

image

laurelfulford:
Thanks!

PR #411, opened 2019-09-21
Cover Block: Reduce the amount of padding on the cover block further

### LINKED ISSUE #365: Cover Block: Reduce side padding further
We've already reduced the padding on the cover block somewhat, when it's nested in a column block, but we can probably take this further.

May also be worth looking into removing the word wrap in certain cases -- or switch it to a ellipsis or some other situation that may look less broken? Needs investigation.

laurelfulford:
### All Submissions:

This PR removes the padding on columns block even further when they're nested in columns; otherwise there's really not a lot of room.

(I'll be tackling the issue with the too-tall final column in a separate PR).

**Before:**

image

**After:**

image

Closes #365.

1. Copy-paste [this test content](https://cloudup.com/cXdf3ez2rSd) into a post.
2. View in the editor and on the front-end; note how much space in on the right and left for the content.
4. Confirm that there's less space on the right and left of the content.

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -579,6 +579,11 @@
 		.wp-block-cover,
 		.wp-block-cover-image {
 			min-height: 330px;
+			padding-left: $size__spacing-unit;
+			padding-right: $size__spacing-unit;
+			.wp-block-cover__inner-container {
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks!

PR #412, opened 2019-09-21
Remove top and bottom no margin from nested blocks

laurelfulford:
### All Submissions:

This PR removes a style that was removing the top margin of a first child element, and the bottom margin of a last child element.

Overall this would be reducing some extra spacing when elements are nested in the content, and was a style brought over from Twenty Nineteen.

In application -- especially as block nesting is becoming more complex -- it was causing unexpected visual issues, like making the last column taller than the rest:
 
image

I think removing the style will cause fewer issues than keeping it in. 

1. Copy paste [this test content](https://cloudup.com/cXdf3ez2rSd) into a post.
2. View on the front-end -- note that the last column appears taller than the rest.
4. Confirm that the issue is now fixed.

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -5,14 +5,6 @@
 	margin: 32px 0;
 	max-width: 100%;
 
-	> *:first-child {
-		margin-top: 0;
-	}
-
-	> *:last-child {
-		margin-bottom: 0;
-	}
-
 	// When the image block is aligned left or right, the markup changes,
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks!

> Looks like those styles came in as part of the original TwentyNineteen fork, so we didn't add them for any particular reason.

Yup, accurate :) In reality, as Gutenberg changes, they aren't working as well as they did circa December 2018!

PR #419, opened 2019-09-23
Remove bottom spacing from blocks nested inside of column or group blocks

laurelfulford:
### All Submissions:

This fixes a similar issue that code removed in #412 was suppose to help with, but with a bit more focus on specific blocks, and making sure it wasn't introducing new issues (like the different heights in the columns block that #412 fixed).

It removes the bottom margin from the last element inside of a group block and each individual columns blocks, and, on larger screens, removes the bottom margin from the bottom of each individual column.

**Before:**

image

**After:**

image

1. Copy-paste [this test code](https://cloudup.com/cH8sR1XhC4S) into the editor -- it has a group block, with a columns block inside; each column contains a cover block so you can easier see the height.
2. Save and publish.
3. Note the bottom spacing on the front-end -- there is quite a bit more grey than there is on the top.
5. View the page on the front-end again; confirm that the top and bottom spacing in the group block is now consistent. 

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -705,12 +705,13 @@
 	//! Columns
 	.wp-block-columns {
 
-		@include media(tablet) {
-			.wp-block-column > * {
+		.wp-block-column > * :last-child {
+			margin-bottom: 0;
+		}
 
-				&:last-child {
-					margin-bottom: 0;
-				}
+		@include media(tablet) {
+			.wp-block-column {
+				margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks Jeff!

> clearly unrelated to this PR but noting in case that's an issue that should be addressed separately.

Oh strange! I can look into this, but can you throw any additional info into an issue if you think they're relevant (the style pack you're using, whether you have a custom colour or if it's light/dark)? Thanks!

PR #421, opened 2019-09-24
Cover Block: Override `height: 100%` when nested in columns block

### LINKED ISSUE #420: Cover Block: Fills up entire height when nested in columns
When you nest a cover block inside of a columns block, it will fill up the entire vertical space... which can look a bit odd when the content in the other column is long:

image

laurelfulford:
### All Submissions:

Right now, when you nest a cover block inside of a columns block, it will fill the available vertical space on the front end (it displays as expected in the editor). This is fine, except when the other column is very long -- then you end up with an oddly tall cover block. Or, if you have more than one thing in the column with the cover block, it will still be 100% of the column's height, and hang out the end. 

I think this was introduced with the height option for the Cover block.

This PR overrides the `height: 100%` in that case, resulting in a more predictable height.

Closes #420 

1. Copy-paste [this test content](https://cloudup.com/cX7lzYzU15E) into the editor
2. View on the front-end; note that the cover block on the right is very tall:

image

4. Confirm that the cover block is now displaying at a more standard height on the front-end:
image

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -570,6 +570,7 @@
 	.wp-block-columns {
 		.wp-block-cover,
 		.wp-block-cover-image {
+			height: auto;
 			min-height: 330px;
 			padding-left: $size__spacing-unit;
 			padding-right: $size__spacing-unit;


laurelfulford:
Thanks Jeff!

PR #424, opened 2019-09-24
Make sure the author bio is not displayed on single posts and pages.

laurelfulford:
### All Submissions:

This PR makes sure that the author bio won't show up on pages when you use the one-column and one-column wide templates. 

1. Set a page using an author with an author description in their profile (the criteria to have the author profile display at the bottom). Publish with the default template.
2. Set a second page; apply the 'one column' template and publish.
3. Set a third page; apply the 'one wide column' template and publish.
4. View all three on the front end; note that the first one doesn't have an author bio on the bottom, but the second and third ones do:

image

6. Confirm that the second and third test pages no longer display the author bio.
7. Double-check a post and confirm that the author bio still displays.

### Other information:

### DIFF
--- a/template-parts/post/author-bio.php
+++ b/template-parts/post/author-bio.php
@@ -5,7 +5,7 @@
  * @package Newspack
  */
 
-if ( (bool) get_the_author_meta( 'description' ) ) : ?>
+if ( (bool) get_the_author_meta( 'description' ) && is_single() ) : ?>
 <div class="author-bio">
 
 	<?php


laurelfulford:
Thanks for looking at this, @claudiulodro!

> I wasn't able to make it work like you describe.

Do you mean you can't recreate the issue?

claudiulodro:
Oh sorry, I misread the testing instructions. I thought the intended behavior was that one of the page templates doesn't have the bio but the other ones do. Turns out the intended behavior is that none of the page templates have a bio, so this is good to go. :)

laurelfulford:
Gotcha! Yeah, this was kind of convoluted 😄 Thanks for the testing!

PR #440, opened 2019-09-30
Adjust mobile menu breakpoint

laurelfulford:
### All Submissions:

This PR increases the breakpoint at which the site will switch to the mobile menu, from a smaller tablet size (782px) to a size that should cover larger tablets (1100px) and also allow for longer menus, since it will switch to the mobile version rather than starting to wrap.

2. Confirm that the menu toggle appears on screens narrower than 1100px.
3. Confirm that the menu toggle works when AMP is enabled.
4. Confirm that the menu toggle works when AMP is disabled.

### Other information:

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -224,7 +224,7 @@ function newspack_primary_menu() {
 		return;
 	}
 	?>
-	<nav toolbar="(min-width: 1099px)" toolbar-target="site-navigation" class="main-navigation" aria-label="<?php esc_attr_e( 'Top Menu', 'newspack' ); ?>">
+	<nav toolbar="(min-width: 900px)" toolbar-target="site-navigation" class="main-navigation" aria-label="<?php esc_attr_e( 'Top Menu', 'newspack' ); ?>">
 		<?php
 		wp_nav_menu(
 			array(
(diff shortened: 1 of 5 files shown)


thomasguillot:
To me, it feels a bit much to still have the mobile menu at 1099px. Larger tablets should be fine with a "normal" menu. Is it mainly to allow longer menus?

For example, in Varia, the "normal" menu appears at 560px.

laurelfulford:
> Is it mainly to allow longer menus?

Yeah, exactly -- mostly with certain header configuration. With the default, there's a ton of space for mentus -- if someone runs out of room there, it would be much better to reorganize the menu. 

But there's a 'short' menu option (under Customize > Header Settings), that limits the space a bit:

image

... and you can get a bit of a pile-up with the current breakpoint before the mobile menu kicks in:

image


What do thing about only changing the breakpoint when the short menu is used, and dropping it down a bit, to, say, `900`. I tried to reuse the breakpoint used for desktop, but it's probably better to make a custom one so it's more flexible.

thomasguillot:
Might be worth adding a new variable, something like `$narrow_desktop_width`.

And yes probably better if it's only used by the short menu.

laurelfulford:
Thanks @thomasguillot! 

I updated the code to create a new breakpoint, then to only use that breakpoint when you also have the short header set -- otherwise, the mobile menu should only appear when you hit the tablet breakpoint.

Can you give this another test run and see how it feels? The cases it would need to be tested would be:

1. Standard height header
2. Standard height header + AMP
3. Short header
4. Short header + AMP

In both cases, it just needs to be confirmed that the mobile menu toggle appears at the correct point, the desktop menu disappears, and when you open the mobile menu it actually includes links (this one more for when testing AMP).

thomasguillot:
I can confirm that it works well.

🚢

laurelfulford:
Thanks @thomasguillot!

PR #446, opened 2019-10-01
Hide Jetpack-specific comments header

### LINKED ISSUE #445: Comments: Header doubled
It looks like the comment headers are doubled in the theme when you use Jetpack comments.

The header should only appear once.

laurelfulford:
### All Submissions:

This PR hides the second comments headed when you have Jetpack comments enabled.

Closes #445.

1. Navigate to WP Admin > Jetpack > Settings > Discussion, and check "Let visitors use a WordPress.com, Twitter, Facebook, or Google account to comment"

image

(You need to use a site that can fully connect to Jetpack -- not just in developer mode -- to enable this).

2. View a post with comments enabled; note the double-header:

image

4. Confirm that the second header is now gone:

image

### Other information:

### DIFF
--- a/sass/site/primary/_comments.scss
+++ b/sass/site/primary/_comments.scss
@@ -54,6 +54,12 @@
 			}
 		}
 	}
+
+	// Hide Jetpack comments' title.
+	.comments-title-wrap + .comment-respond .comment-reply-title {
+		display: none;
+	}
+
 }
 
 #comment {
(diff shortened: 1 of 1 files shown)


laurelfulford:
> If both Core and Jetpack commenting are on, the comment form we see is the Jetpack one, correct?

Yes! It's basically acting just like WP.com comments do: the Core comment form is replaced with one coming from Jetpack, but the comments are stored in the same place/displayed the same way. 

> There seems to be an unusual amount of vertical space above and below the header ("Leave a comment"). Is this intentional?

Noooo -- taking a second swing at this. :)

laurelfulford:
The spacing with the Jetpack comments or standard comments was the same, but I reduced it by about 1/3 -- it's reading to me as a bit much. 

Let me know what you think!

laurelfulford:
Thanks!

PR #450, opened 2019-10-02
Tweaks to the comment templates for Renaming Comments

philipjohn:
### All Submissions:

Changes the comments templates to use the helper from the new [Rename Comments plugin](https://github.com/Automattic/newspack-rename-comments) to enable publishers to rename the comments section if they so wish.

1. Install and activate the [Rename Comments plugin](https://github.com/Automattic/newspack-rename-comments)
2.Apply the PR
3.Configure the Rename Comments plugin
4. Check the comments section is appropriately renamed on the front end.

### Other information:

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -28,13 +28,9 @@ $discussion = newspack_get_discussion_data();
 		<?php
 		if ( comments_open() ) {
 			if ( have_comments() ) {
-				echo ( function_exists('\Rename_Comments\get_text') )
-					? \Rename_Comments\get_text( 'rc_comments_title' )
-					: __( 'Join the Conversation', 'newspack' );
+				echo apply_filters( 'newspack_comment_section_title_nocomments', __( 'Join the Conversation', 'newspack' ) );
 			} else {
-				echo ( function_exists('\Rename_Comments\get_text') )
-					? \Rename_Comments\get_text( 'rc_no_comments_title' )
-					: __( 'Leave a comment', 'newspack' );
+				echo apply_filters( 'newspack_comment_section_title', __( 'Leave a comment', 'newspack' ) );
 			}
 		} else {
 			if ( '1' == $discussion->responses ) {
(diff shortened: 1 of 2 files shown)


comments.php
```diff
@@ -28,9 +28,13 @@
 		<?php
 		if ( comments_open() ) {
 			if ( have_comments() ) {
-				_e( 'Join the Conversation', 'newspack' );
+				echo ( function_exists('\Rename_Comments\get_text') )
+					? \Rename_Comments\get_text( 'rc_comments_title' )
+					: __( 'Join the Conversation', 'newspack' );
```
claudiulodro: IMO it would be clearer and more maintainable to use a filter. e.g.
```
echo apply_filters( 'newspack_theme_comment_form_heading', __( 'Join the conversation', 'newspack' );
```

Then the Rename Comments plugin could just hook into that filter and replace the text easily without us having to have a special case for the Rename Comments hard-coded into the theme.

claudiulodro: Or we could use a more general filter name like `comment_form_heading` if you want Rename Comments to be usable with themes other than the Newspack theme.

philipjohn: You're right, filters would be much more elegant. I've started on that in [
8ae11f3] but I'm having a hard time thinking about an elegant way to do the `%d Comment(s)` strings [here](https://github.com/Automattic/newspack-theme/pull/450/files#diff-2cbb2554146e0dd25cc6f173565ff092R15-R16).

philipjohn: Ah, I may have a solution in [2e871cf] which couples with [https://github.com/Automattic/newspack-rename-comments/commit/434176d0c7288b19b18208705c1e7fed509dc2d2]

claudiulodro: That's clever. I think it should work.

There were a couple issues I noticed in the plugin, as a heads up:

- The comment heading appears to be getting caught in your gettext filter:
Screen Shot 2019-10-03 at 9 59 06 AM
Screen Shot 2019-10-03 at 10 00 04 AM

- [You've left some logging active](https://github.com/Automattic/newspack-rename-comments/blob/master/newspack-rename-comments.php#L313)

The work in the theme looks good to go, though, and this can be merged.

PR #486, opened 2019-10-09
Style 5: Update the navigation to make sure it's using a dark background colour

thomasguillot:
### All Submissions:

* Update the `mobile-sidebar`
* Update sub-menu links
* Add correct background colour to `.middle-header-contain`

1. Check out the branch and execute npm run build
2. Test the theme with style pack 5 and check the menu with and without custom colours

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -340,7 +340,7 @@ function newspack_custom_colors_css() {
 		';
 	}
 
-	if ( true === get_theme_mod( 'header_solid_background', false ) && ! newspack_is_active_style_pack( 'style-3', 'style-4' ) ) {
+	if ( true === get_theme_mod( 'header_solid_background', false ) && ! newspack_is_active_style_pack( 'style-3', 'style-4', 'style-5' ) ) {
 		$theme_css .= '
 			.header-solid-background .middle-header-contain {
 				background-color: ' . $primary_color . ';
--- a/sass/styles/style-5/style-5.scss
+++ b/sass/styles/style-5/style-5.scss
@@ -31,6 +31,7 @@ body {
 .header-solid-background {
 	.site-header,
 	.bottom-header-contain,
+	.middle-header-contain,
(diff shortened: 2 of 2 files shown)


laurelfulford:
Actually, looking at it again, I have a question:

Is the solid header always meant to be black? 

Right now, it will be if you have the default colour palette, but if you change it it doesn't:

Default colours:

image

Custom colours:

image

If this is intentional, don't worry about it! 

If not, it can be added to the style packs in this  `newspack_is_active_style_pack()` function, to exclude this style pack from changing the header colour with the custom colours:

https://github.com/Automattic/newspack-theme/blob/master/inc/color-patterns.php#L343

thomasguillot:
> Is the solid header always meant to be black?
>
> Right now, it will be if you have the default colour palette, but if you change it it doesn't

It is intentional, yes, but I'm open for feedback. Do you think it looks like a bug?

laurelfulford:
> It is intentional, yes, but I'm open for feedback. Do you think it looks like a bug?

I think to sync it up to the other two style packs that have a neutral header colour, instead of the primary one, I would make it black all the time. But _that_ behaviour is also a bit unpredictable/bug-like, so I could just as easily argue it the other way :) 

Part of the reason I went that way was because it seemed unlikely for a site to use the default colour, and I wanted to keep the header neutral. However, those style packs use the primary colours a lot more -- I think as long as it's behaving as designed, it's fine.

thomasguillot:
> I think to sync it up to the other two style packs that have a neutral header colour, instead of the primary one, I would make it black all the time. But that behaviour is also a bit unpredictable/bug-like, so I could just as easily argue it the other way :)
> 
> Part of the reason I went that way was because it seemed unlikely for a site to use the default colour, and I wanted to keep the header neutral. However, those style packs use the primary colours a lot more -- I think as long as it's behaving as designed, it's fine.

Yeah good point. I removed the background colour in 6a2e69e

PR #494, opened 2019-10-10
Fix issue with spacing at the bottom of columns with borders.

laurelfulford:
### All Submissions:

In a previous PR, I removed the space from the bottom of the last element in a column block, to reduce vertical space. This works well, except when paired with the column border style on mobile. Then you get this:

image

This PR adjusts the styles, so when you're using the border styles there's still space after the last item, but only when viewed on small screens:

image

1. Apply this PR and run `npm run build`
2. Copy paste [this test content](https://cloudup.com/c5clcPMG9-8) into a post or page -- it includes two column blocks, one with the border style and one without.
3. View on a desktop-sized screen -- confirm the last item in each column has no bottom margin.
4. Shrink the browser window until the columns wrap; in the block with borders, confirm the last element in each block has a margin, except for the last item in the last column (since it doesn't have a bottom border). For the block without borders, confirm that the last item in each column still doesn't have a bottom margin.

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -732,7 +732,7 @@
 		&.is-style-borders {
 			.wp-block-column {
 
-				& > *:last-child {
+				&:not(:last-child) > *:last-child {
 					margin-bottom: 32px;
 				}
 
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks!

PR #499, opened 2019-10-11
Add a footer menu area

### LINKED ISSUE #457: Add footer menu
From @WPArena in #436:

> Option to select a menu for footer bottom right.

laurelfulford:
### All Submissions:

This PR adds a menu to the footer area. 

Closes #457.

2. Navigate to Customize > Menus; confirm there's a spot for the menu in the footer.
3. Add a menu.
4. View the footer on the front-end, adjusting the screen size; confirm that there are no visual issues.

image

image

image

### Other information:

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -21,6 +21,24 @@
 		<?php get_template_part( 'template-parts/footer/footer', 'widgets' ); ?>
 
 		<div class="site-info">
+
+			<?php if ( has_nav_menu( 'footer-menu' ) ) : ?>
+				<div class="wrapper">
+					<nav class="footer-menu">
+						<?php
+							wp_nav_menu(
+								array(
+									'theme_location' => 'footer-menu',
+									'container'      => false,
+									'items_wrap'     => '<ul id="%1$s" class="%2$s">%3$s</ul>',
+									'depth'          => 1,
+								)
+							);
+						?>
+					</nav>
+				</div><!-- .wrapper -->
+			<?php endif; ?>
+
 			<div class="wrapper">
 				<?php $blog_info = get_bloginfo( 'name' ); ?>
 				<?php if ( ! empty( $blog_info ) ) : ?>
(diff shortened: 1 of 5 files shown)


laurelfulford:
The end goal with is is ultimately be able to pair it with a space to place 'sponsor' logos (a good example is to display logos for sites that are part of Civil).

For flexibility, I imagine this area might be best set up as a widget area (so you can add one or more logo, or potentially, a string of text if needed), but I'm definitely open to suggestions for that! 

Here are some roughs I put together for this; the 'Civil' logo represents where the widget/other content spot would be; it would appear before this new menu, but if it wasn't assigned, the menu would be left-aligned: 

Footer: ALL THE THINGS
Footer: ALL THE THINGS2
Footer: ALL THE THINGS3
Footer: ALL THE THINGS4
Footer: ALL THE THINGS5
Footer: ALL THE THINGS6
Footer: ALL THE THINGS7
Footer: ALL THE THINGS8

thomasguillot:
The PR works as expected but I'm not sure I fully understand the need though. Maybe a widget area with elements set to `justify-content: space-between;` on larger screens might be best?

claudiulodro:
I also think a widget area would work well; it's more flexible than a nav. 👍 If nav elements are needed, the user can just use the navigation widget.

laurelfulford:
Ah, good call you two :) I over-complicated this. 

I'll rework it as a second footer widget area; and style things that would normally display as lists inline, so this kind of thing can be achieved with less fuss.

laurelfulford:
I've reworked this PR so it just adds a single widget area; the widgets will be displayed spaced out across the available space. 

Lists in this widget area are styled so they display inline, so if you add a Navigation widget, the links will display side by side.

Let me know if this makes more sense, and if you run into any issues - thanks!

laurelfulford:
Thanks Thomas -- good catch! 

I've 'flattened out' the lists, and tweaked some other styles (min-width for widgets, making sure a widget title at least sits inline) in bb49088 and 5aa2649.

Let me know what you think, and if any further changes are needed!

laurelfulford:
Thanks Thomas & Claudiu! I think this solution is way nicer/more flexible!

PR #503, opened 2019-10-12
Make buttons display inline-block.

laurelfulford:
### All Submissions:

This PR adds an `display: inline-block` to the `.button` class and `button` tags.

The purpose is to fix a couple weird issues I've seen on at least one site:

1. If you use a link with the `.button` class as the last item in the sidebar, and hit a page with a longer sidebar, it can get cut off.

image

2. When elements with the `.button` class are the last in a section, because they are `display: inline` right now, it can affect (reduce) spacing. For example, these are separate widgets and they should not be this close together on smaller screens when they stack, but the button doesn't really hang on to its bottom margin:

image

1. Create a long sidebar, finishing with a link with the `.button` class.
2. View it on the front-end: 

image

3. Apply the PR and run` npm run build`.
4. Confirm the button no longer looks weirdly smooshed:

image

### Other information:

### DIFF
--- a/sass/forms/_buttons.scss
+++ b/sass/forms/_buttons.scss
@@ -9,6 +9,7 @@ input[type="submit"] {
 	border: none;
 	border-radius: 5px;
 	box-sizing: border-box;
+	display: inline-block;
 	color: $color__background-body;
 	font-family: $font__heading;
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks!

Issue #350, opened 2019-08-28
Author archives: different style


laurelfulford:
Came in from feedback on the sites: 

* Could show more written information about the authors.
* Could show author social links
* Could show author title, photo

Note: having #236 completely resolved -- having fields for the different author information we want to show (out of the box, WordPress has name, email, short bio, gravatar... but not title, or an extensive social links list, or a fuller bio).

laurelfulford:
Closing, as most of this is resolves, minus the job title, which we can work around now by adding it to the description if needed. Can open a new issue specific to that if it comes up again.

PR #506, opened 2019-10-13
Add author avatar to the archive page

laurelfulford:
### All Submissions:

This PR adds the author avatar, if present, to the author archive page.

See #350.

2. Confirm the author avatar now appears on the author archive pages, next to the information about them:

image

image

image

3. Check the appearance on different screensizes.
4. Navigate to Settings > Discussion, and hide avatars; confirm this removes them from the author archive page, too.

### Other information:

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -14,9 +14,33 @@ get_header();
 
 		<header class="page-header">
 			<?php
-				the_archive_title( '<h1 class="page-title">', '</h1>' );
-				the_archive_description( '<div class="taxonomy-description">', '</div>' );
+				if ( is_author() ) {
+					$author_id     = get_query_var( 'author' );
+					$author_avatar = get_avatar( $author_id, 120 );
+
+					if ( $author_avatar ) {
+						echo wp_kses(
+							$author_avatar,
+							array(
+								'img' => array(
+									'src'    => array(),
+									'alt'    => array(),
+									'class'  => array(),
+									'width'  => array(),
+									'height' => array(),
+								),
+							)
+						);
+					}
+				}
 			?>
+			<span>
+				<?php
+					the_archive_title( '<h1 class="page-title">', '</h1>' );
+					the_archive_description( '<div class="taxonomy-description">', '</div>' );
+				?>
+			</span>
+
 		</header><!-- .page-header -->
 
 		<main id="main" class="site-main">
--- a/sass/site/primary/_archives.scss
+++ b/sass/site/primary/_archives.scss
@@ -78,6 +78,30 @@
 			}
 		}
 	}
+
+	&.author .page-header {
+		display: flex;
+
+		.avatar {
+			height: 30px;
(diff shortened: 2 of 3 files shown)


archive.php
```diff
@@ -14,9 +14,33 @@
...
+							array(
+								'img' => array(
+									'src'    => array(),
+									'alt'    => array(),
+									'class'  => array(),
+									'width'  => array(),
+									'height' => array(),
+								),
+							)
+						);
```
thomasguillot: I don't think you need `wp_kses()` here. A simple `echo $author_avatar` should do the trick.

Also for some reason, with Jetpack (and image lazy loading) enabled, the avatar is being displayed twice: [see screenshot](https://cld.wthms.co/YrxfPw) but it's fine when using `echo $author_avatar`

thomasguillot: Note: The avatar being displayed twice appears everywhere.

https://cld.wthms.co/8QllP6

https://cld.wthms.co/UJRlWr

laurelfulford: Thanks Thomas! This should be fixed in 8f62fd7.

I was getting a bit too clever about avoiding a PHPCS error; echoing just `get_avatar()` works just as well. I'll fix that elsewhere in the site in a separate PR.

laurelfulford:
I've fixed the avatar duplication issue, as well as a couple other things:

For some reason the avatar was squished on load in Firefox, then would correct itself when I inspected it. I added a `flex-shrink` value to fix this.

I completely forgot that the archive headers were centred in style 4, so I also added some adjustments there to make the avatar display better.

archive.php
```diff
@@ -19,18 +19,7 @@
...
-									'src'    => array(),
-									'alt'    => array(),
-									'class'  => array(),
-									'width'  => array(),
-									'height' => array(),
-								),
-							)
-						);
+						echo get_avatar( $author_id, 120 );
 					}
```
thomasguillot: We can simplify this with: `if ( $author_avatar ) echo $author_avatar;` rather than duplicating `get_avatar( $author_id, 120 );`

laurelfulford: Ah good call - thanks Thomas! This has been updated in 3950ae7 .

laurelfulford:
Thanks Thomas!

I meant to add this in my note, but we have a PHPCS check that runs on commit, so I couldn't place the function all on one line or else it gets cranky; I also added a line of code so it would ignore `$author_avatar` (which it threw an error for:

` All output should be run through an escaping function (see the Security sections in the WordPress Developer Handbooks), found '$author_avatar'. (WordPress.Security.EscapeOutput.OutputNotEscaped)`

thomasguillot:
> I meant to add this in my note, but we have a PHPCS check that runs on commit, so I couldn't place the function all on one line or else it gets cranky; I also added a line of code so it would ignore `$author_avatar`

Weird 😅 Anyway, this looks great! 😊

PR #508, opened 2019-10-14
Fix avatar duplication

laurelfulford:
### All Submissions:

This PR fixes an issue reported in the review of #506:

The theme's using a `wp_kses()` function unnecessarily with the avatars, causing them to be duplicated when using Jetpack's lazy loading.

`get_avatar()` doesn't actually need this sanitization.

1. Apply the PR.
2. Check a single post; confirm the author avatar is displaying as expected in the byline at the top.
3. Check the author bio at the bottom of the post; confirm the author avatar is displaying as expected.
4. Navigate to Customize > Style Packs and switch to Style 4 (or if you're already on style 4, switch away from it). 
5. Re-confirm that the author bio is displaying as expected. 

### Other information:

### DIFF
--- a/template-parts/post/author-bio.php
+++ b/template-parts/post/author-bio.php
@@ -9,45 +9,23 @@ if ( (bool) get_the_author_meta( 'description' ) && is_single() ) : ?>
 <div class="author-bio">
 
 	<?php
-	if ( ! newspack_is_active_style_pack( 'style-4' ) ) :
+	if ( ! newspack_is_active_style_pack( 'style-4' ) ) {
 		$author_avatar = get_avatar( get_the_author_meta( 'ID' ), 80 );
-		if ( $author_avatar ) :
-			echo wp_kses(
-				$author_avatar,
-				array(
-					'img' => array(
-						'src'    => array(),
-						'alt'    => array(),
-						'class'  => array(),
-						'width'  => array(),
-						'height' => array(),
-					),
-				)
-			);
-		endif;
-	endif;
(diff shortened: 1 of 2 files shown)


template-parts/post/author-bio.php
```diff
@@ -9,45 +9,23 @@
...
-						'width'  => array(),
-						'height' => array(),
-					),
-				)
-			);
-		endif;
-	endif;
+		if ( $author_avatar ) {
+			echo get_avatar( get_the_author_meta( 'ID' ), 80 );
+		}
```
thomasguillot: You can simply write `if ( $author_avatar ) echo $author_avatar;`

template-parts/post/author-bio.php
```diff
@@ -9,45 +9,23 @@
...
-								'width'  => array(),
-								'height' => array(),
-							),
-						)
-					);
-				endif;
-			endif;
+				if ( $author_avatar ) {
+					echo get_avatar( get_the_author_meta( 'ID' ), 80 );
+				}
```
thomasguillot: Same as above, you can simply write `if ( $author_avatar ) echo $author_avatar;`.

laurelfulford: Thanks Thomas! 

This should be fixed in 9d98b51; like in #506, the PHPCS dictates the format so it's not quite as concise as your example, but it's better that we don't have to repeat the avatar size twice -- that's just a recipe for messing something up :)

Issue #174, opened 2019-08-08
Style Pack: Default styles not dequeued in Customizer


laurelfulford:
This issue results in a similar outcome to #52 -- inaccurate customization previews -- but it's not the same cause:

When you switch to a different style pack, the 'default' styles are still there. So some elements -- especially as we move more opinionated styles into a default-specific Sass file -- aren't correctly overridden. 

This came up in #164 as well as a previous issue; it's confusing because what you see in the Customizer looks strikingly different than it does on the front end when looking at the new 'style pack' styles.

laurelfulford:
Closing this one, since we've moved to style packs.

PR #511, opened 2019-10-15
Styles: Make Style 5 observe the custom colours

philipjohn:
On article pages when using Style 5, the category links above the title and other areas do not use the custom colours as selected in the customiser. Instead they use the default Newspack colours.

This change ensures they use the customiser colours, just like other style packs.

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -340,6 +340,21 @@ function newspack_custom_colors_css() {
 		';
 	}
 
+	if ( newspack_is_active_style_pack( 'style-5' ) ) {
+		$theme_css .= '
+			.cat-links a,
+			.cat-links a:visited,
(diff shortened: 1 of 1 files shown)


thomasguillot:
Looks good and worth merging but I will work on a separate PR to fix the underlying issue since this style shouldn't have any background colour for the category links

laurelfulford:
> since this style shouldn't have any background colour for the category links

This should be fixed on the staging site it was occurring on; it turned out to be caused by a plugin.

There is a known issue with the default styles displaying in the Customizer (#174), so the backgrounds on the categories may still appear there, but should not show up on the front-end.

thomasguillot:
Looks like @laurelfulford fixed it so we shouldn't need this PR anymore 😊

PR #513, opened 2019-10-15
Fix up some of the styles in the new footer widget area

laurelfulford:
### All Submissions:

I added a new footer widget area in #499, intended primarily for simple menus, disclaimer text, or 'sponsor' logos. 

I noticed on implementation that it had some display issues: 

1. If you had a menu or category widget as the second & last widget, there was a small space between it and the right edge (due to a right margin on each list item):

2. If you put a short/smaller widget in the second/last widget area, the `min-width` was obvious -- it wouldn't go all the way to the end.

Removing the min-width does mean that it will be important to make sure the widget content is relatively even, but I think this is easier to control on the WordPress UI level; we can also make further adjustments as needed.

Closes # .

1. Navigate to Customize > Widgets > Above Copyright, and add a Text widget with a short string of text.
2. Under it, add a Navigation Menu widget, and assign a short, but multi-level, menu to the Navigation Menu widget. 
3. View on the front-end; note that there's a slight space between the last menu item in the widget, and the end of the availalbe space:

image

4. Navigate back to Customize > Widgets Above Copyright, and switch the widget locations.
5. View on the front-end; note that the shorter widget has a minimum width, so it doesn't reach the right side:

image

6. Apply the PR and run `npm run build`
7. Confirm the short widget sits all the way to the right.

image

8. Switch the widget locations again, confirm that the menu goes all the way to the right.

image

### Other information:

### DIFF
--- a/sass/site/footer/_site-footer.scss
+++ b/sass/site/footer/_site-footer.scss
@@ -113,8 +113,15 @@
 			font-size: inherit;
 
 			@include media( mobile ) {
-				margin-bottom: 0;
-				min-width: 25%;
+				margin: 0 $size__spacing-unit;
+
+				&:first-child {
+					margin-left: 0;
+				}
+
+				&:last-child {
+					margin-right: 0;
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks!

PR #521, opened 2019-10-21
Update check for image size to add body class

laurelfulford:
### All Submissions:

There's a check in the theme that adds a class to the body, `.has-large-featured-image`, if the featured image is more than 1200px.

It's used to add a top border when the image is not; however, the check is `1200 < width` (greater than), but it should be `1200 <= width` (great than or equal to 1200px). Otherwise, if you upload an image that's exactly 1200px wide, it will show both the large image, and an unnecessary border above it.

1. Set up a post and add a featured image that's exactly 1200px wide; view on the front-end: 

image

2. Set up a second post and add a featured image that's less than 1200px; view on front-end:

image

3. Note that both have a border underneath the `.entry-header` (if using Style 3, it's easier to see the border). 

5. Confirm that the first post no longer has the border, but the second one still does:

image

image

### Other information:

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -82,7 +82,7 @@ function newspack_body_classes( $classes ) {
 
 	// Adds a class if singular post has a large featured image
 	$thumbnail_info = wp_get_attachment_metadata( get_post_thumbnail_id() );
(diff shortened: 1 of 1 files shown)


laurelfulford:
🙌

PR #523, opened 2019-10-22
PWA: Add 'offline.php' template

### LINKED ISSUE #357: Add PWA error template
The theme needs a customized PWA error template, it needs to be helpful and fallback gracefully. 

So we need to figure out:
* What it should have; what would be really helpful
* How best to carry over the current site's customizations

laurelfulford:
### All Submissions:

This PR is a first pass at adding a customized PWA error template to the theme.

It should use the theme's branding, and display a PWA error message, like:

IMG_9259

On top of the actual code, I'd be interested to hear feedback about what elements actually make sense to have in this template. Right now, for example, it loads the menu, footer widgets, and ads. Would it be better to strip that kind of stuff out? 

Once the template itself looks okay, we can look at jazzing it up a bit (like maybe adding headlines for the five most recently published articles?).

@westonruter I would especially appreciate your expertise on that part! 

Closes #357.

2. Make sure the PWA plugin is installed and enabled.
3. Test the presence of the PWA error message in one of two ways:

a) You can add `?wp_error_template=offline` to the URL, to see the template and page title (but not actual error message).

**OR** 

a) On a mobile device, view the site and add it to the Home Screen. 
b) Switch the phone to airplane mode.
c) Try loading the site from the link on the homepage; the initial page you saved should still work, but if you click another link that hasn't been saved, you should see the error message.

### Other information:

### DIFF
--- a/offline.php
+++ b/offline.php
@@ -9,13 +9,16 @@ get_header();
 ?>
 	<section id="primary" class="content-area">
 		<header class="entry-header">
-			<h1 class="entry-title"><?php esc_html_e( 'Error', 'newspack' ); ?></h1>
+			<h1 class="entry-title"><?php esc_html_e( 'Oops! It looks like you&rsquo;re offline.', 'newspack' ); ?></h1>
 		</header><!-- .entry-header -->
 		<main id="main" class="site-main">
-			<?php wp_service_worker_error_message_placeholder(); ?>
+			<?php
+			if ( function_exists( 'wp_service_worker_error_message_placeholder' ) ) {
+				wp_service_worker_error_message_placeholder();
+			}
+			?>
 		</main><!-- .site-main -->
 	</section><!-- .content-area -->
-
 <?php
 get_footer();
 
--- a/offline.php
+++ b/offline.php
@@ -1,0 +1,26 @@
+<?php
+/**
+ * The main template file
+ *
+ * This is the most generic template file in a WordPress theme
+ * and one of the two required files for a theme (the other being style.css).
+ * It is used to display a page when nothing more specific matches a query.
+ * E.g., it puts together the home page when no home.php file exists.
+ *
+ * @link https://developer.wordpress.org/themes/basics/template-hierarchy/
+ *
+ * @package Newspack
+ */
+
+get_header();
+?>
+	<section id="primary" class="content-area">
+		<main id="main" class="site-main">
+			<?php wp_service_worker_error_message_placeholder(); ?>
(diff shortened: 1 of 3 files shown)


amedina:
Is the red color part of the theme or is it being used only for the error page?

If the latter is the case, you could experiment with a color/design that aligned with the theme palette, so that the error condition is experienced more gracefully by the user. One example I love is the [404 error page](https://slack.com/fsdfds) in `slack.com`. It does not feel as an error condition.

laurelfulford:
> Is the red color part of the theme or is it being used only for the error page?

Ah, that was not a great choice for that screenshot, but it was coming from the settings I had on my test site (a solid background on the header, plus a very red primary colour). 

I've attached a couple more screenshots using different colours/settings, to get a better feel of what it would actually look like with different settings:

IMG_9261

IMG_9260

offline.php
```diff
@@ -0,0 +1,25 @@
...
+ * The PWA Offline template file
+ *
+ * @package Newspack
+ */
+
+get_header();
+?>
+	<section id="primary" class="content-area">
+		<header class="entry-header">
+			<h1 class="entry-title"><?php esc_html_e( 'Oops! It looks like you&rsquo;re offline.', 'newspack' ); ?></h1>
```
westonruter: Technically the server could also be offline. See https://github.com/xwp/pwa-wp/pull/190

So maybe just “Offline” would be better.

```suggestion
			<h1 class="entry-title"><?php esc_html_e( 'Offline', 'newspack' ); ?></h1>
```

laurelfulford: Ah, good call! This has been updated in f92f9c1.

laurelfulford:
Thanks @westonruter! 

I've updated the template to remove menus and widgets in 48f8181, as well as WordAds; I need to do a little more work to get the ads added by the Newspack Plugin also removed.

> Then once the PWA plugin facilitates listing the offline pages available (xwp/pwa-wp#211), these can be added to the content of the template, allowing the user to navigate to those previously-cached URLs which we know are available (as opposed to the nav menu and widget links which may not be).

Right! Yes, that'd be a lot more useful :)

laurelfulford:
Thanks @jeffersonrabb! 

>  I assume this is the following, which will be addressed separately elsewhere?

I actually meant to continue working on it in this PR, but I didn't make that very clear! I pushed up what I had until I could get `remove_action` working with the ads coming from Newspack. 

Since this is approved, I can merge this PR and work on that in a separate one.

PR #540, opened 2019-10-29
Add check for menus before outputting mobile menu toggle

laurelfulford:
### All Submissions:

This PR checks for a primary, secondary or tertiary menu before outputting a 'Menu' toggle on mobile screens.

It's largely to make sure an empty menu space isn't added to the PWA offline.php template, but it also makes sense in the odd case where a site might not have a menu to open.

1. Apply the PR.
2. Test the offline.php template, and make sure the 'Menu' toggle can't be seen on smaller screens (one way to test is to add `?wp_error_template=offline` to the end of your site URL).
3. Make sure the menu toggle still appears on other pages on smaller screens.
4. Empty the primary, secondary and tertiary menu spaces and confirm the menu toggle no longer appears on any page. 

### Other information:

### DIFF
--- a/inc/template-tags.php
+++ b/inc/template-tags.php
@@ -243,6 +243,15 @@ if ( ! function_exists( 'newspack_the_posts_navigation' ) ) :
 endif;
 
 /**
+ * Check if any header menus are applied; used to show menu toggle on smaller screens.
+ */
+function newspack_has_menus() {
+	if ( has_nav_menu( 'primary-menu' ) || has_nav_menu( 'secondary-menu' ) || has_nav_menu( 'tertiary-menu' ) ) {
+		return true;
+	}
+}
+
+/**
  * Displays primary menu; created a function to reduce duplication.
  */
 function newspack_primary_menu() {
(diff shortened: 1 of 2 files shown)


inc/template-tags.php
```diff
@@ -225,6 +225,15 @@ function newspack_the_posts_navigation() {
...
 endif;
 
+/**
+ * Check if any header menus are applied; used to show menu toggle on smaller screens.
+ */
+function newspack_has_menus() {
+	if ( has_nav_menu( 'primary-menu' ) || has_nav_menu( 'secondary-menu' ) || has_nav_menu( 'tertiary-menu' ) ) {
+		return true;
+	}
+}
```
claudiulodro: This should probably return `false` if the check fails, as opposed to `null`. Currently, you'll have issues if you try and do something like `false === newspack_has_menus()`

laurelfulford: Oh, good call! Added in 3c5e497.

laurelfulford:
Thanks!

PR #541, opened 2019-10-29
Style 2: Update dropcap style

laurelfulford:
### All Submissions:

We had a request to simplify the Style 2 dropcap, removing the border, and making it use the secondary colour instead.

Since the style pack is only used on one live site, and it's not using dropcaps, this seems like a safe time to make this change.

1. Navigate to Customize > Style Packs and switch to Style 2.
2. In a post or page, add a paragraph block, and switch on the dropcap -- it should look something like this, with smaller text and a border using the primary colour:

image

3. Apply the PR and run` npm run build`.
4. View the dropcap in the editor and on the front-end -- it should now be borderless, larger, and using your secondary colour:

image

5. If your test site isn't already using custom colours, switch to custom colours and change the secondary colour. Note: because the dropcap is text against a white background, it uses the 'contrast' functionality, which swaps out too-light colours with a mid-grey. As an aside, this is _very_ aggressive right now and it'd be nice to allow a bit more range of colours, but for now it's something to keep in mind; if you pick a lighter colour, it could appear not to work. 

### Other information:

### DIFF
--- a/inc/color-patterns.php
+++ b/inc/color-patterns.php
@@ -280,7 +280,7 @@ function newspack_custom_colors_css() {
 			}
 
 			.has-drop-cap:not(:focus)::first-letter {
-				border-color: ' . $primary_color . ';
+				color: ' . newspack_color_with_contrast( $secondary_color ) . ';
 			}
 		';
 	}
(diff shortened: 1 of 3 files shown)


laurelfulford:
Thanks Claudiu!

PR #572, opened 2019-11-12
Cover Block: Simplify front-end styles

laurelfulford:
### All Submissions:

This PR removes the `.entry .entry-content` selector from the Cover block styles.

1. Copy-paste [this test content](https://cloudup.com/cNe7Yz8Srp9) into the editor, and publish the post.
2. View the different cover blocks on the front-end and note their appearance.
4. View the cover blocks on the front-end again and confirm no issues were introduced. 

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -168,6 +168,14 @@
 	position: relative;
 }
 
+
+//! Newspack Article Block
+.wpnbha {
+	figcaption {
+		max-width: 100%;
+	}
+}
+
 //! Columns
 .wp-block-columns {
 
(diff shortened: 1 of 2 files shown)


laurelfulford:
Oh, good catch, Jeff! 

I think the original is the error -- that title is centred otherwise, and I can't think why it would be left-aligned only when nested in the cover block. I _can_ see the style causing it -- it was to override some of the block's older default styles, from when it included an h2 and paragraph block in it. 

I tidied up the styles a bit to correct this in the editor, and circled through the other style packs to make sure these headers were otherwise displaying as expected. Just let me know if anything else doesn't look correct!

laurelfulford:
Thanks!

PR #579, opened 2019-11-19
Columns Block: Fix issue when nested in Group block

### LINKED ISSUE #578: Column Blocks in Group Blocks are too wide
When you nest a column block inside of a Group block, it's a bit wider than usual, with the content sitting closer to the right side than the left:

image

The problem is even more obvious when the column block uses the border styles -- the content pushes past the right edge a bit:

image

This may be related to the updated styles we gave the column block; in that case, it's probably best fixed in the Newspack Blocks plugin than the theme -- needs some more investigation!

laurelfulford:
### All Submissions:

A recent change to the Group block made it so, when you nested a column block inside of it, it would hit the right side.

This PR fixes that. 

Closes #578.

1. Copy-paste [this test content into a page](https://cloudup.com/cmRCU2GTFhP) -- it includes two group block + columns block, with article blocks inside; one of the columns blocks has borders, the other does not.
2. View on the front-end; note extra spacing on the left, and the little-to-no spacing on the right:

image

image

4. Confirm that the spacing is now consistent:

image

image

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -172,7 +172,8 @@
 .wp-block-columns {
 
 	.wp-block-column > * {
-		margin: 32px 0;
+		margin-bottom: 32px;
+		margin-top: 32px;
 
 		&:first-child {
 			margin-top: 0;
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks Claudiu!

> That cat picture in your screenshots is amazing!!

Cute animal photos makes testing bugs that much better! 😄

PR #608, opened 2019-12-03
Add option to hide featured images on individual posts

### LINKED ISSUE #604: "Hidden" featured image option

We've had requests for the ability to hide the featured image on posts, so people can have thumbnails, etc. for posts but not show the image within the post (and potentially "replace" it with an image block at the top of the post content).

I think a new "Hidden" display option in the featured image options would work well without needing anything super complex.

The featured image option seems the easiest way to go IMO. I'm open to other ideas if anyone has them, though.

laurelfulford:
### All Submissions:

This PR adds a 'position' option to hide the Featured Image on individual posts. 

Closes #604 .

2. Create a post; add a featured image, and set it to 'Hidden'; leave it on the default template.
3. Confirm that the image is hidden on the front-end, and that the border that divides the title and content is present:

image

4. Switch to the One Column template; confirm it's the same.
5. Switch to the One Column Wide template; confirm it's the same.
6. Try a few of the different featured image setting with these two templates -- 'Small', 'Behind' and 'Default', to make sure no display issues are introduced.

### Other information:

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -75,13 +75,15 @@ function newspack_body_classes( $classes ) {
 		$classes[] = 'has-sidebar';
 	}
 
+	$current_featured_image_style = get_post_meta( get_the_ID(), 'newspack_featured_image_position', true );
+
+
 	// Adds class if singular post or page has a featured image.
-	if ( is_singular() && has_post_thumbnail() ) {
+	if ( is_singular() && has_post_thumbnail() && 'hidden' !== $current_featured_image_style ) {
 		$classes[] = 'has-featured-image';
 	}
 
 	// Adds a class to single artcles, if they're using a special featured image style.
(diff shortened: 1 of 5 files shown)


akkiverma:
Please provide this option as site-wide. Like if I turned-off site-wide then I can enable it for particular posts where required. It'll be super useful when one is migrating from another theme with hundreds of posts already published.

claudiulodro:
Thanks for the quick turnaround on this feature! I think our publishers will be happy with it.

laurelfulford:
Thanks for the review, @claudiulodro! 

> Hopefully we don't need any more featured image styles because that logic is getting complicated. :)

Hahah yeah... I'm looking at setting a default, and I think that's going to be the final one -- it's going to make it a bit worse. But I think once you can hide the featured image, and set that as your default, you're pretty much free to do anything else you want, image-wise, and we shouldn't need anything else in the theme :)

PR #613, opened 2019-12-04
Add 'default' option for featured image position

### LINKED ISSUE #609: An option to make Featured Posts "Hidden" or "Any Other" for all posts Site Wide
I see that the option to hide Featured Image has been approved and will be included in the future updates. But there is also a need to have this feature to be enabled site-wide too so that the publishers don't have to open each existing post and set the required option. or this way it'll be way better if the site-wide feature image options may contain all of the existing options like:

- Default
- Small
- Behind
- Beside
- Hidden

Please consider providing this option. It'll be a very big help.

laurelfulford:
### All Submissions:

This PR adds a 'default' option for the featured image placement in the Customizer, under Featured Image Settings.

Unfortunately, it's not possible to set a default for post meta, but I think that that would muddy things when switching between options (was this post set to 'small' by someone editing the page, or was that the default setting and it should be changed when the default is switched?). So this might for best. 

It does make for some disconnect, though -- if this otherwise seems like a good approach, we could probably pass the current default to the API and have it appear on a per-post basis next to the 'Default' label -- eg. `Default (set to "Large" in the Customizer)`. If not, I'm definitely open to trying different approaches! 

As part of this update, I also created a function that sets all the 'featured image position' logic in one spot -- hopefully that helps make the code a little easier to follow!

Lastly, I removed the option under 'Content Options' to hide featured images on single posts, since this will overlap that behaviour. 

Closes #609.

1. Set up a couple posts using the featured image position options like 'Small', 'Beside' and 'Hidden', and one with just the default setting (this is to help test that existing options stick after this change is applied). 
2. Apply the PR and run `npm run build`
3. Check your posts; confirm that they still have their featured images set.
4. Navigate to Customizer > Featured Image Settings, and confirm the default is currently set to 'Large'.
5. Pick another default. 
6. Cycle through your example posts again, and confirm that all but the 'default' one are the same -- the one using the 'default' image option should now be using whatever was set in the Customizer. 
7. Edit the post you created that uses the default image setting; confirm that it's still set to 'Default', but a new 'Large' option also displays:

image

8. Try switching it to 'Large' and confirm it displays with the original default display, overriding whatever you have set in the Customizer. 

A couple more notes:

1. If you use an image less than 1200px wide but pick "Large" or have "Large" as the default, it will still fall back to the 'Small' image display on the individual post where the image is too small (where it displays in the content area, with a max-width the same as the content area).
2. When using the 'wide' template, the 'small' image option doesn't really read as small -- since it just matches the content area, it can display up to 1200px wide. This is the same as it would've behaved prior to this PR, but I noticed it in my testing. 

### Other information:

### DIFF
--- a/inc/template-functions.php
+++ b/inc/template-functions.php
@@ -5,6 +5,47 @@
  * @package Newspack
  */
 
+if ( ! function_exists( 'newspack_featured_image_position' ) ) :
+	/**
+	 * Returns current post's featured image position.
+	 *
+	 * @return string
+	 */
+	function newspack_featured_image_position() {
+		// Get thumbnail
+		$thumbnail_info = wp_get_attachment_metadata( get_post_thumbnail_id() );
+		$img_width      = $thumbnail_info['width'];
+
+		// Get per-post image position setting.
+		$image_pos = get_post_meta( get_the_ID(), 'newspack_featured_image_position', true );
+
+		// Get default image position setting from the Customizer.
+		$default_image_pos = get_theme_mod( 'featured_image_default', 'large' );
+
+		// Set a position value to return.
+		$position = '';
+
+		if ( ! has_post_thumbnail() ) {
+			return;
+		}
+
+		// First, check for a per-post image setting.
+		if ( '' !== $image_pos ) {
+			$position = $image_pos;
+		// If this post doesn't have a setting, fall back to the default.
+		} else {
+			$position = $default_image_pos;
+		}
+
+		// Fallback to the small inline posiiton if the image isn't large enough to be, uh, large.
+		if ( 'large' === $position && 1200 > $img_width ) {
+			$position = 'small';
+		}
+
+		return $position;
+	}
+endif;
+
 /**
  * Adds custom classes to the array of body classes.
  *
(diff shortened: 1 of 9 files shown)


inc/template-functions.php
```diff
@@ -5,6 +5,48 @@
...
+if ( ! function_exists( 'newspack_featured_image_position' ) ) :
+	/**
+	 * Returns current post's featured image position.
+	 *
+	 * @return string
+	 */
+	function newspack_featured_image_position() {
+		// If we're not on a single page, or if there's no thumbnail, return.
+		if ( ! is_single() || ! has_post_thumbnail() ) {
+			return;
```
claudiulodro: This should probably return `''` instead of `null`.

js/src/extend-featured-image-editor.js
```diff
@@ -16,7 +16,8 @@ class RadioCustom extends Component {
 				label={ __( 'Featured Image Position' ) }
 				selected={ meta.newspack_featured_image_position }
 				options={ [
-					{ label: __( 'Default' ), value: '' },
+					{ label: __( 'Default (set in Customizer)' ), value: '' },
```
claudiulodro: I like the `Default (set to "Large" in the Customizer)` idea to make this label more useful. 👍

inc/template-functions.php
laurelfulford: Whoops -- this should be fixed in 57d5bd1!

js/src/extend-featured-image-editor.js
laurelfulford: Yeah, I think it will be most helpful! It's taking me a bit to figure out how to do it, though :) I think it might be best to get this PR merged and tackled that separately. Let me know what you think!

laurelfulford:
This is mostly ready to review again because I haven't managed to get the label updated to match the current default from the Customizer yet :) I think it may make sense to handle that in a separate PR (just for speed alone!), but let me know what you think, @claudiulodro!

laurelfulford:
Thanks!

PR #594, opened 2019-11-26
WIP: Make updates the theme to work better as a parent theme

laurelfulford:
### All Submissions:

* [ ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes # .

### How to test the changes in this Pull Request:

1.
2.
3.

### Other information:

* [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #630, opened 2019-12-11
Update theme/folder structure for move to parent/child themes

laurelfulford:
### All Submissions:

This PR updates the theme folder structure to nest the actual Newspack Theme inside of a /newspack-theme directory, and adjusts the build processes to generate the release zip into a different location, and add a couple new build processes that can be built out to include child themes as they're added. 

Related #594.

**Updated Dec 16** to improve testing instructions.

1. Clone the repository directly into the `/themes` folder using `git clone git@github.com:Automattic/newspack-theme.git .`. You will need to empty the directory first; if you still get the error `fatal: destination path '.' already exists and is not an empty directory.`, you can check for hidden files using `ls -a`.
Alternatively, you can clone the repo into a location separate from your themes directory, and symlink the nested 'newspack-theme' directory to a wp-content/themes directory on your test site. 
2. Apply the PR.
3. In the root of the directory, run `npm install && composer install`.
4. In the same location, try running `npm run build` and `npm run release:archive`. 
5. Confirm the theme is running as expected on your site.
6. Confirm that a zip has been generated in the `/releases` directory. 

Note: with the current structure, you can only run the build from the root directory, not from within the theme (since each of those directories don't have dependencies installed). It would work to install dependencies there, too, but we probably do want to build _all_ the themes every time we need to build the themes, so the current setup reflects the ideal processes. 

### Other information:


akkiverma:
Does it mean that the theme now support using child theme?

laurelfulford:
Hi @akkiverma -- the theme doesn't _not_ support child themes at the moment :) We're looking into moving the style pack functionality into separate child themes, and keeping all of the themes in this repo, which is what this draft PR and #594 are related to. It may take some time for them to actually land though.

laurelfulford:
Have rebased this PR, and moved the NPM dependencies into the package.json in the root, so there's no need to run `npm install` in each theme. Will update the testing steps to reflect that.

newspack-theme/package.json
```diff
@@ -0,0 +1,39 @@
+{
```
jeffersonrabb: Instead of a `package.json` in each child theme directory could the build scripts all live in the `package.json` at the root? I tried the following and it seemed to work:

- Copied the `scripts` object from `./newspack-theme.package.json` over to `./package.json`. 
- Prepended `newspack-theme/` to all paths in the build scripts.
- Moved `newspack-theme/babel.config.js`, `newspack-theme/postcss.config.js`, and `newspack-theme/webpack.config.js` to the root.
- Deleted `newspack-theme/package.json` and `newspack-theme/package-lock.json`

Also had to make some change to the `import`s in the two Sass plugins, `newspack-ads.scss` and `woocommerce.scss`. Changed them to look like this:

```
@import "../../sass/variables-site/variables-site";
@import "../../sass/mixins/mixins-master";
```

If this works it would simplify the structure a bit and make for an easier integration of https://github.com/Automattic/newspack-theme/pull/640.

laurelfulford:
Heck, this took way longer than it should have to untangle the merge conflicts :) 

I've held off updating the README for now, but can circle back to that in a separate PR -- I was having a hard time getting the symlinks to work for some reason, and want to make sure I have the recommended steps correct before adding that.

laurelfulford:
Thanks for the test, Jeff!

From what I can tell, only the Post Subtitle JS files are being generated to `dist` on `npm run build` -- none of the others are 😕 

If I move the JS files to the root and update webpack.config.js to point there, they all get generated, so I'm sure it's something weird I've done here with updating the paths:

https://github.com/Automattic/newspack-theme/pull/630/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dL15-R25

The weird bit is, the second bit works with what's currently in the PR (the part that grabs the post subtitle files), so I'm not sure why the first one doesn't.

laurelfulford:
Got it! It was this change:

https://github.com/Automattic/newspack-theme/pull/630/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR17

In that case, 'js' was not a path, so adding 'newspack-theme' was breaking it -- switching it back to 'js' gets it working again 😅

laurelfulford:
Replaced with the (very unfortunately numbered!) #666, so the commits are easier to follow. 

The issue with the JS build above should be fixed there as well!

PR #634, opened 2019-12-16
Image block: Update styles to prevent display issues in mobile Safari

### LINKED ISSUE #547: Image Block: Initial load on mobile can push caption way down
Sometimes when loading a post on a mobile device, the caption on image blocks, when floated left or right, can initially be pushed down quite a bit. You can see where the caption is in this screenshot by looking for the little block underneath the caption:

IMG_9301

This often resolves after the page sits for a while. 

I'm also not sure if this is related a) the theme's image styles, b) related to AMP, or c) caused by the combination of the two. 

It also appears it can be fixed by switching the image block from `display: table` to `display: block`, and the `figcaption` from `display: table-caption` to `display: block`, at least on smaller screens. This should undergo some testing to make sure it doesn't cause other issues, though.

laurelfulford:
### All Submissions:

This PR switches the image block from using 

Closes #547.

Note: this has been a lot easier for me to recreate on a non-local site, possibly because of image load times, or some other factor.

1. Add an image block with a caption, and float it to the right. 
2. View in Safari using an iOS device, or using Safari on your laptop in responsive design mode.
3. Scroll to where the image is on the page; if the caption is displaying correctly, refresh the page a few times and you should get the caption stuck like this:

image

5. Refresh the page repeatedly again; confirm that the caption doesn't get stuck like above, or even flashes a layout like above:

image

6. Try the other image alignment options (left, center, wide, full, none) with and without captions to make sure that this update didn't introduce any new issues.

### Other information:

### DIFF
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -682,9 +682,7 @@
 		}
 
 		.alignleft > figcaption,
-		.alignright > figcaption,
-		.alignleft > figure,
-		.alignright > figure {
+		.alignright > figcaption {
 			display: block;
 		}
 	}
--- a/sass/blocks/_blocks.scss
+++ b/sass/blocks/_blocks.scss
@@ -680,6 +680,13 @@
 		figcaption {
 			text-align: left;
 		}
+
+		.alignleft > figcaption,
+		.alignright > figcaption,
+		.alignleft > figure,
+		.alignright > figure {
+			display: block;
+		}
 	}
 
 	//! Galleries


laurelfulford:
Oh, good catch @claudiulodro! 

I'm not able to recreate this one with this PR applied. Are you getting this with the PR, or the original Custom CSS? The Custom CSS was only applied to smaller screen sizes -- I'm _hoping_ that's why it's happening for you, but if not let me know and I can spin up a second issue/PR to address it :)

claudiulodro:
I was testing with the custom CSS. I bet that was the problem. Good catch!

claudiulodro:
Yep, with the latest theme release, the device rotate issue is fixed 👍

laurelfulford:
Yay! Thanks for confirming :)

PR #681, opened 2020-01-08
Improve spacing on accent headers, compared to article blocks.

### LINKED ISSUE #650: Vertical alignment of 'accent-header' vs section header
image

'Section header' -- part of the article block -- is on the left. 'Accent header' -- a special class for the header block -- is on the right. Ideally, these two options will have the same vertical spacing so they can be used interchangeably. 

A short-term work around for now is to use a header block with the `accent-header` class in lieu of the section header option in the Homepage Posts block.

laurelfulford:
### All Submissions:

This PR makes the spacing around the article block section header, and the theme's `.accent-header` class more consistent with each other

Closes #650 .

1. Copy-paste [this test content](https://cloudup.com/cgewpNb7COD) into a page:
2. Cycle through the different style packs, and note the spacing differences between the headers with the `.accent-header` class, and the section header on the article blocks:

**Default:**

image

**Style 1:**

image

**Style 2:**

image

**Style 3:**

image

**Style 4:**

image

**Style 5:**

image

4. Confirm that the vertical spacing is more consistent with the two header types:

**Default:**

image

**Style 1:**

image

**Style 2:**

image

**Style 3:**

image

**Style 4:**

image

**Style 5:**

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -183,7 +183,9 @@
 		margin-bottom: 32px;
 		margin-top: 32px;
 
-		&:first-child {
+		&:first-child,
+		&.wpnbha:first-child .article-section-title,
+		&.accent-header + div.wpnbha {
 			margin-top: 0;
 		}
 
--- a/newspack-theme/sass/styles/style-1/style-1.scss
+++ b/newspack-theme/sass/styles/style-1/style-1.scss
@@ -20,6 +20,12 @@
 	}
 }
 
+.accent-header,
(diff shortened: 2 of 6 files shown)


laurelfulford:
Thanks @claudiulodro!

> Just as a heads-up/question do you want the same styles applied to sidebar widget titles? 

In general, they're kind of a variation, but not identical :) I found when building the theme that making them the same got to be a bit 'much' with some of the style packs (like with 1 and 3, where there's a little visual accent next to the element). So they do share the class, which is confusing, but there's some widget specific styles that override some of the styles.

Just let me know if you have any questions about this!

Issue #683, opened 2020-01-09
Add styles for the Organic Profile block


laurelfulford:
There's a PR to recommend the Organic Profile block here: https://github.com/Automattic/newspack-plugin/pull/374

The theme should include some styles to make sure it fits in with the overall look and feel of the different styles. Some possibilities include:
- adjusting font sizes
- removing the drop shadow
- adjusting the image size

laurelfulford:
Resolved with #684.

PR #684, opened 2020-01-09
Organic Profile Block: Overwrite some of the default styles to make the block more generic to Newspack

thomasguillot:
### All Submissions:

This PR remove some of the custom styles and makes the block a bit more generic:

* Adjusted the margin/padding
* Removed the box-shadow
* Social Icons to follow a 8px grid
* The image was being displayed via a background-image with a `background-size: cover` property on larger screens. I switched this to using `` all the time.

See #683 and https://github.com/Automattic/newspack-plugin/pull/374

1. Add the Organic Profile Block to your site.
2. Create a page/post with that block.
3. Check the front-end and back-end

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -855,7 +855,7 @@
 		}
 	}
 
-	// Jetpack Blocks
+	//! Jetpack Blocks
 	.jp-relatedposts-i2 {
 		border-top: 1px solid $color__border;
 		border-bottom: 1px solid $color__border;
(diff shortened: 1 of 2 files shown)


thomasguillot:
> Do you think we should do some sort of standardizing for the image sizes? e.g. if you're making a Staff page or something, you need to make sure all the images are the same size otherwise it looks something like this
>
> I think this is probably why the background was `cover` in the original styling. Not a big deal either way, just throwing it out there. :) This should be good to merge either way.

That's a good point and you're probably right about the why.

I don't mind switching back to background `cover`. I wonder what @laurelfulford thinks about this 😊

thomasguillot:
@laurelfulford 08dd70e should fix it

PR #685, opened 2020-01-09
Make sure column block border styles are being applied correctly

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR goes with a related Blocks PR, https://github.com/Automattic/newspack-blocks/pull/325, and should be tested with it. 

### How to test the changes in this Pull Request:

Instructions to test both PRs together can be found at https://github.com/Automattic/newspack-blocks/pull/325

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #686, opened 2020-01-09
Update editor classes to match what's currently in Gutenberg

laurelfulford:
### All Submissions:

As of Gutenberg 7.2, the legacy classes `.editor-block-list__layout` and `.editor-block-list__block` [were removed from the editor](https://github.com/WordPress/gutenberg/pull/19050).

These are currently used in the editor for the custom colours and typography; this PR updates them to their replacement classes `.block-editor-block-list__layout` and `.block-editor-block-list__block`

1. Verify that you're running Gutenberg 7.2.
2. Under the Customizer, apply custom colours and fonts.
3. View a few posts/pages in the editor; note that the colours/fonts are not being applied.
4. Apply the PR.
5. Confirm that your custom colours/fonts are now being applied. 

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -380,75 +380,75 @@ function newspack_custom_colors_css() {
 		 * - buttons
 		 */
 
-		.entry-meta .byline a {
+		.block-editor-block-list__layout .block-editor-block-list__block .entry-meta .byline a {
 			color: ' . newspack_color_with_contrast( $primary_color ) . '; /* base: #0073a8; */
 		}
 
-		.editor-block-list__layout .editor-block-list__block .wp-block-quote:not(.is-large):not(.is-style-large),
-		.editor-styles-wrapper .editor-block-list__layout .wp-block-freeform blockquote {
+		.block-editor-block-list__layout .block-editor-block-list__block .wp-block-quote:not(.is-large):not(.is-style-large),
+		.editor-styles-wrapper .block-editor-block-list__layout .wp-block-freeform blockquote {
 			border-color: ' . $primary_color . '; /* base: #0073a8; */
 		}
 
-		.editor-block-list__layout .editor-block-list__block .wp-block-pullquote.is-style-solid-color:not(.has-background-color),
-		.editor-block-list__layout .editor-block-list__block .wp-block-newspack-blocks-donate.tiered .wp-block-newspack-blocks-donate__tiers input[type="radio"]:checked + .tier-select-label {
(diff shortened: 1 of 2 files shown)


jeffersonrabb:
It looks like this is about replacing all instances of `.editor-block-list__layout` with `.block-editor-block-list__layout` and all instances of `.editor-block-list__block`with `.block-editor-block-list__block`. Based on this assumption, I searched the codebase and found a few occurrences of `editor-block-list__layout` and `.editor-block-list__block`, all in the editor stylesheets for various stylepacks. Is this anything to worry about? (Maybe not, given that we're working on sunsetting these stylesheets anyway?)

newspack-theme/inc/color-patterns.php
```diff
@@ -380,93 +380,93 @@ function newspack_custom_colors_css() {
 		 * - buttons
 		 */
 
-		.entry-meta .byline a {
```
jeffersonrabb: This is the only change that doesn't appear to be a replacing `editor-block-list__layout .editor-block-list__block`. Is it correct?

laurelfulford: There's that, and here: https://github.com/Automattic/newspack-theme/pull/686/files/a7eca439d09f988db35e07de3fceaeba36c4f66f#diff-38fb28a9c52ba7b44eaa303759a7aaadL363

Both added the `.block-editor-block-list__layout .block-editor-block-list__block` classes where there weren't any before, for consistency; for the second change, there was also a missing `,` after the `.cat-links` selector, causing some of the `text-transform: uppercase` styles not to work (something I found when testing this). Apologies, I should've called that out more clearly!

laurelfulford:
> Based on this assumption, I searched the codebase and found a few occurrences of editor-block-list__layout and .editor-block-list__block, all in the editor stylesheets for various stylepacks. Is this anything to worry about? (Maybe not, given that we're working on sunsetting these stylesheets anyway?)

Gah, good catch! Those should be replaced, too -- I was focused on the colours/typography because they were the parts I noticed that had stopped working, I should have checked the other stylesheets, too. Fresh commit to fix that is incoming!

laurelfulford:
Apparently I can't read today 🤦‍♀ You're right, those styles won't be needed anymore as they are now:

#685 is a theme PR that makes tweaks to the columns block, to go with changes being made to that markup. As part of it, I was able to simplify the editor styles, and remove the `.editor-block-list__layout` and `.editor-block-list__block` classes entirely. I think it probably makes sense to leave them as is with this PR, and let that one take care of those files, to avoid conflicts. 

Just let me know if you think that's a bad idea, or if you have any questions about this!

laurelfulford:
Thanks Jeff! 🙌

PR #689, opened 2020-01-11
Group Block: Sit right under header on home page

### LINKED ISSUE #688: Homepage: Remove space if first block is a group block
If the first block is a group block on a homepage with a background colour, it would be cool to remove the space between it and the header, for a more seamless look for some layouts.

A similar thing could be looked into for a group block at the very end.

laurelfulford:
### All Submissions:

This PR makes it so on the static front page, if you hide the page title, if you add a full-width group block as the first thing on the page, it will sit right below the header. This is to help recreate similar layouts to some of our cohort's live sites. 

When using Style 2, this group block will cover the 'overlap' style, similar to how the 'behind'/'beside' featured image styles are handled. I'm open to approaching this in another way, if anyone thinks this could be handled better!

Lastly, this PR also removes the `overflow: hidden` from the `.site-content` class, to allow the group block to be bumped up. This style was in place to make sure that full-width blocks didn't cause side scrolling. It's also on the `#page` element, though, and having it there alone prevents this issue, so it isn't needed on `.site-content`. 

Closes #688.

1. On your test site, if it's not already set up this way, set a static page as the homepage and hide the title using the controls under Customizer > Homepage Settings.
2. Edit the homepage, and add a Group block as the first thing. Make it full-width, and give it a background colour so it's easier to see.
3. Save and publish; view on the front page -- note the gap:

image

4. Apply the PR and run `npm run build`
5. View the homepage again; confirm the group block sits immediately under the header, with no gap; resize the window and confirm it stays that way:

image

image

6. Navigate to Customizer > Header Settings, and test group styles with the short header setting:

image

7. Navigate to Customizer > Style Packs, and switch to Style 2.
8. Confirm that on desktop-sized screens, the group block covers the 'overlap' style, including the wide border at the top of the content. On mobile-sized screens, the group block should sit immediately below the 2px border:

image

image

9. Edit the homepage again; test a narrower group block at the top, and confirm the gap returns:

image

10. Try other blocks at the top of the homepage, and confirm they sit the standard distance from the header:

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -976,6 +976,13 @@
 	}
 }
 
+.newspack-front-page.hide-homepage-title .entry-content > .wp-block-group.alignfull:first-child {
+	margin-top: 0;
+	@include media( tablet ) {
+		margin-top: calc( #{ -0.5 * $size__spacing-unit } - 1px ); // minus 1px to offset bottom border on header
+	}
+}
+
 .entry-content .has-text-color a,
 .entry-content .has-text-color a:hover,
 .entry-content .has-text-color a:visited {
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -21,7 +21,6 @@
 
 .site-content {
 	min-height: 30vh;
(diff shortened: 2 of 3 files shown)


laurelfulford:
Thanks! 😄

PR #692, opened 2020-01-13
Default Style: Make 'Tertiary menu' styles more consistent across header settings

laurelfulford:
### All Submissions:

Right now if you change the Header Settings (eg. setting a solid background) while using the Default Style pack, the tertiary menu's appearance changes, and it's a bit confusing: You lose the button-like styles, and font weights -- for example:

image

This PR updates these styles so they're consistent across all header styles. It also makes the menu item using the `menu-highlight` class use the secondary colour when using the solid-background option. 

(This second bit may make sense to roll out to the other style packs; would be interested in feedback on that option, as it won't affect live sites). 

Edited to add: I've update the PR to take mobile styles into account. The button styles aren't applied exactly the same there, to try to keep things simple: the 'default' style in the mobile menu is a plain link, and links with the class `menu-highlight` have a darker version of the primary colour (which is always used as the background of the mobile menu). It's not as prominent as it appears on the desktop version, but seems less likely to be very clash-y, since it would otherwise always be the secondary colour against the primary colour, which may not sit side by side nicely.

2. Navigate to Customizer > Style Packs, and pick the default style pack.
3. Navigate to Customizer > Menus, and add a menu to the 'tertiary' location; add at least two menu items and give on the class `menu-highlight`.
4. View on the front-end, and confirm you have something like:

**Default colours:**

image

image

**Custom colours:**

image

image

5. Navigate to Customizer > Header Settings and switch to 'Solid Background'. The 'default' tertiary menu links should still have a slight background, a darker shade of the header; the menu item with the `menu-highlight` class should use your secondary colour:

**Default colours:**

image

**Custom colours:**

image

image

image

image

6. Navigate to Customizer > Header Settings; uncheck 'solid background' and check 'short header' (another case where the button style was not being applied):

**Default colours:**

image

**Custom colours:**

image

image

7. Navigate back to Customizer > Header Settings, and also check 'Solid Background' again; you should have something like:

**Default colours:** 

image

**Custom colours:**

image

image

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -199,10 +199,17 @@ function newspack_custom_colors_css() {
 
 	if ( newspack_is_active_style_pack( 'default' ) ) {
 		$theme_css .= '
+			.nav3 a {
+				background-color: ' . newspack_adjust_brightness( $primary_color, -20 ) . ';
+				color: ' . $primary_color_contrast . ';
+			}
+			.h-sb .site-header .nav3 .menu-highlight a {
+				background-color: ' . $secondary_color . ';
+				color: ' . $secondary_color_contrast . ';
+			}
 			.cat-links a,
 			.cat-links a:visited,
-			/* Header default background; default height */
-			body.h-db.h-dh .site-header .nav3 .menu-highlight a {
+			.site-header .nav3 .menu-highlight a {
 				background-color: ' . $primary_color . ';
 				color: ' . $primary_color_contrast . ';
 			}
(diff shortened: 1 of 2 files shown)


laurelfulford:
I just realized I missed double-checking the mobile styles for this one -- pulling the 'Needs Review' tag for now, will circle back to fix and re-add once it's ready for a set of eyes!

laurelfulford:
I've updated the PR to take the mobile menu into account.

In an attempt to keep things simple, for the mobile menu I left the regular link as plain, and the `menu-highlight` link as the one with the darker primary colour.

The mobile menu always uses the primary colour as a background, which may clash with the secondary colour, so I figured this would be the safest route. I'd appreciate a second opinion about that, though, and a second review to quickly confirm that I haven't caused more problems with this update -- thanks! :)

laurelfulford:
Thanks Phil! 🙌

PR #698, opened 2020-01-15
Remove escaping from custom colour and typography CSS

laurelfulford:
### All Submissions:

I noticed in working a the 'Style 2' child theme PR that the `esc_html()` function used to escape the Custom Colour CSS in the child themes was turning `>` to `&gt;`, breaking some of the CSS selectors. `wp_kses()`, used by the Typography CSS, is doing the same thing.

Since the Colour and Typography CSS is not user-generated, but is hard-coded strings passed from a different file, they don't actually need escaping. So I've removed the functions, and marked the functions as being 'safe'. 

**Edited to add:** This PR has been updated to escape custom colour and typography variables in the color-patterns and typography files, and to remove the unused colour filter code (more details in [this comment](https://github.com/Automattic/newspack-theme/pull/698#issuecomment-575227414)).

1. Apply the PR.
2. Try switching the custom colours and type; ensure you don't see any errors.
3. Try switching the theme to 'Newspack Sacha', and repeat step 2.
4. Try switching the theme to 'Newspack Scott', and repeat step 2.
5. Review the changes in the color-patterns.php and child-color-patterns.php files; ensure each outputted colour variable is escaped with `esc_html()`, and that there are no typos.
6. Review the changes in the typography.php and child-typography.php files; ensure each outputted typography variable is escaped with `wp_kses()`, with the allowed tags value of `null`.
7. Check the editor and make sure the colours populating the palette match your current colour settings.

### Other information:


philipjohn:
Because of the way the CSS is constructed here, and because it can be filtered, we definitely need to adequately escape this output.

I'd recommend keeping this change but escaping the individual variables within the `newspack_sacha_custom_colors_css()` function to ensure that dynamic output is safe. None of that dynamic output should suffer from the problem being addressed here by the looks of it.

That does leave the problem of the `newspack_sacha_custom_colors_css` filter, because if anything did manage to execute code, they could use that filter to alter the output of the entire website, even though we've escaped the dynamic elements of the CSS. Unless we're already using that filter somewhere, I'd suggest removing it entirely unless we absolutely need it - otherwise it's a potential security hole that we can do without.

A good finishing touch would be to document clearly on the `newspack_sacha_custom_colors_css()` function that it's output is expected to be safe.

laurelfulford:
> Because of the way the CSS is constructed here, and because it can be filtered, we definitely need to adequately escape this output.

Giant 🤦‍♀ from me. Thanks Phil!

> Unless we're already using that filter somewhere, I'd suggest removing it entirely unless we absolutely need it - otherwise it's a potential security hole that we can do without.

This was a left-over from Twenty Nineteen that's only really needed if we want to make the default colours different in the child themes. Due to the complexity of, well, everything in the theme, I can see that causing more issues than it would solve (which is exactly why all of the style packs use Newspack blue). 

I'll work on untangling this, and update the PR. Thanks for catching my not-great assumptions, Phil!

laurelfulford:
I've reworked this PR pretty substantially to:

* Remove the unused color-filter code.
* Escape the output in all of the color-pattern.php and typography.php files. The former uses `esc_html()`, and the latter uses `wp_kses()`, passing `null` for allowed tags, to make sure the quotes needed for the font families aren't escaped. 
* Updated the colour CSS output -- there was still a `data-` attribute passing the current `$primary_color` that was used to immediately preview colour changes in Twenty Nineteen. This wasn't being used in the Newspack theme -- the complexity of the colours made it tricky to do, so colour changes are previewed with a refresh. 

I'll update the above testing steps to reflect these changes.

newspack-sacha/inc/child-typography.php
```diff
@@ -20,7 +20,7 @@ function newspack_sacha_custom_typography_css() {
 			.has-drop-cap:not(:focus)::first-letter,
 			.taxonomy-description,
 			.entry .entry-content blockquote, .entry .entry-content blockquote cite, .wp-block-pullquote cite {
-				font-family: $font_header;
+				font-family: ' . wp_kses( $font_header, null ) . ';
```
philipjohn: That's weird - I'd have expected passing `null` here to error out, because the docs say it's a required parameter.

laurelfulford:
Thanks Jeff!

PR #699, opened 2020-01-15
Copy Theme Settings When Switching

jeffersonrabb:
### All Submissions:

One problematic aspect of the switch from Style Packs to Child Themes is that most theme settings are lost when switching between two themes in the Newspack family. Most theme settings would be expected to persist, such as colors, typography, Author Bio settings, and Featured Image Settings. This branch is an attempt to solve this by copying over all theme mods when switching from one Newspack theme to another. Whenever a theme switch occurs, if the theme slug (or stylesheet) begins with `newspack-` then the settings will be duplicated from old to new.

1. On `master`, Install the Newspack Theme, and set up a complicated configuration in the customizer. 
2. Switch to Newspack Sacha. Verify the theme settings are  lost. 
3. Switch back to Newspack Theme and you should see all the settings from before.
4. Switch to this branch, then switch themes again to Newspack Sacha. This time all the settings should  carry over. 
5. Verify that switching to non-Newspack themes does not cause anything strange to happen. 

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -539,6 +539,27 @@ function newspack_register_meta() {
 }
 add_action( 'init', 'newspack_register_meta' );
 
+
+/**
+ * Migrate theme settings when switching within the family of Newspack themes.
+ *
+ * @since Newspack Theme 1.0.0
+ */
+function newspack_migrate_settings( $old_name, $old_theme = false ) {
+	$theme           = wp_get_theme();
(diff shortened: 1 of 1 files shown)


jeffersonrabb:
> The only "issue" I noticed was kind of an experience one: if you pick 'Live Preview' when switching themes, or switch themes from inside of the Customizer, the preview does not show the settings that are transferred. This makes sense, because after_switch_theme hasn't fired yet.

I tried messing with the `start_previewing_theme` action and it doesn't seem to fire when previewing a theme. I'm going to take your advice and merge this, then look into improvements to previewing separately. Thanks for looking!

PR #760, opened 2020-02-05
Fixes author avatar sizing

### LINKED ISSUE #752: Incorrect Avatar Crops On Rivard Report
**Describe the bug**
On Rivard Report, some author avatars are incorrectly cropped. At first glance, it appears this occurs for any avatars that are not perfect squares.

- I initially suspected this was a thumbnail regeneration issue, so I ran `wp media regenerate` on the avatar. No change. Also downloaded the original and re-uploaded. No change.
- The problem is not limited to multi-author stories. Here's a single-author story with the same: https://therivardreport.com/what-it-will-take-to-convince-san-antonio-voters/
- Tried using the same Avatar on a local instance and here it worked fine. On local, Jetpack Image Accelerator is not activated, which may be a clue.

**Screenshots**
Live example: https://therivardreport.com/reports-detail-energy-waste-in-northside-isd-citys-largest-school-district/

Screen Shot 2020-02-02 at 3 48 21 PM

**Expected behavior**
All avatars are perfect un-cropped circles.

laurelfulford:
### All Submissions:

This PR fixes an avatar sizing issue that happens when you combine WP User Avatar, Jetpack's  site accelerations, and non-square avatars. This combination either causes edges of the image to show (AMP enabled), or cause the image to be stretched to fit (AMP disabled). 

As part of this PR I also increased the size of the images loaded, to improve their appearance when they were taller/wider, as well as corrected a directory name issue in Newspack Sacha, so that child theme's author-bio.php would actually be loaded. 

Note: I ended up using `preg_replace()` to add the `object-fit="cover"` to the coauthors plus avatars, as there isn't a proper way to add attributes like with `get_avatar()`, just classes. If there' as better way, please let me know!

Also, apologies in advance, this one is a bit of a pain to review!

Closes #752.

1. Start with a test site where you can enable Jetpack site acceleration. 
2. Under AMP settings, switch to 'transitional' for easier testing.
3. Using WP User Avatar, add a tall, skinny image as a user's profile image.
4. Using WP User Avatar, add a wide, short image as a user's profile image.
5. Assign those authors to a couple posts, include both to one post together as Co-Authors. 
6. View without AMP enabled - note how the avatars look -- the tall image will be a bit vertically smushed, and the wide image, a bit horizontally smushed:

image

7. Enable AMP - note how you can now see the avatar edges:

image

8. Apply the PR and run `npm run build`.
9. Check the non-AMP version again and confirm the images look correct: 

image

10. Check again with AMP enabled: 

image

11. Turn off Co-Authors plus, and check the single article byline with and without AMP enabled.
12. Switch focus to the author bio at the bottom of the post. Check it with Co-Authors Plus enabled, with and without AMP, and with Co-Authors Plus disabled, with and without AMP.
13. Switch to 'Newspack Sacha' and re-test the above again (the markup is a bit different for this child theme). 
14. Navigate to an author archive. Check it with Co-Authors Plus enabled, with and without AMP, and with Co-Authors Plus disabled, with and without AMP.
15. Check the comments and confirm they still have rounded avatars.
16. Switch to Newspack Scott, and confirm that avatars are square (not rounded).
17. Double-check that the regular avatars display correctly, with and without Co-Authors Plus; with and without WP User Avatar enabled, and with and without AMP. 

### Other information:

### DIFF
--- a/newspack-sacha/template-parts/post/author-bio.php
+++ b/newspack-sacha/template-parts/post/author-bio.php
@@ -19,20 +19,21 @@ if ( function_exists( 'coauthors_posts_links' ) && is_single() ) :
 
 			if ( 'guest-author' === get_post_type( $author->ID ) ) {
 				if ( get_post_thumbnail_id( $author->ID ) ) {
-					$author_avatar = coauthors_get_avatar( $author, 80 );
+					$author_avatar = coauthors_get_avatar( $author, 160 );
 				} else {
 					// If there is no avatar, force it to return the current fallback image.
 					$author_avatar = get_avatar( ' ' );
 				}
 			} else {
-				$author_avatar = coauthors_get_avatar( $author, 80 );
+				$author_avatar = coauthors_get_avatar( $author, 160 );
 			}
+			$author_avatar = preg_replace('/(<img\b[^><]*)>/i', '$1 object-fit="cover">', $author_avatar);
 			?>
 
 			<div class="author-bio">
 				<div class="author-bio-text">
 					<div class="author-bio-header">
-						<?php echo wp_kses( $author_avatar, newspack_sanitize_avatars() ); ?>
+						<?php echo '<span class="author-avatar">' . wp_kses( $author_avatar, newspack_sanitize_avatars() ) . '</span>'; ?>
 
 						<div>
 							<h2 class="accent-header">
@@ -69,10 +70,10 @@ elseif ( (bool) get_the_author_meta( 'description' ) && is_single() ) :
 	<div class="author-bio-text">
 		<div class="author-bio-header">
 			<?php
-				$author_avatar = get_avatar( get_the_author_meta( 'ID' ), 80 );
+				$author_avatar = get_avatar( get_the_author_meta( 'ID' ), 160, '', '', array( 'object-fit' => 'cover' ) );
 				if ( $author_avatar ) {
 					// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-					echo $author_avatar;
+					echo '<span class="author-avatar">' . $author_avatar . '</span>';
 				}
 			?>
 
--- a/newspack-scott/sass/style.scss
+++ b/newspack-scott/sass/style.scss
@@ -100,7 +100,7 @@ Nicolas Gallagher and Jonathan Neal https://necolas.github.io/normalize.css/
 }
 
 /* Avatar */
-.avatar,
+.author-avatar,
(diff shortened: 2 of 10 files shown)


jeffersonrabb:
Here's an possible alternate approach: 

It seems the real problem here is the resizing Jetpack Image Accelerator (Photon) applies. Jetpack has a filter to allow plugins/themes to prevent downsizing, https://developer.jetpack.com/hooks/jetpack_photon_override_image_downsize/. If we can successfully use this filter to stop Photon from downsizing thumbnails, we _might_ be able to solve the problem in a much simpler and more universal way. We could avoid all of the Co-Authors logic, all of the AMP logic, and the fix should work for the block as well as the theme without changes to the blocks codebase. The filter only gives us the image's ID and in some cases dimensions. I tried an approach where we override Photon downsizing for any image that has equal dimensions that are less than 100px, which I _think_ should capture avatars well.

On `master` try adding this to `functions.php` and see if it works?

```
function newspack_jetpack_photon_override_image_downsize( $default, $args ) {
	if ( is_array( $args['size'] ) ) {
		list( $width, $height ) = $args['size'];
		if ( $width === $height && $width < 100 & $width > 0 ) {
			return true;
		}
	}
	return $default;
}

add_filter( 'jetpack_photon_override_image_downsize', 'newspack_jetpack_photon_override_image_downsize', 10, 2 );
```

laurelfulford:
JEFF! This is SO much better!

I had to bump the image size up to `120` to get it working on the archive page, but in my testing it looks like it's correctly treating the avatars in all cases in both the theme and block that I can see. 

As I understand it, this will affect all images sized 120px wide/high or smaller, right? This doesn't seem like it would be problematic -- the theme doesn't set any other images that size -- I'm just trying to make sure I understand. 

Let me know if you'd like me to go ahead and spin up a fresh PR with this approach, instead!

jeffersonrabb:
> As I understand it, this will affect all images sized 120px wide/high or smaller, right? This doesn't seem like it would be problematic -- the theme doesn't set any other images that size -- I'm just trying to make sure I understand.

Yes, this would apply to any image meeting the criteria — there doesn't seem to be any additional data we can use to ensure we are applying this only to avatars. My main concern is inadvertently affecting non-avatars, although it may be that any small image with a square aspect ratio actually shouldn't be downsized anyway—i.e. if it's not an avatar it's probably something avatar-like. Anyway, I'll open a new PR and we can see if anyone can find any flaws in the approach?

laurelfulford:
Closing this in favour of #764!

PR #333, opened 2019-08-27
Cover Block: Fix section header contrast issue with homepage block

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR fixes a contrast issue that happens when you nest the homepage block into the cover block -- the section header of the homepage block isn't white on the front end for Style Packs 1-4. 

Thankfully the fix is just making the selectors less specific. 

Style 1:

image

Style 2:

image

Style 3:

image

Style 4:

image

### How to test the changes in this Pull Request:

1. Copy this [test content](https://cloudup.com/cmGhyEkvMGg) into a post.
2. Apply the PR and run `npm run build.`
3. Cycle through the style packs and confirm the 'section header' text is white on the front-end (it's already working as expected in the editor). For Style packs 1 and 3, the accents (the blocks before or above) should still use your primary colour.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #793, opened 2020-02-21
Add separate custom colour control for the Header Background

### LINKED ISSUE #645: Header background colour as separate option
Right now if you pick the 'solid background' option for the header, it uses the primary colour, when you're using style packs default, 1 and 2. 

Styles 3, 4 and 5 use dark grey/black.

To reduce confusion/make the options more flexible, it might be good to add a separate header bg colour option.

laurelfulford:
### All Submissions:

This PR adds a separate option to change the header background colour independent of the child theme's default -- right now, the default theme, Scott and Nelson will us the primary colour, and Katharine, Sacha and Joseph use grey or black. 

Unfortunately, the control is a bit convoluted, in an attempt to not impact existing sites. I'm hoping once it's rolled out, and settings updated on each site, we can strip out some of the child theme specific code (like the grey/black defaults -- payback for my own 'cleverness' there), and any instance of a theme still using the primary colour. This may not be the best approach, and I'm definitely open to chatting with anyone about it, and making as many improvements as possible before it goes out there. 

It's possible that this should also impact how the colours work for the 'featured image beside' style after it's merged -- right now they use the primary colour, but if the header also uses the primary colour as a background colour, they'll switch to dark grey. This is more confusing cleverness, and should be switched to something more straight-forward -- a light-grey/dark grey/primary colour toggle as an option, or yet another colour picker, since there will be no simple way to avoid clashes with any-possible-header-colour. 

Note: this PR needs #698 to land first, and after that will need to be updated to remove any merge conflicts. That issue fixes the CSS escaping that's currently causing selectors to get messed up, and causing contrast issues in the header in some cases -- without it, you could see issues in this PR that are actually fixed there. 

As expected, this one will be a bit hellacious to test -- apologies in advance!

Closes #645.

First, test the behaviour of the new options in the Customizer:

2. Switch to the default theme, default style pack, default header settings, default colour settings. 
3. View the 'Colors' panel:

image

4. Switch to 'Custom' and note the typical appearance:

image

5. Switch to the 'Header Settings' panel and toggle on Solid Header'.
6. Switch back to the 'Colors' panel, and you should see:

image

The 'default' setting will keep the theme's default behaviour -- either using the primary colour, or a grey/black. 

7. Toggle this option to 'Custom' and confirm you have a colour picker for the header: 

image

8. To make sure these panels show/hide as expected, try changing the settings on the Colors panel at random, and go back and forth to the Header Settings to toggle 'Solid Background' on and off. The purpose is to see if you can catch something showing when it shouldn't, eg. just the Header Background Color picker when the default colour palette is picked, or anything to do with the header colour when 'Solid Background' is unchecked. 

Next, check that the options actually work:

1. Return the settings to Solid Background for the header, and pick the 'Custom' Header background colour option. It should default to grey:

image

2. Try a couple colour options to make sure the contrast is working (one dark, one light):

image

image

3. Try the same with the short header option selected:

image

image

4. Test the mobile menu:

image

5. Silently curse Laurel for getting you to review this PR as you look at step 6.
6. Repeat steps 1-3 with the child themes (no need to test step 4, as the child themes don't do anything unique to its colours). There are screenshots below for what each should look like:

_(The contrast issues in these screenshots is fixed by #698; I can update screenshots once that's merged)._

**Scott:**

Default (primary colour):

image

Custom light:

image

image

Custom dark:

image

image

**Nelson:**

Default (primary colour):

image

Custom light: 

image

image

Custom dark:

image

image

**Katharine:**

Default (dark grey - #333):

image

Custom light:

image

image

Custom dark:

image

image

**Sacha:**

Default (dark grey - #333):

image

Custom light:

image

image

Custom dark:

image

image

**Joseph:**

Default (black):

image

Custom light:

image

image

Custom dark:

image

image

I did make the same updates in the style packs code, but I think outside of making sure it's not throwing errors, I don't think it needs the same level of testing -- if we do have reports of issues outside of the cohort, it's a good opportunity to get someone switched to a child theme :) 

### Other information:

### DIFF
--- a/newspack-joseph/inc/child-color-patterns.php
+++ b/newspack-joseph/inc/child-color-patterns.php
@@ -37,11 +37,11 @@ function newspack_joseph_custom_colors_css() {
 		$theme_css .= '
 			/* Header solid background */
 			.h-sb .middle-header-contain {
-				background-color: ' . $header_color . ';
+				background-color: ' . esc_html( $header_color ) . ';
 			}
 			.h-sb .top-header-contain {
-				background-color: ' . newspack_adjust_brightness( $header_color, -10 ) . ';
-				border-bottom-color: ' . newspack_adjust_brightness( $header_color, -15 ) . ';
+				background-color: ' . esc_html( newspack_adjust_brightness( $header_color, -10 ) ) . ';
+				border-bottom-color: ' . esc_html( newspack_adjust_brightness( $header_color, -15 ) ) . ';
 			}
 
 			/* Header solid background */
@@ -56,7 +56,7 @@ function newspack_joseph_custom_colors_css() {
 			.h-sb.h-sh .nav1 ul.main-menu > li > a:hover,
 			.h-sb .top-header-contain,
 			.h-sb .middle-header-contain {
-				color: ' . $header_color_contrast . ';
+				color: ' . esc_html( $header_color_contrast ) . ';
 			}
 		';
 	}
--- a/newspack-joseph/inc/child-color-patterns.php
+++ b/newspack-joseph/inc/child-color-patterns.php
@@ -14,8 +14,14 @@ function newspack_joseph_custom_colors_css() {
 	if ( 'default' !== get_theme_mod( 'theme_colors', 'default' ) ) {
 		$primary_color   = get_theme_mod( 'primary_color_hex', $primary_color );
 		$secondary_color = get_theme_mod( 'secondary_color_hex', $secondary_color );
+
+		if ( 'default' !== get_theme_mod( 'header_color', 'default' ) ) {
+			$header_color          = get_theme_mod( 'header_color_hex', '#666' );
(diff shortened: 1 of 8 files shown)


claudiulodro:
It's cool if you want to address that above issue in a follow-up PR, not sure how complex the logic will be to make things work nicely.

laurelfulford:
Thanks @claudiulodro! 

I think that should be pretty straight-forward to fix -- let me give it a quick try. Worst case, if it goes out without it, the 'default' colour setting is not something most sites have on for any length of time.

laurelfulford:
I think I squashed that incorrect customizer control issue in 02e6c63 -- @claudiulodro do you mind taking a second look to confirm before I merge? Thanks!

claudiulodro:
Looks good. Thanks for the speedy fix. 👍

laurelfulford:
Thanks for catching that! 🙌

PR #805, opened 2020-02-27
Refine the German translation of "Join the conversation"

philipjohn:
Our only German publisher mentioned the translation was not quite right here.

> „Nehmen Sie an der Konversation teil” should be „Reden Sie mit!”

So this PR makes that change.


laurelfulford:
This option looks like it's already built into the above plugin -- the second to last field ("Heading (no comments)" replaces the Join the Conversation text on the front-end, even though the latter does appear when you already have comments. 

@philipjohn if this seems like an okay approach in this case, let me know and I can update the site with this plugin to fix!

philipjohn:
Yeah, looks like this is more than just a translation. I'm going to go back and re-assess.

PR #829, opened 2020-03-13
Homepage Posts: Remove top margin on accent-header with AMP

### LINKED ISSUE #791: Homepage Posts + Columns: 'Load More' container affects header styles
Right now in the theme, there are styles in place to remove the margin from the first block in a column, and the margin from the `h2` used for the section header when the homepage posts block is the first block in a column.

However, the selectors are pretty specific, targeting the first `h2` in the block. If you have 'load more posts' enabled and AMP, another container is added, making this selector no longer work.

These styles need to be updated to make sure the header sits right at the top, regardless of the block settings.

laurelfulford:
### All Submissions:

Right now the theme adds margins to the top and bottom of every block, but does remove them in a few cases -- one example is when they're at the top of a column. 

The homepage posts also have the top margin removed from the section title when populated; this is done by checking for the article block being the first child. 

When AMP and Load More are both enabled, an extra container is added to the block, screwing up these styles; this PR fixes that. 

Closes #791

1. Set up a column block; in each column, add a homepage article block with 'Load More' turned on. Add a section title to each block. Save and publish.
2. Navigate to Dashboard > AMP, and set AMP mode to transitional.
3. Open your saved page in two tabs; one in AMP mode and one without. Compare the top spacing and note that the one with AMP enabled has a top margin:

**Without AMP:**

image

image

**With AMP:**

image

image

5. Confirm that the AMP version no longer has the top margin on the section header:

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -172,7 +172,7 @@
 		margin-top: 32px;
 
 		&:first-child,
-		&.wpnbha:first-child .article-section-title,
+		&:first-child .article-section-title, // when Load More is enabled.
 		&.accent-header + div.wpnbha {
 			margin-top: 0;
 		}


laurelfulford:
Thanks Jeff!

PR #833, opened 2020-03-18
Prevent id selector format error in pre-commit checks

laurelfulford:
### All Submissions:

The SCSS formatting checks are not liking the underscore in the `#wp-block-jetpack-mailchimp_consent-text` ID selector, so this PR updates the selector to use an attribute selector instead with part of the ID.

1. Add a Mailchimp block to a test site, and note the text size in the paragraph disclaimer:

image

3. Refresh the block; confirm the paragraph font size has not changed.

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1125,7 +1125,7 @@
 		width: 100%;
 	}
 
-	#wp-block-jetpack-mailchimp_consent-text {
+	p[id^='wp-block-jetpack-mailchimp'] {
 		font-size: $font__size-sm;
 	}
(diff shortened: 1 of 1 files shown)


laurelfulford:
> Only question—the selector doesn't have any impact over the text in the editor, only front-end. Is this how it was intended?

Good question! We don't have the exact same selector in the editor styles, though there are some with a mix of dashes and underscores on classes (for the front-end, it was an ID). 

If those end up being an issue, we might need to change the prettier format rules -- a lot appear to be coming from the Gutenberg editor and Jetpack, and could be harder to dance around 😕 

I'm going to merge as is, and we'll tackle that when it comes up!

PR #855, opened 2020-03-26
Add support for Yoast Breadcrumbs

thomasguillot:
### All Submissions:

Since we require Yoast and it has a built-in breadcrumbs functionality, it'd make sense to add support for it.

I also noticed that quite a few news sites are actually using breadcrumbs.

Screenshot 2020-04-01 at 10 31 29

_Note: this would only be visible for tablet >= screens_

1. Enable breadcrumbs via: wp-admin > SEO > Search Appearance > Breadcrumbs
2. Navigate to a post
3. Notice the breadcrumbs below the header

### Other information:

### DIFF
--- a/newspack-nelson/inc/child-color-patterns.php
+++ b/newspack-nelson/inc/child-color-patterns.php
@@ -88,6 +88,9 @@ function newspack_nelson_custom_colors_css() {
 			.nav1 .sub-menu a,
 			.h-sb .site-header .highlight-menu .menu-label,
 			.h-sb .site-header .highlight-menu a,
+			.h-sb .site-breadcrumb,
+			.h-sb .site-breadcrumb a,
+			.h-sb .site-breadcrumb .breadcrumb_last,
 			.h-sb .site-footer {
 				color: ' . esc_html( $header_color_contrast ) . ';
 			}
--- a/newspack-nelson/inc/child-color-patterns.php
+++ b/newspack-nelson/inc/child-color-patterns.php
@@ -35,7 +35,9 @@ function newspack_nelson_custom_colors_css() {
 		/* Header short height; default background */
 		.h-sh.h-db .site-header,
 		.site-content #primary,
-		#page .site-header {
+		#page .site-header,
+		/* Yoast Breadcrumb */
+		.has-highlight-menu .site-breadcrumb .wrapper {
 			border-color: ' . esc_html( newspack_adjust_brightness( $primary_color, -40 ) ) . ';
 		}
 
--- a/newspack-nelson/inc/child-color-patterns.php
+++ b/newspack-nelson/inc/child-color-patterns.php
@@ -35,9 +35,7 @@ function newspack_nelson_custom_colors_css() {
 		/* Header short height; default background */
 		.h-sh.h-db .site-header,
 		.site-content #primary,
-		#page .site-header,
-		/* Yoast Breadcrumb */
-		.has-highlight-menu .site-breadcrumb .wrapper {
+		#page .site-header {
 			border-color: ' . esc_html( newspack_adjust_brightness( $primary_color, -40 ) ) . ';
 		}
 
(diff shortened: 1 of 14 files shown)


thomasguillot:
> In my testing, I found the breadcrumbs don't inherit the custom fonts

Fixed in https://github.com/Automattic/newspack-theme/pull/855/commits/40380d6adeb860dc481a8f94dccc2ca40a932449

I actually decided to avoid using the heading font to avoid confusion with menu items

thomasguillot:
> have some placement issues with Newspack Nelson

Fixed in https://github.com/Automattic/newspack-theme/pull/855/commits/4a5e18ed2515b665d02eb23dfbcbd384112a7e9f

The negative margin applied to the content was breaking the position of the breadcrumbs.

thomasguillot:
> When the Highlight menu is used, the proximity to each other in all but Newspack Nelson also seems a bit weird
>
> I'm not sure if we'd want to separate these with a line, not allow both to be used, or something else?

Fixed in https://github.com/Automattic/newspack-theme/pull/855/commits/d586b0e31b5ff26f33d502bab3de7424147139be

So displaying both is not ideal and maybe this should be documented. I added some css to make sure it looks ok though.

thomasguillot:
> Also, personally, I think the breadcrumb on posts -- the bulk of a new site's content -- is a bit redundant. It's not adding links to anything that isn't easily available already, and posts don't have a complex structure that seems to call for navigation to help figure out where you are (compared to, say, a site with a dense page structure). I might be missing an important use case here though!

Here are some thoughts:
 
* News sites with multiple sub categories can be quite difficult for visitors to know where they are without breadcrumbs (e.g. Sports > Football > England > Premier League)
* According to [a bunch of websites](https://duckduckgo.com/?q=breadcrumbs+seo), it helps with SEO
* And since it's an option, it's completely entirely up to the user to display it or not and at least we offer the option

thomasguillot:
> With Newspack Nelson, what do you think about switching the border to a slightly darker version of the current header background, instead of grey, since it sits on top of the header background?

Took care of it in https://github.com/Automattic/newspack-theme/pull/855/commits/fae04aaccd97effaf5da0326c687b5a2531435ac

> For Newspack Katharine, I find the doubled little boxes a bit odd (especially since you can end up with a lot of the primary colour with the Highlight Menu active). What do you think about switching this to the dotted line used elsewhere?

Good point, https://github.com/Automattic/newspack-theme/pull/855/commits/5d84c6ebc88a71c1c150d53a134aa422c2a74f86

> Newspack Scott has a similar issue (though it reads more clearly as perhaps an error?):

Yeah it was a typo: https://github.com/Automattic/newspack-theme/pull/855/commits/93bfcf35b17e8a8c2e3cf4f3d29c0d4a25c82884

> Lastly I think I preferred the header font for the breadcrumb trail, but get how using the body font helps differentiate it from the menu :) So I'd be good with whatever you prefer here.

I actually switched back to Headings in https://github.com/Automattic/newspack-theme/pull/855/commits/37449cfaa79f779a2fd68476b4519283b84a109b but I made sure the custom fonts are working 😊

newspack-nelson/inc/child-color-patterns.php
```diff
@@ -35,7 +35,9 @@ function newspack_nelson_custom_colors_css() {
 		/* Header short height; default background */
 		.h-sh.h-db .site-header,
 		.site-content #primary,
-		#page .site-header {
+		#page .site-header,
+		/* Yoast Breadcrumb */
+		.has-highlight-menu .site-breadcrumb .wrapper {
```
laurelfulford, line 40: Small nitpick -- the `.has-highlight-menu .site-breadcrumb .wrapper` style should use `$header_color` rather than `$primary_color` for the border -- `$header_color` will use the primary colour if the header is using the default background, but if not, you can end up with a bit of a mismatch. This is my site with a yellow primary colour but a reddish header colour: 

image

thomasguillot: Fixed in https://github.com/Automattic/newspack-theme/pull/855/commits/1458cd0392cc53a8f53022723a0a4db1b6c96477
Hopefully I didn't miss anything else!

thomasguillot:
> I have one small issue remaining that's new -- the border in Newspack Nelson always uses the primary colour, even if you have a separate custom header colour. I've suggested what to change in the code comment, but please let me know if you have any questions at all. Otherwise it's looking good!

Thank you, Laurel!

I took care of it in https://github.com/Automattic/newspack-theme/pull/855/commits/1458cd0392cc53a8f53022723a0a4db1b6c96477. Hopefully this time everything is fine!

laurelfulford:
Documentation has been added here: https://newspack.blog/support/themes/yoast-breadcrumbs/

PR #867, opened 2020-04-08
Checkout: Option to hide order details table

laurelfulford:
### All Submissions:

Right now, the 'Orders' table on the WooCommerce checkout screen is large, space-consuming, and makes little sense when used strictly for donations.

I'm not sure we want to remove this entirely, say, for sites that may use WooCommerce for a store as well, but we can at least add an option to hide it. This PR adds a Customizer option to show, hide, or toggle the Orders table on the checkout page. 

1. On a Newspack site with donations set up, add a recurring donation to your cart using the Donate block.
2. You should be taken to the Donation page automatically; note the large Order details table:

image

4. View the page again and note that the table is gone, replaced with a simple one liner describing your order: 

image

5. Navigate to Customizer > WooCommerce > Order Details, and confirm you have the following options:

image

6. Switch to 'Hide, with ability to toggle open', and click Publish. 
7. View the checkout page again; confirm that you have a 'Show Details' toggle:

image

8. Click the toggle and confirm it opens and closes the Order Details table, sans 'Order Details' header:

image

9. I believe this form will typically be used with AMP disabled on the page, so the form details update better, but this should be tested both with and without AMP enabled. So either enable or disable AMP (depending on your current setting), and repeat testing step 8.
10. Navigate back to Customizer > WooCommerce > Order Details, and pick "Show"
11. Confirm that the table now displays by default, with no toggle:

image
 
### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -344,10 +344,12 @@ function newspack_scripts() {
 		}
 
 		$newspack_l10n = array(
-			'open_search'       => esc_html__( 'Open Search', 'newspack' ),
-			'close_search'      => esc_html__( 'Close Search', 'newspack' ),
-			'expand_comments'   => esc_html__( 'Expand Comments', 'newspack' ),
-			'collapse_comments' => esc_html__( 'Collapse Comments', 'newspack' ),
+			'open_search'        => esc_html__( 'Open Search', 'newspack' ),
+			'close_search'       => esc_html__( 'Close Search', 'newspack' ),
+			'expand_comments'    => esc_html__( 'Expand Comments', 'newspack' ),
+			'collapse_comments'  => esc_html__( 'Collapse Comments', 'newspack' ),
+			'show_order_details' => esc_html__( 'Show details', 'newspack' ),
+			'hide_order_details' => esc_html__( 'Hide details', 'newspack' ),
 		);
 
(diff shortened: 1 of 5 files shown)


philipjohn:
I think this is actually really good, @laurelfulford ! It gets rid of the big ugly where it doesn't make sense but if a publisher does end up selling other products (e.g. event tickets) this would allow them to show those appropriately.

laurelfulford:
Made a minor update to improve the 'order summary' section of the page:

image

More than one item will be stacked with a small vertical space between them.

Marking this as ready for review!

laurelfulford:
Thanks for testing, Jeff!

What's kind of weird is in that second screenshot, the 'summary' version of the order is also not using the same styles. Does it have a `span` around everything in that line (no classes)?

It's like the CSS isn't working at all, but only for that one version. I can't recreate it yet, but still digging into it, might need to spin up a new site to see it!

jeffersonrabb:
> It's like the CSS isn't working at all, but only for that one version. I can't recreate it yet, but still digging into it, might need to spin up a new site to see it!

This appears to happen only when I'm logged in. If I test in an incognito browser, it works as expected. I'll ping you directly with a link to a site where you can see this.

laurelfulford:
Thanks Jeff!

PR #873, opened 2020-04-09
One Column template: Make max width a px value, not %

### LINKED ISSUE #870: One Column template: Doesn't need to be 65% wide
Right now the One Column template has the content set to 65%, which is the same as the default, two-column template.

This means the content starts to get narrow on smaller screens (like tablets), before it really needs to. It could actually have  max-width of 780px instead, to maximize space on mid-sized screens.

laurelfulford:
### All Submissions:

Right now, the one column template has a content area of 65% of the site's 1200px width, with makes it 780px wide. This matches how this is done for the default two column template.

This means that the content area of the one column template shrinks as you make the browser window narrower -- however, this doesn't make sense when you don't have a sidebar to make room for, and leads to weirdly narrow content at some tablet screen sizes. 

This PR changes this template's styles, so the content area has a max-width of 780px, meaning it will fill up more space on tablet-sized screens.

Closes #870 .

1. Apply the one-column template to a post and a page.
2. View on the front end; confirm that the content width is 780px wide when the browser window is large, but shrinks as you scale down the browser window.
4. Confirm that the post and page have not changed in appearance when the browser window is wide.
5. Try sizing down the browser window again; confirm that the content area remains at 780px wide, until the window becomes narrower than that, at which point, it fills the available space. 

### Other information:

### DIFF
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -64,8 +64,7 @@
 .blog #main,
 .search #main,
 .page .main-content,
-.single .main-content,
-.newspack-front-page.page-template-single-feature .site-main {
+.single .main-content {
 	@include media( tablet ) {
 		width: 65%;
 	}
@@ -94,6 +93,8 @@
 .newspack-front-page.page-template-single-feature .site-main {
 	margin-left: auto;
 	margin-right: auto;
+	max-width: 780px;
+	width: 100%;
 }
 
 .page-template-single-feature {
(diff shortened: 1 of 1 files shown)


laurelfulford:
I wouldn't say unexpected but maybe... sub-optimal? It just doesn't seem to make sense to shrink the one column content area when it's not making room for something else (like another column). 

I can hold off merging until it's confirmed that it won't introduce issues with ads!

jeffersonrabb:
> I can hold off merging until it's confirmed that it won't introduce issues with ads!

No need - go ahead and merge. If it does affect the ads code it will be a pretty subtle change that we can figure out later.

laurelfulford:
Ok - thanks Jeff!

PR #887, opened 2020-04-20
Make sure amp styles work with logo centering

laurelfulford:
### All Submissions:

There is a weird case where AMP will add the following CSS to a site, and it will cause the centred logo not to work when they have a width, height and sizes attribute:

```
[width][height][sizes]:not(.i-amphtml-layout-responsive) {
    display: block;
    position: relative;
}
```

This PR adds some 'backup' CSS in this cases, so rather than relying on the `img` to be `display: inline-block`, it will also work when it's display block.

1. Test on a site that's displaying this issue.
2. Confirm that the logo is not centring.
4. Confirm that it fixes the issue.
5. If possible, test on a site that is not displaying this issue on master and confirm that it doesn't cause any issues. 

### Other information:

### DIFF
--- a/newspack-theme/sass/site/header/_site-header.scss
+++ b/newspack-theme/sass/site/header/_site-header.scss
@@ -216,6 +216,7 @@
 
 			img {
 				height: auto;
+				margin: auto;
 				max-width: 100%;
 			}
 		}


jeffersonrabb:
Here are the conditions to replicate the issue:

- AMP transitional mode
- Center Logo theme mod
- Newpack Pop-ups installed 

Newspack Pop-ups enqueues various AMP scripts even if the request is non-AMP, which is probably the cause of the issue.

laurelfulford:
Thanks Jeff!

PR #890, opened 2020-04-21
Feat: Add options to WooCommerce thank you page

laurelfulford:
### All Submissions:

Based on some customer feedback, this PR adds options to simplify the WooCommerce 'Thank You' page. Specifically, it:

1. Adds an option to change the page title (it uses the same 'page' as the checkout page, so you can't just change the page title normally).
2. Adds an option to change the 'thank you' text.
3. Strips down the content of the page to just the transaction details (date, price, payment method...)
4. Adds an option to show the customer's billing information under the transaction details, but hides it by default.

Understandably, this may not cover what everyone wants to do. One step further would be to also add an option to redirect the 'thank you' page to a specific static page that a publisher could set up. So a publication that was only doing donations could build a page that went into a lot of detail about where the money went, and how else people could get involved, and redirect the 'thank you' page there. There's a filter we can use to do this, and I think it could be a pretty quick easy win for publishers who'd like to craft a specific message for this page.

Taking it even further, there's also a WooCommerce extension that will let you do this same thing, on a product-by-product basis: https://woocommerce.com/products/custom-thank-pages/. This could be helpful for publishers who want to sell multiple different products, and have unique thank you messages for each (donations, access to limited products, physical products, etc). 

1. Start with a site with donations set up for easier testing, and Stripe in testing mode. 
2. Set up a page with the Donate block.
3. On that block, click 'Donate Now'. 
4. Fill out the billing information, and submit the order.
5. View the Thank You page:

image

7. Refresh the page:

image

8. Navigate to Customize > WooCommerce > Thank You Page:

image

10. Try changing the page title, the "Thank You" message, and check Display Customer Details. Publish.
11. View the page again; confirm that your title and message appear, and that the customer details appear at the bottom:

image

### Other information:

### DIFF
--- a/newspack-theme/woocommerce/order/order-details-customer.php
+++ b/newspack-theme/woocommerce/order/order-details-customer.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * Order Customer Details
+ *
+ * This template can be overridden by copying it to yourtheme/woocommerce/order/order-details-customer.php.
+ *
+ * HOWEVER, on occasion WooCommerce will need to update template files and you
+ * (the theme developer) will need to copy the new files to your theme to
+ * maintain compatibility. We try to do this as little as possible, but it does
+ * happen. When this occurs the version of the template file will be bumped and
+ * the readme will list any important changes.
+ *
+ * @see     https://docs.woocommerce.com/document/template-structure/
+ * @package WooCommerce/Templates
+ * @version 3.4.4
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+$show_shipping = ! wc_ship_to_billing_address_only() && $order->needs_shipping_address();
+// temporarily hide shipping for all:
+$show_shipping = false;
+?>
+<section class="woocommerce-customer-details">
+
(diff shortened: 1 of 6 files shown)


newspack-theme/woocommerce/order/order-details-customer.php
```diff
@@ -0,0 +1,68 @@
...
+ * @see     https://docs.woocommerce.com/document/template-structure/
+ * @package WooCommerce/Templates
+ * @version 3.4.4
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+$show_shipping = ! wc_ship_to_billing_address_only() && $order->needs_shipping_address();
+// temporarily hide shipping for all:
+$show_shipping = false;
```
claudiulodro, line 22: Probably want to remove this temporary thing. If correctly configured, the donations should be "virtual products", which means they don't use shipping.

laurelfulford: Thanks Claudiu! I was seeing a shipping address in my testing, but when I tried it on a different test site, I wasn't -- I messed with the product settings at one point, though, breaking how this should work on my localhost.  

I've removed that check in 2b36fec!

Issue #893, opened 2020-04-22
Custom Colours and Typography: Editor preview stopped working


laurelfulford:
Some of the classes have changed in the editor, causing our custom colours and typography not to be applied to the editor.

laurelfulford:
Both the related PRs have been merged, so I'm going to close this issue.

Issue #871, opened 2020-04-08
Button block: Outline style not inheriting colours right


laurelfulford:
The Outline style in the button block has a couple issues:

* When you set the text to white when inside of a cover block, once visited, the text/border will be dark. 
* Hover styles aren't really right (I don't think they should have backgrounds; doing like an opacity shift is probably best). Right now they go dark, which isn't readable against dark backgrounds (unlike the standard buttons, which have a background). 

This may be an opportunity to simplify the Button/Colour styles.

laurelfulford:
This should now be fixed with #899 merged; I didn't include the ticket in that PR.

PR #899, opened 2020-04-24
Update button styles to work with new markup

### LINKED ISSUE #894: Button Block: Updates to Gutenberg cause issues with outline style
Recent changes to Gutenberg have caused a doubled-border effect on the outline-style buttons:

image

A different element in the block is being styled by Gutenberg compared to the theme. Will need to make sure that any fix will work for legacy versions of blocks too (including the previous "Button" version, not just the "Buttons" version), to make sure no sites get broken.

laurelfulford:
### All Submissions:

In Gutenberg 7.9, the button block was updated to remove the `.wp-block-button` container around each link, and add that class to the link itself instead. Unfortunately, the theme relied on that class structure to apply the button styles correctly; this PR updates the styles to work with the new button markup, as well as the old button markup.

The custom colours are not being applied in the editor, even with this fix; #893 may fix this in part but I think I'll need to circle back and do a separate PR for the buttons to fine tune after that's merged. In the meantime, please disregard!

Closes #894 and #871.

1. Make sure you have the latest version of the Gutenberg plugin.
2. Set a custom colour, and a custom typeface.
3. Add button blocks to the editor with the following settings: default; custom text and background color; gradient background; outline styles with default; outline styles with custom text, and outline styles with custom text and background color. 
4. View on the front-end; be disappointed:

image

6. Confirm the front end now looks correct:

image

7. Confirm the editor is at least picking up the right fonts and shapes, but not the colours:

image

8. Turn off the Gutenberg plugin.
9. Edit the page again; add the same types of buttons again. Your original buttons will now be throwing an error, but that can be ignored; the new buttons you're adding will use the old markup, to make sure the styles are backwards compatible. When sites are updated to 7.9+, this will no longer be needed, as the markup gets updated too, but it'll be needed to keep buttons styles working on older versions of the editor.
10. View on the front-end and confirm both sets of buttons are applying custom styles:

image

11. When you turn Gutenberg back on, you'll note the "old" buttons have extra outlines when using the outline style, but this will be fixed Gutenberg (https://github.com/WordPress/gutenberg/pull/21816)

image

### Other information:

### DIFF
--- a/newspack-joseph/inc/child-typography.php
+++ b/newspack-joseph/inc/child-typography.php
@@ -35,7 +35,7 @@ function newspack_joseph_custom_typography_css() {
 			.widget,
 			.widget-title.accent-header,
 			.accent-header,
-			.entry .entry-content .wp-block-button .wp-block-button__link,
+			.entry .entry-content .wp-block-button__link,
 			.entry .article-section-title,
 			button,
 			input[type="button"],
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -71,7 +71,8 @@ function newspack_custom_colors_css() {
 		.entry .entry-content .has-primary-color,
 		.entry .entry-content *[class^="wp-block-"] .has-primary-color,
 		.entry .entry-content *[class^="wp-block-"].is-style-solid-color blockquote.has-primary-color,
-		.entry .entry-content *[class^="wp-block-"].is-style-solid-color blockquote.has-primary-color p {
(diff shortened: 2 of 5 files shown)


laurelfulford:
Thanks Jeff! 

> This was a tricky one to assess with the colors not working in the editor

Yeah, it's a bit of a pain to test -- sorry about that! 😕 I'll be re-reviewing how the buttons preview in the editor with both this and #893 merged in, and make any further adjustments if needed there!

laurelfulford:
Created a new issue to dig into the not-totally-there button preview in the editor:

#905

PR #911, opened 2020-05-01
Button Block: Make sure editor colours are correct

### LINKED ISSUE #905: Button Block: Custom colours not applied in the editor
Custom colours are currently not all applied in the editor:

image

A few more tweaks are needed to get this previewing correctly again (even if they are short-term additions). 

Related: #898

laurelfulford:
### All Submissions:

This PR makes sure the button blocks pick up the custom colours in the editor.

It also fixes an issue with the button outline styles on the front end; the button block markup was rolled back in 8.0, so the "fixed" styles were no longer as fixed as they could be 😅 

Closes #905.

This PR should be tested with both Gutenberg 7.9.1 and Gutenberg 8.0; though the markup changes were rolled back in 8.0, they will be re-added down the road with more lead time so themes have a chance to fix the breaking changes before they're changed.

2. Download and roll back to a copy of [Gutenberg 7.9.1](https://downloads.wordpress.org/plugin/gutenberg.7.9.1.zip), if your site isn't running it already.
3. Add a combination of buttons (solid and outline), using the different custom colours.
4. View in the editor and on the front-end, and confirm your buttons are using the colours you applied:

image

image

5. Upgrade to Gutenberg 8.0:
6. In the editor, re-add the same types of buttons with the same settings; the original set may now be throwing errors in the editor, but it will display correctly on the front-end.
7. Confirm that the second set uses the correct colours in the editor and on the front-end:

image

image

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -638,7 +638,8 @@ function newspack_custom_colors_css() {
 		}
 
 		.block-editor-block-list__layout .block-editor-block-list__block a,
-		.block-editor-block-list__layout .block-editor-block-list__block .is-style-outline .wp-block-button__link:not(.has-text-color),
+		.block-editor-block-list__layout .block-editor-block-list__block .is-style-outline .wp-block-button__link:not(.has-text-color), /* legacy selector */
+		.block-editor-block-list__layout .block-editor-block-list__block .is-style-outline.wp-block-button__link:not(.has-text-color),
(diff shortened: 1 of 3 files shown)


laurelfulford:
Thanks Jeff! This should be fixed in 4daa89f.

I didn't focus enough on the default colours; when testing this again, can you please try with both the default colours and custom colours, just to make sure I didn't somehow introduce something new? 😬

laurelfulford:
Thanks Jeff! As long as nothing else changes, we should be good... 😅

PR #901, opened 2020-04-25
Header: Add simplied header option for subpages

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR adds a theme option to use a very simple header on subpages. This header will only include the logo (no site title or tagline), a search toggle, and a menu toggle; the menu under the toggle includes all menus assigned, and any widgets assigned to the slide out widget area if populated and turned on.

Closes #877.

### How to test the changes in this Pull Request:

1. Apply the PR and run `npm run build`.
2. Navigate to Customize > Header Settings; the Header Options should now be split into three panels, with the existing one split over Appearance and Slide-out Sidebar, and a new one called Subpage Header:

image

image

3. Check the box labelled  "Use simple header on subpages"; it should open an option to upload an Alternative Logo, but ignore that for now:

image

4. View the homepage, and confirm that your header is displaying as expected:

image

5. View a regular single post, and confirm that you now have a simplified header (shown below with the default, solid background, and centered settings):

image

image

image

6. Try opening the menu and confirm that it contains your menus: 

image

7. Check out a post with the 'behind' featured image style; confirm the header sits on top of the image (default and centred logo options shown below; background colours will not be applied to the header, even if assigned): 

image

image

8. Check out a post with the 'beside' featured image styles; confirm the header elements sit over the solid colour:

image

image

9. Try scaling down the browser window -- the mobile header should kick in as usual, with a solid background and logo aligned left, regardless what your header settings are:

image

10. Navigate back to Customize > Header Settings > Subpage Header, and upload an Alternative Logo. The use case here is if you have a dark logo that doesn't show up very well on the Behind and Beside featured image styles, this option can be used to replace it in those cases. So ideally the alternative logo will be a 'flipped' version of your site's real logo:

image

11. View the Featured Image Behind and Featured Image beside posts again; confirm the flipped logo is used:

image

image

12. Scale down the browser window and confirm that the other logo is switched to at the same point the header gets a solid background):

image

image

[...]

Issue #610, opened 2019-12-03
Logo Resizer: Preview sometimes incorrect, inverted


laurelfulford:
Sometimes when you use the logo resizer, it previews "backwards" in the Customizer -- when you try to make it smaller, it gets bigger, and vis versa, and it doesn't actually preview the full range of sizes available. 

The logo will display correctly on the front-end, but this can be really confusing when trying to nail down the size you want.

PR #930, opened 2020-05-13
Header: Correct centered logo sizing

laurelfulford:
### All Submissions:

#901 introduced an issue with the centered logo scaling down on tablet-sized screens; it would stay at its full size, rather than going down to a reduced size for mid-sized screens, before scaling down again for mobile-sized screens.

**Note:** I recommend testing against a live site with a centered logo, using the same logo graphic (like Rivard or OK Watch; Technode too, though there is custom CSS in place to make the logo smaller than default on mobile). This way, it's easy to confirm that the behaviour pre-901 (on the customer site), and with this PR is the same. 

1. Set up a test site header with a centered logo, and other settings to roughly line up with the site you're comparing against (menus, logo size). 
2. View on the front end.
3. Scale down the browser window to roughly tablet-sized (before the mobile menu kicks in). 
4. Contrast against the site you're comparing against; the logo on that site should be shrunk down, but the logo on your test site will be maintaining the same size.
6. Refresh the test site; confirm the logos are the same size.
7. Try different screen sizes and confirm the logo matches the site it's being compared against.
8. Try both AMP and non-AMP, and make sure the logo is sizing similarly for both. 

### Other information:

### DIFF
--- a/newspack-theme/sass/site/header/_site-header.scss
+++ b/newspack-theme/sass/site/header/_site-header.scss
@@ -183,7 +183,6 @@
 .h-cl {
 	@include media( tablet ) {
 		.site-header .middle-header-contain .wrapper > div {
-			flex: 1;
 			width: 35%;
 
 			&.site-branding {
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks Jeff! 

> Here's a weird phenomenon this introduces... In the customizer, the logo now appears to be shrinking as you move the size slider to the right.

😕 This has come up before from time to time (https://github.com/Automattic/newspack-theme/issues/610) but I know it came up again after the subpage header was introduced, too -- it makes sense that these changes are introducing more cases where things might not work as expected.

Worst case, we can remove the JavaScript-updated preview for the short-term -- you would need to wait for the Customizer to reload before seeing the updated logo style, but it might be less disconcerting.

PR #935, opened 2020-05-14
Header: correct logo size display issues with AMP

### LINKED ISSUE #838: Header broken with AMP on smaller screens
**Describe the bug**
With AMP enabled, the site logo image is too wide, and causes a layout issue:

image

With AMP plugin deactivated, it's ok:

image

**Expected behavior**
The header, when AMP plugin is active, should look as it does with the plugin deactivated.

laurelfulford:
### All Submissions:

When scaling down images, AMP will size an image using `object-fit: contain`; if the image's size or max-size isn't the same aspect ratio, this will cause the image to appear centered inside of the max-width space, causing a gap on both sides. This happens when the image is scaled down for mobile, or for the subpage header. 

This PR should fix most cases; the only odd case is when a site is using the subpage header option, and have uploaded an alternative logo for the featured image beside and featured image behind posts that is not the same aspect ratio as the original logo (this is the case with MTFP). The logo won't be sized quite right -- for example, if the original is wider than the alternative version, the alternative version may be scaled down more than it needs to be. This is an odd case, and not the actual use case for that logo, so I think it's safe to fix this on a one-off basis with Custom CSS, but we can revisit and calculate a separate size for that logo if it turns into a bigger thing. 

Closes #838 

1. Set up a header with a left-aligned square logo, and turn on AMP to Transitional mode for easier testing. 
2. Under Customizer > Header Settings > Subpage Header, check 'Use simple header on subpages'.
3. View the AMP version of a page, and scale down the browser window to mobile size -- note the gap that appears on either side:

image

image

image

4. Navigate to a subpage, and note a similar issue in the scaled-down subpage header:

image

6. Revisit both pages, and confirm the logo spacing looks better; also confirm it's the same without AMP:

image

image

7. Test on a subpage with the featured image behind and featured image beside setting, and confirm it's displaying as expected, with and without AMP:

image

image

8. Try some other logo shapes -- horizontal and vertical -- and confirm they all display well with and without AMP:

image

image

image

image

image

image

image

image

image

image

### Other information:

### DIFF
--- a/newspack-theme/inc/logo-resizer.php
+++ b/newspack-theme/inc/logo-resizer.php
@@ -98,7 +98,7 @@ function newspack_customize_logo_resize( $html ) {
 		$mobile_max_height = 65;
 
 		$subhead_max_width  = 200;
-		$subhead_max_height = 50;
+		$subhead_max_height = 60;
 
 		$mobile  = newspack_logo_small_sizes( $img['width'], $img['height'], $mobile_max_width, $mobile_max_height );
 		$subhead = newspack_logo_small_sizes( $img['width'], $img['height'], $subhead_max_width, $subhead_max_height );
--- a/newspack-theme/inc/logo-resizer.php
+++ b/newspack-theme/inc/logo-resizer.php
@@ -152,7 +152,7 @@ function newspack_logo_resize_min_max( $short, $long, $short_max, $long_max, $pe
 }
 
 /**
- * Helper function to return smaller versionf of the logo size
+ * Helper function to return smaller version of the logo size
  */
 function newspack_logo_small_sizes( $width, $height, $maxwidth, $maxheight ) {
 	$smallsize['width']  = round( $maxheight * ( $width / $height ) );
--- a/newspack-theme/inc/logo-resizer.php
+++ b/newspack-theme/inc/logo-resizer.php
@@ -154,16 +154,18 @@ function newspack_logo_resize_min_max( $short, $long, $short_max, $long_max, $pe
 /**
  * Helper function to return smaller version of the logo size
  */
-function newspack_logo_small_sizes( $width, $height, $maxwidth, $maxheight ) {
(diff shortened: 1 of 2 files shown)


newspack-theme/inc/logo-resizer.php
```diff
@@ -128,6 +151,21 @@ function newspack_logo_resize_min_max( $short, $long, $short_max, $long_max, $pe
 	return $size;
 }
 
+/**
+ * Helper function to return smaller versionf of the logo size
```
jeffersonrabb, line 155:
```suggestion
 * Helper function to return smaller version of the logo size
```

newspack-theme/inc/logo-resizer.php
```diff
@@ -128,6 +151,21 @@ function newspack_logo_resize_min_max( $short, $long, $short_max, $long_max, $pe
...
+ */
```
jeffersonrabb, line 156: Here's a few suggested changes to the function for readability and convention. Should be no change whatsoever to the functionality:

```
/**
 * Helper function to return smaller versionf of the logo size
 */
function newspack_logo_small_sizes( $width, $height, $max_width, $max_height ) {
	$size = array(
		'width'  => round( $max_height * ( $width / $height ) ),
		'height' => $max_height,
	);

	if ( $size['width'] > $max_width ) {
		$size['height'] = round( $max_width * ( $height / $width ) );
		$size['width']  = $max_width;
	}

	return $size;
}
```

newspack-theme/inc/logo-resizer.php
laurelfulford: Whoops, thanks! Fixed in 6133c91.

newspack-theme/inc/logo-resizer.php
laurelfulford: Ah, much nicer - thanks! Updated in c249184.

laurelfulford:
Thanks Jeff!

PR #948, opened 2020-05-25
Refactor: Remove style pack code

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

The theme has had a warning about style packs being removed since late February (https://github.com/Automattic/newspack-theme/pull/799), or since the 1.0.0-alpha.33 (pre-1.0.0) release, so it seems like it should be safe to remove the style pack code.

**This PR makes the following updates:** 

* Removes the inc/stylepacks.php, /js/src/style-packs-customizer.js and classes/class-newspack-style-packs-core.php files.
* Updates functions.php to remove references to the above PHP files, and to remove functions to check for style packs, and warn about style packs being used. 
* Removes stylepack-specific Sass files (which just included imports for the related child theme sass files, to generate identical CSS for the style packs). 
* Updated the inc/typography.php and inc/color-patterns.php files to remove references to specific style packs, and to update parent-theme specific styles with a check for `! is_child_theme()` instead of checking for the default style pack.
* In each child theme, remove the customizer.php file that removes the Style Packs panel, and remove references to style packs in code comments, replacing them with 'child theme'
* Remove old style pack-specific styles for Newspack Nelson (ads-related), and Newspack Sacha (editor preview-related); neither style would have been working with the child themes, so will be added to them in separate PRs to keep this as simple as possible _(laughs nervously)_. 
* Updates the author-bio.php in the parent theme to remove checks for style-4. 
* Updates the build script to remove the style pack CSS files. 

### How to test the changes in this Pull Request:

These steps are a little all over the place, but they'll make sure to touch on all the areas of the themes that have been updated:

1. Apply PR and run `npm run build`.
2. If you aren't already, switch to the Newspack parent theme. 
3. Set some custom fonts; and make sure 'Use all-caps for accent text' is checked
4. Set custom colours, and a custom header colour -- both of these will help test to make sure changes to how the parent theme applies custom fonts/colours still work after being updated.
5. Test the following sections with the parent theme:
[...]

PR #951, opened 2020-05-25
Nelson: Move style pack-specific ad styles to the child theme

### LINKED ISSUE #949: Newspack Nelson: Move ad fix to child theme
In the ad styles, there are some styles that should be applied to Newspack Nelson, but they are only applied to the defunct style 2 stylepack: 

https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/sass/plugins/newspack-ads.scss#L21-L32

Styles to remove the margin, and hide `.newspack_global_ad.global_below_header` should be added to the Newspack Nelson child theme.

laurelfulford:
### All Submissions:

While putting together #948, I ran across some styles using the style pack CSS class to make sure below header ads were not displayed in Newspack Nelson, since the theme doesn't really have the space for it. 

This PR moves those styles to Newspack Nelson; they also rely on #948 being merged, since prior to that the styles use `:not( .style-pack-style-2 )` as part of the selector. That class will no longer exist after that PR is merged, but with it in place, this CSS is not specific enough to override it. 

Closes #949.

1. Start with a test site where you can apply ads with Google Ad Manager (I did not have this fully set up on my test site, but enough that I could assign empty ads to spaces).
2. Switch to Newspack Nelson. 
3. Add an ad to the 'Global: Below Header' location.
4. View on the front-end; note how the content area is pushed very far down, and the ad (in this case, my invisible ad), is sitting at the top of the content area:

image

6. Confirm that the content area is not pushed down, and that the ad location is hidden via CSS:

image

### Other information:

### DIFF
--- a/newspack-nelson/sass/style.scss
+++ b/newspack-nelson/sass/style.scss
@@ -65,6 +65,7 @@ Nicolas Gallagher and Jonathan Neal https://necolas.github.io/normalize.css/
 
 .site-content,
 .newspack-front-page.hide-homepage-title .site-content,
+body:not( .newspack-front-page ) .newspack_global_ad.global_below_header + .site-content,
 .site-breadcrumb {
 	@include media( tablet ) {
 		margin-top: #{-3.5 * $size__spacing-unit};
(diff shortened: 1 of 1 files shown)


adekbadek (approved):

laurelfulford:
Thanks Adam!

PR #972, opened 2020-06-12
Fix $content_width setting in theme

laurelfulford:
### All Submissions:

As part of the work for https://github.com/Automattic/newspack-blocks/pull/503, I realized the theme has the incorrect `$content_width`. It's currently set to 960; for the homepage and one-column wide templates, it should be set to 1200; for regular templates it should be set to 780.

$content_width isn't really used with Gutenberg, but I noticed when switching the Newspack Block to not include image sizes, and when Photon was disabled, the 'Large' image size fallback was only up to 960px wide. So this update should fix that.

1. Apply the PR.
2. Confirm that the homepage, regular posts, and a post with the one-column wide template don't throw templates.

### Other information:


laurelfulford:
Thanks Adam!

PR #993, opened 2020-07-02
fix: make sure amp fallback script always loads latest version

laurelfulford:
### All Submissions:

Unlike other scripts in the theme, the amp-fallback.js had a hard-coded version number of `1.0`, rather than one based off of the theme's current version (`wp_get_theme()->get( 'Version' )`). With the current set up, there's a risk that the old file will be cached, even when it's updated, because the version number is static; changing it to dynamically update will make sure browsers load the new version each time there's a new theme release. 

1. Load a non-AMP post or page on your test site.
2. View source and search for `amp-fallback.js`.
3. Confirm that the version number in the query string for the file is `?ver=1.0`.
4. Apply the PR.
5. Confirm that the version number now reflects the theme's dev release -- something like `amp-fallback.js?ver=1.0.0-alpha.33`. On live sites, it will use the 'official' release version (`?ver=1.11.0`).

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -352,7 +352,7 @@ function newspack_scripts() {
 			'hide_order_details' => esc_html__( 'Hide details', 'newspack' ),
 		);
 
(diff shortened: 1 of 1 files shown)


laurelfulford:
Thanks Jeff!

PR #999, opened 2020-07-10
Trust Indicators support

claudiulodro:
### All Submissions:

This PR adds support for Trust Project's Trust Indicators plugin. It does this in the following ways:
- Adds a Type of Work label next to the category label on posts.
- Adds a "Why trust sitename" link in the post byline. This follows @laurelfulford's designs where possible, however I did take a couple liberties in the hope of keeping the maintenance and logic simple.
- Enhances the author archive pages to display additional Trust Project-related info.

A bunch of screenshots are available within the details:
<details>

### Newspack Theme - Post

post-newspack

### Joseph - Post

post-joseph

### Katharine - Post

post-katharine

### Nelson - Post

post-nelson

### Sacha - Post

post-sacha

### Scott - Post

post-scott

### Mobile Post

mobile-post

### Behind Featured Image

behind

### Beside Featured Image

beside

### Author Archives

katharine-author

joseph-author

mobile-author

</details>

0. Get and activate a copy of the Trust Indicators plugin. Let me (or Jeff or Andrew) know if you need a copy.
1. In user settings, fill in the trust fields:

Screen Shot 2020-07-10 at 12 29 59 PM

2. In WP Admin > Settings > Trust Indicators, add a standards page:
Screen Shot 2020-07-10 at 12 30 54 PM

3. At the bottom of the settings, set 'post' to the Article Post types:
Screen Shot 2020-07-10 at 12 31 22 PM

4. On posts you should now see a Type of Work taxonomy, pre-populated with categories. Select one or more and verify the type of work appears on the frontend next to the category, like in the example screenshots above in details.

Screen Shot 2020-07-10 at 12 32 32 PM

5. On a variety of themes, and post settings (behind featured image, etc.), and mobile/desktop verify author archives and posts look good.

### Other information:

### DIFF
--- a/newspack-joseph/sass/style.scss
+++ b/newspack-joseph/sass/style.scss
@@ -257,6 +257,7 @@ input[type='submit'] {
 	@include media( tablet ) {
 		align-items: center;
 		display: flex;
+		flex-wrap: wrap;
 
 		.posted-on {
 			margin-left: #{1.5 * $size__spacing-unit};
@@ -447,14 +448,5 @@ textarea {
 			border-right: 0;
 			border-top: 0;
 		}
-	}
-}
-
-/* Trust Indicators */
-
-@include media( tablet ) {
-	.entry-meta a.trust-label {
-		margin-left: #{1.5 * $size__spacing-unit};
-		margin-top: 0;
 	}
 }
--- a/newspack-joseph/sass/style.scss
+++ b/newspack-joseph/sass/style.scss
@@ -444,3 +444,12 @@ textarea {
 		}
 	}
 }
+
+/* Trust Indicators */
+
+@include media( tablet ) {
+	.entry-meta a.trust-label {
+		margin-left: #{1.5 * $size__spacing-unit};
+		margin-top: 0;
+	}
+}
--- a/newspack-nelson/sass/style.scss
+++ b/newspack-nelson/sass/style.scss
@@ -568,19 +568,18 @@ blockquote {
 
 @include media( tablet ) {
 	.featured-image-beside {
-		.entry-meta a.trust-label {
-			color: #FFF;
+		.entry-meta .trust-label {
+			color: #fff;
 		}
-	} 
-}
-
-.featured-image-behind {
-	.entry-meta a.trust-label {
-		color: #FFF;
 	}
 }
 
-.entry-meta a.trust-label {
-	margin-top: #{.5 * $size__spacing-unit};
+.featured-image-behind {
+	.entry-meta .trust-label {
+		color: #fff;
+	}
 }
 
+.entry-meta .trust-label {
+	margin-top: #{0.5 * $size__spacing-unit};
+}
--- a/newspack-nelson/sass/style.scss
+++ b/newspack-nelson/sass/style.scss
@@ -558,3 +558,24 @@ blockquote {
 		}
 	}
 }
+
+/* Trust Indicators */
+
+@include media( tablet ) {
+	.featured-image-beside {
+		.entry-meta a.trust-label {
+			color: #FFF;
+		}
+	} 
+}
+
+.featured-image-behind {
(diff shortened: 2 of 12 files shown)


claudiulodro:
This looks much improved! Thanks @laurelfulford!

> On the author archive page, why is the sidebar hidden? I’m just wondering about centering everything under it; the alignment between the edge of the author info and stories doesn’t quite look intentional:

Good point. It was part of the publisher's spec, but I don't think it necessarily makes sense for all sites. I've removed it in https://github.com/Automattic/newspack-theme/pull/999/commits/356af2f877243d28cb48a1bd25654122877d611f , and we can just use some custom CSS on the site if they absolutely require the sidebar hidden there.

> Before tweaking styles, I wanted to ask what is the purpose of the Type of Works taxonomy? Right now, they look like the categories, but they aren’t links like the categories are. When using the default theme, they aren’t differentiated (they both use the same solid background colour), but they are quite distinct in the other themes, since they don’t. I feel like they should perhaps be moved, or at least styled not to look like the categories in the default theme (and not be visually stronger than categories, like in the child themes), but I could be totally misunderstanding them!

The terms are for "labeling the type of journalistic work to distinguish factual reports from analysis, opinion and other genres." There's no reason they need to look in any particular way and I don't believe anyone is attached to a specific presentation, as long as a clear label exists. [Here's the spec if you want to look at it.](https://docs.google.com/document/d/1pLUM-JvGDl5NMUNMi2Hl74MnsoBGw_B3cYIAQ8NyO_M/edit)

> These labels are showing up in the search results right now — should they be?

Nope! Good catch. I've fixed in https://github.com/Automattic/newspack-theme/pull/999/commits/1c4affc5efcc37a835373492ec77fc0edcf390cd

> It seems unlikely, but would these labels ever be shown on Sponsored Content? Right now the plan is to add a label in front of the categories for sponsored content, but I don't want that to get too confusing; it may already get moved to make sure it stands out more, but it can also be moved to keep things tidy :)

"Sponsored" is actually one of the default types of work available, so a site using Trust Indicators and with sponsored content would 100% have one of those labels. I think the best solution may be to not output the type of work on sponsored posts, otherwise it'll say something like "SPONSORED POST_TITLE CATEGORY SPONSORED". (not sure what your plans are for the upcoming sponsored post feature, though)

laurelfulford:
Thanks @claudiulodro!

> There's no reason they need to look in any particular way and I don't believe anyone is attached to a specific presentation, as long as a clear label exists. Here's the spec if you want to look at it.

This was super helpful -- thank you! I reworked the visuals on these a bit in 8b606b1, keeping them in the Category area, but moving them in front with a colon, like:

image

Though the Trust Project doesn't recommend assigning more than one based on the spec, if you do, the types will display in  a comma-separated list:

image

In the other child themes, it's not as distinctly separate (since the categories don't have backgrounds), but they are still a different colour in most cases, excluding Newspack Joseph: 

image

... and then when there's the featured image beside or behind, they'll just inherit the default colour:

image

> (not sure what your plans are for the upcoming sponsored post feature, though)

The original plan for the sponsored posts was to add 'Sponsored' text in a similar location/style to the above, but it was (rightly!) pointed out that that, specifically, should probably be more noticeable :) It might end up with the same placement -- so not showing both the trust project and our sponsored labels on a post would be ideal -- but I'm going to explore also moving it above, and definitely going for stronger visuals (like colour) to help it stand out more than the original plan.

claudiulodro:
That looks much better than what I came up with! It looks good in my testing. Thanks for all your help.

SanjayKumarMonu:
@claudiulodro from where we could get the Trust Indicator Plugin?

Can you provide the same to me to so that I can test it on my website.

claudiulodro:
@SanjayKumarMonu you need to apply and be accepted by [The Trust Project](https://thetrustproject.org/).

PR #1001, opened 2020-07-15
Make secondary menu visible when using the short header

### LINKED ISSUE #991: Header: Allow secondary menu when using the 'short' header
Right now when you pick the 'short' header option, it only includes the primary and tertiary menus -- the social menu is displayed in the footer, and the secondary menu is not displayed at all.

Given that it doesn't add much height, it seems like it'd be pretty harmless to include the secondary menu in that case as well, if assigned (and if not, still don't show it). 

That section should not show if only the social menu is assigned (to prevent it from suddenly appearing on some sites), but both menus should probably display when they are both assigned; if the social menu is unwanted there, it can always be hidden with Custom CSS. This behaviour could possibly be fine-tuned with another theme option.

laurelfulford:
### All Submissions:

Right now, when you're using the 'short' header, the secondary menu will not appear on desktop, even if assigned. 

There's not really a great reason for this, since the short header is still shorter than the standard header, and you can have that section not-appear simply by not assigning a secondary menu. 

This PR updates the short header to act like the regular header: if the secondary menu is assigned, it will appear there; if the social links menu or slide out menu is assigned, they will also appear in the top bar _only_ when the secondary menu is assigned. Otherwise, they'll appear in their fallback positions (in the footer or to the left of the logo, respectively). 

As part of this change, I went through all of our current live and staging sites, and just found one that uses the Secondary Menu with the short menu (CalMatters) -- in their case, it acts as a simplified menu for mobile, and I was able to preemptively add some CSS to hide it so this change won't cause issues.

Closes #991.

1. Enable AMP, and open the Console to watch for errors; this will help make sure no errors are thrown when AMP is trying to move a menu from the mobile menu space to the desktop menu space, but can't find the containing element. 
3. Navigate to Customizer > Header Settings > Appearance, and uncheck 'Center Logo' and 'Short Header' if checked.
4. Add a menu to the primary, secondary, tertiary, social locations, and turn on and add widgets to the Slide-out Sidebar (under Header Settings > Slide-out Sidebar). 
5. View on the front-end -- you should have something like: 

image

6. Disable the Secondary Menu; confirm that the top section disappears, and the menu toggle moves next to the logo (this is a deviation from current behaviour, but if any publisher wants the toggle at the top, assigning an empty menu to the Secondary menu slot is a good work-around). The social links will not 'keep' this menu section open on their own, which matches the current behaviour -- they will appear in the footer:

image

7. Under Header Settings > Appearance, check 'Center Logo'; confirm the slide-out toggle appears on the left:

image

8. Under Header Settings > Appearance, uncheck 'Center Logo' and check 'Short Header'; confirm the slide-out menu toggle appears to the left of the logo: 

image

9. Assign a secondary menu; confirm the top bar now appears, with the secondary menu and slide-out menu toggle:

image

10. Under Header Settings > Appearance, center the logo:

image

11. Remove the Secondary Menu again; confirm the menu toggle moves to the left of the primary menu:

image

### Other information:

### DIFF
--- a/newspack-theme/header.php
+++ b/newspack-theme/header.php
@@ -77,7 +77,7 @@ endif;
 			<?php if ( has_nav_menu( 'secondary-menu' ) || ( true === $show_slideout_sidebar && false === $header_simplified ) ) : ?>
 				<div class="top-header-contain desktop-only">
 					<div class="wrapper">
-						<?php if ( true === $show_slideout_sidebar ) : ?>
+						<?php if ( true === $show_slideout_sidebar && has_nav_menu( 'secondary-menu' ) ) : ?>
 							<button class="desktop-menu-toggle" on="tap:desktop-sidebar.toggle">
 								<?php echo wp_kses( newspack_get_icon_svg( 'menu', 20 ), newspack_sanitize_svgs() ); ?>
 								<?php echo esc_html( get_theme_mod( 'slideout_label', esc_html__( 'Menu', 'newspack' ) ) ); ?>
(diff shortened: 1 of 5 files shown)


laurelfulford:
Removing 'Needs Review' briefly; I don't think it makes sense to keep the menu toggle at the top when it's alone; I did that because I thought it was needed to prevent changes on one of the sites, but just realized a work around could be adding an empty menu to the secondary menu location. Reworking a bit!

laurelfulford:
Updated the logic a bit, as well as the testing steps/screenshots; setting back to needing review again.

laurelfulford:
Thanks Derrick!

laurelfulford:
The Header and Menu docs have been updated so they no longer say this menu is not available when using the short header settings.

PR #1058, opened 2020-09-04
Image Block: prevent gaps above captions in mobile Safari

### LINKED ISSUE #1055: Image Block: Caption placement in mobile Safari
**Describe the bug**
In some cases, when viewing the image block in mobile Safari, there's a gap between the image and its caption, and the caption overlaps on the text following it.

IMG_1485

This seems to happen when the image is centered, and can 'fix' itself when you try to inspect it using the iOS Simulator. 

Switching the image and caption to use `display: block` instead of `display: table` seems to fix, but it needs further testing to make sure it doesn't introduce different issues.

laurelfulford:
### All Submissions:

Mobile Safari randomly adds a lot of space above captions on page load, though they eventually snap into place. This can be a bit disconcerting, though. The issue seems to be related to `aligncenter` and the block using `display: table`; this PR fixes the issue by switching the block to use `display: block`.

Closes #1055.

1. Set up an image block with a caption, and align the block center.
2. View on an iOS device in Safari, or using the iOS simulator on Mac. Refresh a few times; turning on AMP can also help in recreating the issue more consistently, though I'm not sure why.
3. Refresh until you observe this weird issue:

IMG_1507

5. Clear your browser cache; repeat refreshing/reloading the page to make sure the issue doesn't reoccur (unfortunately, since it doesn't happen all the time, it needs to be hammered a bit to verify):

IMG_1508

6. Confirm on desktop that there are no new issues introduced with this change.

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -702,16 +702,19 @@
 			display: block;
 		}
 
-		.aligncenter {
-			margin: 0 auto;
-		}
-
 		figcaption {
 			text-align: left;
 		}
 
-		.alignleft > figcaption,
-		.alignright > figcaption {
+		.alignleft,
+		.aligncenter,
+		.alignright {
+			figcaption {
+				display: block;
+			}
+		}
+
+		.aligncenter {
 			display: block;
 		}
 	}


adekbadek (approved):
The issue was reproducible right off the bat for me on an AMP page, no need for refreshes.

laurelfulford:
Right on - thanks for testing, Adam!

PR #1059, opened 2020-09-07
Add 'time ago' date option to theme

laurelfulford:
### All Submissions:

This PR adds an option to display post (and block) dates using a 'time ago' option, and set a cut off for when this date format is applied in seconds (the default is two weeks). 

2. Navigate to Customizer > Template Settings, and check "Use 'time ago' date format"; as part of this, confirm that the "Cut off for 'time ago'..." section is not visible by default, but appears when "Use 'time ago' date format" is checked:

image

3. View posts of various ages, on single posts, archives, search, and in the Homepage Posts and Post Carousel blocks, and confirm posts newer than two weeks old use the 'time ago' format:

Blocks:
image

image

Single post:
image

Search:
image

Archive:
image

4. Confirm that the date format is not being picked up where it shouldn't be (like in the archive date header for www.yoursite.com/2020/09, or the Recent Posts widgets when dates are set to show (dates in widgets seem to get confused and pick up the current post date, not their actual dates)).
5. Navigate to Customize > Template Settings, and change the cut off from 14 days to a higher or lower value.
6. Confirm that the new cut off is being accurately picked up eg. if you changed it to 7 days, confirm that posts older than 7 days are now showing the date format set in WordPress, not the 'time ago' format.

Notes: 
* The `'Y-m-d\TH:i:sP' !== $format` check is to make sure `get_the_date( DATE_W3C )` is not filtered (which is used for the `datetime` attribute in the `time` tag).
* `DateTime()` and `$date->getTimestamp()` are used instead of `current_time( 'u' )` to avoid this PHPCS error and potential issue: `Calling current_time() with a $type of "timestamp" or "U" is strongly discouraged as it will not return a Unix (UTC) timestamp. Please consider using a non-timestamp format or otherwise refactoring this code. (WordPress.DateTime.CurrentTimeTimestamp.Requested)`

### Other information:

### DIFF
--- a/newspack-theme/footer.php
+++ b/newspack-theme/footer.php
@@ -17,6 +17,7 @@
 
 	<footer id="colophon" class="site-footer">
 
+		<?php remove_filter( 'get_the_date', 'newspack_convert_to_time_ago', 10, 1 ); ?>
 		<?php get_template_part( 'template-parts/footer/footer', 'branding' ); ?>
 		<?php get_template_part( 'template-parts/footer/footer', 'widgets' ); ?>
 
--- a/newspack-theme/inc/customizer.php
+++ b/newspack-theme/inc/customizer.php
@@ -643,7 +643,7 @@ function newspack_customize_register( $wp_customize ) {
 		)
 	);
 
-	// Add option to hide the whole author bio.
+	// Add option to set a featured image default.
 	$wp_customize->add_setting(
 		'featured_image_default',
 		array(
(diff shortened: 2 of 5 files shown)


thomasguillot (approved):
Works as expected! 😊

Should we have the same option for comments?

laurelfulford:
Thanks Thomas! 

> Should we have the same option for comments?

We could! I can build on it in a separate PR. I'm leaning towards adding a separate control (since it would need a separate filter anyway), but let me know if you think it should be controlled by the same one.

thomasguillot:
> I can build on it in a separate PR. I'm leaning towards adding a separate control (since it would need a separate filter anyway).

Yes and yes 😊  Also in the customiser we have "comments settings" section where it'd fit nicely.

laurelfulford:
Cool! I've created a separate issue for the comments:  #1064.

philipjohn:
This is great, Laurel, thank you for looking at it so quickly!

laurelfulford:
Documentation added here: https://newspack.pub/support/themes/date-formats/

PR #1060, opened 2020-09-08
Add simplified gradient set

laurelfulford:
### All Submissions:

This PR replaces Gutenberg's default set of gradients with a simplified set of all grey options. The purpose is to keep the options available, but to remove the examples Gutenberg provides; they're great to show off what gradients can do, but don't really fit in visually on news sites. Hopefully this will encourage use of the simpler options, and exploring custom gradients that better fit in with existing site colour schemes. 

I'd be interested both in testing feedback, and a yay-nay opinion on the gradients. I had also considered including some really basic gradients (blue, red, yellow), but I'm not sure if that would negate making this change at all. 

1. Set up at least one gradient using the default set, like as a group background.
3. Confirm that the default gradients are replaced with a smaller set: 

image

4. Confirm that the gradient you added still works. 
5. Apply each of the new gradients to an element, like a group block, and confirm it works in the editor and on the front-end:

image

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -210,6 +210,42 @@ if ( ! function_exists( 'newspack_setup' ) ) :
 			)
 		);
 
+		add_theme_support(
+			'editor-gradient-presets',
+			array(
+				array(
+					'name'     => __( 'Black to medium gray', 'newspack' ),
+					'gradient' => 'linear-gradient(135deg, rgb(17, 17, 17) 0%, rgb(85, 85, 85) 100%)',
+					'slug'     => 'grad-1',
+				),
+				array(
+					'name'     => __( 'Medium gray to light gray', 'newspack' ),
+					'gradient' => 'linear-gradient(135deg, rgb(85, 85, 85) 0%, rgb(221, 221, 221) 100%)',
+					'slug'     => 'grad-2',
+				),
+				array(
+					'name'     => __( 'Light gray to white', 'newspack' ),
+					'gradient' => 'linear-gradient(135deg, rgb(221, 221, 221) 0%, rgb(255, 255, 255) 100%)',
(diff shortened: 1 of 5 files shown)


thomasguillot:
This is a great idea and it works well.

How about adding 2 more gradients?
* Primary to Primary Variation
* Secondary to Secondary Variation

So users can have gradients with their selected colours.

laurelfulford:
> This is a great idea and it works well.

I can't take credit -- this was suggested by @jeffersonrabb :) 

> How about adding 2 more gradients?

Oh, good idea! I'll work on that today.

laurelfulford:
@thomasguillot Primary and secondary options added in 1ef96d9! 

A couple notes:

* This should be tested with the default colour palette, and with custom colours, on the front-end and in the editor, just to make sure nothing is funky. 
* I opted to make these gradients the opposite of the greys (light to dark, instead of dark to light) because publications often use grey as a secondary colour; hopefully this helps make them a bit different if there are near-matching greys in the unchanging set. Also open to flipping them.
* I reordered them (adding the primary/secondary first), so the CSS classes changed.

thomasguillot (changes requested):
For some reason, radial gradient doesn't work with primary/secondary colours (`#`) seems to be missing.

Kapture 2020-09-08 at 17 33 37
Screenshot 2020-09-08 at 17 34 00

laurelfulford:
Oh good catch, Thomas!

It looks like the radial gradients 'want' the colours to be set as rgb in the functions.php, so I've updated that.

It makes sense in retrospect, but these styles are also added inline (instead of using the CSS classes), so once a radial gradient is set up, unlike the regular one, it won't change when you change the custom colours.

thomasguillot (approved):
🚢

laurelfulford:
Thanks!

PR #1080, opened 2020-09-17
fix: relative time filter should use the correct $post object

dkoo:
# All Submissions:

This updates the `get_the_date` filter callback `newspack_convert_to_time_ago` to use the `$post` object returned by the filter, rather than the global `$post`. The global `$post` causes this filter to output the formatted date from the main post query object for all post dates rendered on a given page.

Closes https://github.com/Automattic/newspack-plugin/issues/671.

1. Before checking out this branch, go to Jetpack > Settings > Traffic and ensure that the Related Posts module is turned on.
2. Visit a single post while logged in, then click Customize from the admin toolbar.
3. Go to Related Posts and turn ON the "show date" option.
4. Still in the Customizer, go to Theme Settings and turn ON the “Use "time ago" date format” option. Set the cut-off option to something long enough that it will affect the posts in your environment.
5. View a single post that has a publish date within the cut-off time (so that it shows the relative date).
6. Scroll down to the Related Posts module and observe that the relative publish dates shown for all related posts incorrectly match the publish date of the main post.
7. Check out this branch and refresh the post. Now observe that the publish dates shown for each related post match the publish date for that post, rather than the main post.

### Other information:

### DIFF
--- a/newspack-theme/footer.php
+++ b/newspack-theme/footer.php
@@ -17,7 +17,7 @@
 
 	<footer id="colophon" class="site-footer">
 
-		<?php remove_filter( 'get_the_date', 'newspack_convert_to_time_ago', 10, 2 ); ?>
+		<?php remove_filter( 'get_the_date', 'newspack_convert_to_time_ago', 10, 3 ); ?>
 		<?php get_template_part( 'template-parts/footer/footer', 'branding' ); ?>
 		<?php get_template_part( 'template-parts/footer/footer', 'widgets' ); ?>
 
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -540,8 +540,7 @@ function newspack_the_custom_logo() {
 /**
  * Change date to 'time ago' format if enabled in the Customizer.
  */
-function newspack_convert_to_time_ago( $post_time, $format ) {
-	global $post;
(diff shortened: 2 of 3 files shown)


jeffersonrabb (changes requested):
I'm getting this fatal when navigating to a date archive page:

```
PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function newspack_convert_to_time_ago(), 2 passed in /srv/www/wordpress-simple/public_html/wp-includes/class-wp-hook.php on line 289 and exactly 3 expected in /srv/www/newspack-repos/newspack-theme/newspack-theme/inc/template-functions.php:543
```

Looks like the accepted args parameter needs to be raised to 3 every time a filter is added or removed with `newspack_convert_to_time_ago` as the callback. This issue is in:

https://github.com/Automattic/newspack-theme/blob/fix/relative-time-filter/newspack-theme/footer.php
https://github.com/Automattic/newspack-theme/blob/fix/relative-time-filter/newspack-theme/sidebar.php
https://github.com/Automattic/newspack-theme/blob/fix/relative-time-filter/newspack-theme/inc/template-functions.php

dkoo:
> Looks like the accepted args parameter needs to be raised to 3 every time a filter is added or removed with `newspack_convert_to_time_ago` as the callback.

Ack, good catch! Fixed in af062dc.

jeffersonrabb (approved):
That did the trick!

PR #1090, opened 2020-09-28
fix: switch menu dropdowns to use primary color to improve contrast

### LINKED ISSUE #1087: Menus: dropdown hover sometimes has insufficient contrast
Right now the primary and secondary menus use a slightly darkened shade of the primary colour as the background on hover.

This can result in a low-contrast situation, though, since the theme also uses the primary colour itself to figure out whether the text on hover should be black or white. If the primary colour is kind of in the 'middle', the contrast code may pick black as the text colour, but when you darken the primary colour slightly white would be more appropriate.

We should either switch this to use the original primary colour, or calculate the appropriate colour to use for sufficient text contrast based on the darkened primary colour, and not the original.

laurelfulford:
### All Submissions:

The menu dropdowns originally used a darker version of the primary colour as a background on hover. This works well with the default blue for contrast (since the text would be white against the primary colour and the darker version of the primary colour), but in some cases, there isn't sufficient contrast. You can pick a primary colour that's light enough that the theme would pair it with black text, but the darker version of the primary colour as a background would make that difficult to read. 

This PR switches the dropdowns to just use the primary colour. 

Closes #1087.

1. Set up a primary menu with dropdowns.
2. Set a primary colour that's dark-ish, but would use black as a text colour -- for example, `#f25800`.
3. View the dropdown menu:

image

5. Check the dropdown menu again; the background will be a bit lighter, and have better contrast:

image

### Other information:


thomasguillot (commented):
Nice one, it works as expected.

Visually I do find it quite aggressive on the eye though. Have you thought about using a darker, almost black, shade of grey, with maybe an underline for the hover? It would make it more subtle but at the same time more balanced.

laurelfulford:
> Have you thought about using a darker, almost black, shade of grey, with maybe an underline for the hover? It would make it more subtle but at the same time more balanced.

That's a good point, especially with some of the bolder primary colours (especially with it flipping from white to black text and back!).

I've updated it to just swap in the darker hover colour in 0581a07; the text decoration would help it pop more, but I waffled about that since it's a lot more of a noticeable change to push to live sites. Let me know if you think this will work or if it'd benefit from popping more!

thomasguillot (approved):
This is much better 😊

laurelfulford:
Thanks Thomas!

PR #1091, opened 2020-09-29
Newspack Scott: Make sure footer widgets pick up correct custom font

### LINKED ISSUE #1066: Newspack Scott: Font issues
Newspack Scott is not inheriting the custom fonts in a couple locations:

* The tagline
* Some footer widgets

This should also be tested in the other child themes.

laurelfulford:
### All Submissions:

The CSS styling the footer widgets in Newspack Scott is too specific, so the custom fonts don't override what the child theme is setting with certain widgets.

See #1066.

1. Switch your test site to Newspack Scott.
2. Add at least one text widget to the footer and above copyright areas; add a few other widget types to the mix, too, to test.
3. Apply a custom header font. 
4. View on the front end; note that the text widget isn't picking up your custom header font like the other widgets are -- in this example, it should be using Impact but it's still using the default Fira Sans Condensed:

image

6. Confirm that the widgets are now picking up your custom fonts:

image

7. Switch back to the default fonts and confirm that they are also getting picked up correctly.

### Other information:

### DIFF
--- a/newspack-scott/sass/style.scss
+++ b/newspack-scott/sass/style.scss
@@ -305,7 +305,6 @@ Nicolas Gallagher and Jonathan Neal https://necolas.github.io/normalize.css/
 	}
 }
 
-#colophon .widget,
 // If custom font is set, `#colophon .site-info` selector would override it
 // (See issue #663)
 .site-info {


thomasguillot (commented):
It certainly fixes the issue but I don't think the "Heading" font should be used as a body font.

laurelfulford:
> It certainly fixes the issue but I don't think the "Heading" font should be used as a body font.

Fair enough :) This is how the widgets are set up in most of the themes, so I think it should be handled in a different issue -- Newspack Scott just had too specific of styles in the footer to pick up the custom fonts. 

I think I'd prefer to keep the header font in cases like, menus and lists, and change it to the body for paragraphs (in part, to avoid making drastic changes to live sites), but happy to discuss further! Let me know what you think, and I can create a new issue for that change.

thomasguillot:
> This is how the widgets are set up in most of the themes, so I think it should be handled in a different issue

Ok no problem. I guess for future child themes we should reconsider this. Let's keep out current child themes as is then.

thomasguillot (approved):
👍

laurelfulford:
Thanks Thomas! It is a good point -- I think it worked well for the default fonts in the child themes, but it makes sense that it wouldn't work for all possible header options -- my choice of Impact was especially egregious, since it's bold all the time :)

PR #1104, opened 2020-10-07
Add option to turn off Yoast's primary category, so all categories display

### LINKED ISSUE #861: Selectively use Yoast's "primary category" feature

The theme [currently uses](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/inc/template-tags.php#L150-L160) the "Primary Category" as provided for by the Yoast SEO feature. The code comment says (my emphasis) "Only display Yoast primary category *if set*" but it's not possible to _not_ set a primary category:

Screenshot from 2020-04-02 12-14-36

Some publishers use a well thought-through category structure and expect them all to show on a single article page, so it seems a little redundant to them for the theme to only print out one.

A better solution may be to make the theme reflect the wording in the [code comment](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/inc/template-tags.php#L150):

> Only display Yoast primary category if set.

This would probably require changing the category drop-down Yoast inserts to add a default option of "None" or similar.

An alternative could be that there's a more general theme setting, in the customiser, that tells the theme whether or not to use the "Primary Category" feature globally. This might result in the drop-down being hidden when disabled. I'd suggest that in this scenario the Primary category should be disabled by default, but open to opposing viewpoints.

**Additional context**

This came out of a specific request from a Newspack publisher.

laurelfulford:
### All Submissions:

This PR adds an option to the Customizer to turn off Yoast's Primary Category feature, so you can display all of your categories on theme pages (single posts, search, blog posts page). 

This change doesn't affect the Homepage Posts block; if we do opt to add a control for that it will need to be done on the block level, since the block uses a combination of Yoast's Primary Category feature, with a 'grabbing the first category' fallback.

Closes #861 

1. Start on a site with Yoast enabled. 
2. Set up a post with multiple categories. 
3. Apply the PR.
4. View on the front-end; confirm that only one category is showing:

image

5. Navigate to the Customizer > Template Settings; confirm you now have a checkbox to enable Yoast's Primary Category feature, and that it's enabled by default:

image

6. Uncheck the primary category option.
7. Confirm that it updates the category display to show all of your posts categories:

image

8. Uncheck the option and confirm the post goes back to showing your primary category.
9. Disable the Yoast plugin and confirm that the category option disappears from the Customizer > Template Settings panel.

### Other information:

### DIFF
--- a/newspack-theme/inc/customizer.php
+++ b/newspack-theme/inc/customizer.php
@@ -723,6 +723,25 @@ function newspack_customize_register( $wp_customize ) {
 		)
 	);
 
+	// Add option to turn off Yoast's Primary Category functionality.
+	if ( class_exists( 'WPSEO_Primary_Term' ) ) {
+		$wp_customize->add_setting(
+			'post_primary_category',
+			array(
+				'default'           => 'true',
+				'sanitize_callback' => 'newspack_sanitize_checkbox',
+			)
+		);
+		$wp_customize->add_control(
+			'post_primary_category',
+			array(
+				'type'    => 'checkbox',
+				'label'   => __( 'Use Yoast\'s primary category functionality', 'newspack' ),
+				'section' => 'post_default_settings',
+			)
+		);
+	}
+
 	/**
 	 * Comments settings
 	 */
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
:shipit:

laurelfulford:
Thanks!

laurelfulford:
Documentation added here: https://newspack.pub/support/themes/yoast-primary-category/

PR #1116, opened 2020-10-19
fix: make pullquote citations inherit correct color

### LINKED ISSUE #1115: Pullquote block: Citation not picking up custom colour
In the pullquote block the citation doesn't always pick up the custom colour on the front-end -- this should be tested (it may be specific to only some child themes) and fixed.

One possible approach would be to swap the different colour text for the citation with a semi-transparent version of the main colour.

laurelfulford:
### All Submissions:

This PR primarily fixes a couple issues where the citation in the pullquote block is not picking up the correct colour:

* In some cases when a text colour is set, the citation is not picking it up on the front-page.
* When nested in a cover block, the quote itself automatically switches to white, but the citation does not on the front end. 

Closes #1115

1. Set up a couple pullquote blocks -- one default, one with a text colour, one with a background colour, and one in a cover block without any text colour set (it should switch to white on its own). 
2. Save and publish.
3. View on the front-end; note the quotes with the text colour still have grey citations, and the one in the cover block has a nearly illegible citation:

image

image

5. Confirm that the citations are now picking up the same text colour as the rest of the pullquote blocks:

image

image

6. Since its the only child theme that was edited, confirm that the colours are displaying as expected on the front-end with Newspack Katharine:

image

image

### Other information:

### DIFF
--- a/newspack-katharine/sass/style.scss
+++ b/newspack-katharine/sass/style.scss
@@ -283,10 +283,6 @@ figcaption,
 		}
 	}
 
-	cite {
-		color: $color__text-light;
-	}
-
 	cite,
 	&.is-style-solid-color blockquote cite {
 		text-transform: uppercase;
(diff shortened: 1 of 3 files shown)


adekbadek (approved):
😎

laurelfulford:
Thanks Adam!

Issue #751, opened 2020-02-02
Pages: Add featured image placement options to pages, too


laurelfulford:
It would be handy to make the same featured image placement options -- available for posts -- to be available for pages as well. 

Even being able to hide the featured image would be helpful -- unlike posts they're not used elsewhere on the site but they would still be used for social media.

philipjohn:
This was requested in zen-2898636

PR #1120, opened 2020-10-21
Make sure featured image placements aren't applied to CPTs

### LINKED ISSUE #1119: Featured Images: Custom post type picking up featured image defaults
This issue needs investigation:

One staging site has a custom post type that's picking up the 'featured image behind' setting in the Customizer. 

This needs to be replicated on a test site, and fixed.

laurelfulford:
### All Submissions:

Right now, if you set a default featured image placement in the Customizer, it will be applied to all CPTs, not just posts. 

This PR adds a check to apply the 'small' image size to any CPTs. 

Closes #1119.

1. Set up the Newspack Listings plugin (currently in dev; ping for a link), and set up a listing with a larger featured image.
2. Set up a regular post with a large featured image.
3. Navigate to Customizer > Template Settings, and set the default featured image placement to 'behind'.
4. View both of your posts on the front-end; note that the behind image placement is also picked up by the listing, but can't be turned off because it's not a CPT option:

Listing CPT:
image

Regular post: 
image

5. Apply the PR.
6. Confirm that the featured image is now displaying as 'small' on your CPT:

image

7. Confirm that the 'behind' image placement is still being applied to your regular post.

### Other information:

### DIFF
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -140,12 +140,14 @@ function newspack_body_classes( $classes ) {
 	}
 
 	// Adds special classes, depending on the featured image position.
-	if ( 'behind' === newspack_featured_image_position() ) {
-		$classes[] = 'single-featured-image-behind';
-	} elseif ( 'beside' === newspack_featured_image_position() ) {
-		$classes[] = 'single-featured-image-beside';
-	} elseif ( is_single() ) {
-		$classes[] = 'single-featured-image-default';
+	if ( is_singular( 'post' ) ) {
+		if ( 'behind' === newspack_featured_image_position() ) {
+			$classes[] = 'single-featured-image-behind';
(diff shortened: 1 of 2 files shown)


adekbadek (commented):
This works as described, but I wonder how would a user turn on the behind-title featured image placement for a CPT? Or do we want this placement to be available _only_ for regular posts?

laurelfulford:
> This works as described, but I wonder how would a user turn on the behind-title featured image placement for a CPT? Or do we want this placement to be available only for regular posts?

The placement was only intended to work for regular posts, though I am working on another issue to add a separate option for pages, too (#751). The CPT in this case was picking up the post default by accident, and they don't have their own per-post option to override it. 

It could make sense to add for specific CPTs down the road, but we'd probably want to do it on a per-CPT basis, rather than a blanket option -- the default setting you might want for Listings is likely different than what you'd want for WooCommerce Products, etc :)

laurelfulford:
Thanks Adam!

PR #1131, opened 2020-10-30
Handle theme colors outside post content

adekbadek:
### All Submissions:

https://github.com/Automattic/newspack-popups/pull/292 introduces an "Above site header" placement option. 

1. Follow testing steps in https://github.com/Automattic/newspack-popups/pull/292
2. On `master`, observe the theme colors are not applied to blocks used in an "above header" campaign
3. Switch to this PR, observe that the theme colors are applied

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -41,12 +41,12 @@ function newspack_custom_colors_css() {
 			.mobile-sidebar,
 			/* Header default background; header default height */
 			body.h-db.h-dh .site-header .nav3 .menu-highlight a,
-			.has-primary-background-color,
-			*[class^="wp-block-"].has-primary-background-color,
-			*[class^="wp-block-"] .has-primary-background-color,
-			*[class^="wp-block-"].is-style-solid-color,
-			*[class^="wp-block-"].is-style-solid-color.has-primary-background-color,
-			.wp-block-file .wp-block-file__button,
+			.entry .entry-content .has-primary-background-color,
+			.entry .entry-content *[class^="wp-block-"].has-primary-background-color,
+			.entry .entry-content *[class^="wp-block-"] .has-primary-background-color,
+			.entry .entry-content *[class^="wp-block-"].is-style-solid-color,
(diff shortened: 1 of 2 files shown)


laurelfulford (changes requested):
Unfortunately, removing the `.entry .entry-content` classes from the custom colour code cause them to stop working in the content area :( This is largely because these the styles in the theme overall are needlessly complicated; I spent time early on trying to move blocks out of that selector, but they're kind of in a half-in-half-out status at the moment. 

I think there's a couple possible alternatives:

* For the short-term, just duplicate the color styles with the `.newspack-inline-popup` class instead of the `.entry .entry-content` class. Pros: requires very little testing, and has the least potential to be 'dangerous' for existing styles. Cons: adds a lot more CSS. 

Or:

* Go through the _blocks.scss and move all of the colour-related styles outside of the `.entry .entry-content` selector. It's not a full move away from that selector, but it should enable us to simplify the colour styles to apply everywhere. Pros: Doesn't add a lot of CSS bulk. Cons: will require more testing to make sure existing styles in the content aren't broken. 

I'm happy to take a swing at option two and see if it's as 'easy' an approach as I hope it will be. If timing is an issue, it might be best to go with option one for the short-term. 

Just let me know if you have any questions at all about this!

adekbadek:
It seems like the changes in `color-patters.php` were not needed after all – the theme colors are picked up with the change to `_blocks.scss` only.

laurelfulford (changes requested):
Unfortunately, because the custom colour and text styles use the `.entry .entry-content`, they aren't applied to the campaigns.

Here is my terrible campaign in the editor, using purple and green as the site's custom colours, and Chalkboard and Impact for the site fonts:

image

When placed on the top, only the most general custom styles (paragraph and header) are applied; all colours, and some more specific fonts (like for buttons) are not:

image

I think we might either need to do the 'nuclear' option (adding a bunch of classes), or, ideally, simplify these selectors in the theme a bit 😕 I can start on the latter; the issue is just that they're going to be time-consuming to test, but I think it'll be best in the long run.

adekbadek:
Closing in favour of https://github.com/Automattic/newspack-theme/pull/1133

PR #1133, opened 2020-11-06
Simplify block selectors so styles to be applied outside of the content area

laurelfulford:
### All Submissions:

This PR is a replacement for #1131, to work with the new above header campaign placement coming in https://github.com/Automattic/newspack-popups/pull/292.

It greatly simplifies the block selectors in the themes and custom font and colour code, so it can be picked up outside of the content area. It also has the added benefit of making the styles shorter overall, and make them easier to override with Custom CSS.

As part of testing I created some before and after screenshots, as well as some test content for both the post body and campaign body, all available here: https://cloudup.com/ca1itJfrJH7 (edited to add: the light pink area in the screenshots is the above-header campaign). I think it's a good starting point for additional testing, though I'd appreciate any new blocks/setting that you all can dream up, so we're not all testing the exact same thing. The main issue I see with these changes is any regressions in how the blocks are styled in the normal content area.

This PR does not address the campaign content width; everything fills the width of the browser window, and the full and wide alignments aren't applied. We may want to restrict the content width in some way (like the same as the theme's 1200px wide header) -- open to feedback on that! 

1. Set up a test post with a lot of different blocks and settings (see the below list for the blocks that have changed and potential concerns).
2. Set up a test campaign with a few different blocks and settings -- it's probably safe to limit this one to things like text, buttons, cover blocks, anything that's more typical to promotional material.  
3. Navigate to Customizer > Colors and set distinct custom colours; make them darker so they won't be flipped to grey when used for text. 
4. Navigate to Customizer > Typography, and set custom fonts; check the option to make accent text all-caps.
4. Turn off AMP; if you've set up your blocks right, there will be too much CSS for it all to load with AMP enabled. 
5. Flip between the master branch and this PR, for each of the themes, on your test page. Confirm that a) styles are being picked up correctly in your campaign, and b) styles remain the same in the post body.

#### Blocks

The following blocks are affected by this update -- I've added a star next to the ones that are less important due to use/changes:

* Homepage posts block
   * With an accent header set
   * With and without text colour set
* Columns block
   * With and without borders
* Paragraph block
   * With and without background
   * With and without drop cap
* Button block
   * Using colours available from the palette
   * Using custom colours/gradients
   * Using outline style
   * Using outline style with different colours
* Pullquote block
   * With various settings: main colour (note: this is added inline and will not be picked up after the custom colours are changed), and solid background.
* Cover block
   * With link in content (to make sure it still picks up the white text colour)
* Group Block
* Quote Block
* Image Block
   * With a caption
* Gallery Block
   * With a caption
* Separator Block
* Table Block
* Audio Block*
* Video Block*
* Preformatted*
* Verse*
* Block File*
* Block Code*

##### Newspack Joseph
* Changed custom typography selectors for `wp-block-pullquote` -- it should inherit the header font.
* Changed the custom typography selector for `wp-block-button` -- unlike other themes, it should inherit the body font.
* Changed the selector for `wp-block-separator`; it should still apply the double-border style to the regular and wide style.

##### Newspack Katharine
* Simplified the custom colour selector for the `entry-meta` - need to make sure it's using the primary colour in the blocks and single posts
* Simplified the `wp-block-separator` selector -- need to make sure it's still using the dotted border.
* Changed selectors for the homepage posts block section title; need to make sure it's smaller, has a greater space underneath it, and is all caps. 

##### Newspack Nelson
* Changed selectors for the homepage posts block section title; need to make sure it's smaller, has a greater space underneath it, and is all caps. 
* Simplified the `wp-block-separator` selector -- need to make sure it's still bolder and darker

##### Newspack Sacha
* Simplified the custom colour selector for the `entry-meta` - need to make sure it's using the primary colour in the blocks and single posts
* In the custom typography, made the `blockquote`, `wp-block-quote` and `wp-block-pullquote` less specific; need to confirm that they're picking up the header font. 

### Other information:

### DIFF
--- a/newspack-joseph/inc/child-typography.php
+++ b/newspack-joseph/inc/child-typography.php
@@ -51,7 +51,7 @@ function newspack_joseph_custom_typography_css() {
 				font-family: ' . wp_kses( $font_header, null ) . ';
 			}
 			.block-editor-block-list__layout .block-editor-block-list__block .article-section-title,
-			.block-editor-block-list__layout .block-editor-block-list__block .accent-header {
+			.block-editor-block-list__layout .block-editor-block-list__block.accent-header {
 				font-family: inherit;
 			}';
 	}
@@ -76,7 +76,7 @@ function newspack_joseph_custom_typography_css() {
 			}
 		';
 		$editor_css_blocks .= '
-			.block-editor-block-list__layout .block-editor-block-list__block .accent-header,
+			.block-editor-block-list__layout .block-editor-block-list__block.accent-header,
 			.block-editor-block-list__layout .block-editor-block-list__block .article-section-title {
 				text-transform: uppercase;
 			}
--- a/newspack-joseph/inc/child-typography.php
+++ b/newspack-joseph/inc/child-typography.php
@@ -17,7 +17,7 @@ function newspack_joseph_custom_typography_css() {
 	if ( get_theme_mod( 'font_header', '' ) ) {
 		$css_blocks .= '
 			.has-drop-cap:not(:focus)::first-letter,
-			.entry .entry-content .wp-block-pullquote {
+			.wp-block-pullquote {
 				font-family: ' . wp_kses( $font_header, null ) . ';
 			}
 
@@ -35,7 +35,7 @@ function newspack_joseph_custom_typography_css() {
 			.widget,
 			.widget-title.accent-header,
 			.accent-header,
-			.entry .entry-content .wp-block-button__link,
+			.wp-block-button__link,
 			.entry .article-section-title,
 			button,
 			input[type="button"],
--- a/newspack-joseph/inc/child-typography.php
+++ b/newspack-joseph/inc/child-typography.php
@@ -35,7 +35,7 @@ function newspack_joseph_custom_typography_css() {
 			.widget,
 			.widget-title.accent-header,
 			.accent-header,
-			.wp-block-button__link,
+			.wp-block-button .wp-block-button__link,
 			.entry .article-section-title,
 			button,
 			input[type="button"],
--- a/newspack-joseph/sass/style-editor.scss
+++ b/newspack-joseph/sass/style-editor.scss
@@ -141,6 +141,7 @@ Newspack Sacha Editor Styles
 		flex: 1 0 0.25rem;
 		height: 1px;
 		margin: 0 0 0 0.25rem;
+		position: relative !important;
 	}
 }
 
--- a/newspack-joseph/sass/style.scss
+++ b/newspack-joseph/sass/style.scss
@@ -103,7 +103,7 @@ figcaption,
 .widget,
 .widget-title.accent-header,
 .accent-header,
-.entry .entry-content .wp-block-button .wp-block-button__link,
+.wp-block-button__link,
 .entry .article-section-title,
 button,
 input[type='button'],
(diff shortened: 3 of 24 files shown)


adekbadek:
I'm unclear on which differences are expected and which are regressions – many differences are visible on the images on cloudup, e.g.:

In "Newspack Default – Before":

image

In "Newspack Default – After":

image

However, in my local testing, the differences in these blocks (Group, Buttons) are different:

before:

image

after:

image

(the only difference is in the missing button background in "after")

Maybe the files in the cloudup uploads are mislabelled?

laurelfulford:
> I'm unclear on which differences are expected and which are regressions – many differences are visible on the images on cloudup

Apologies! I should have been more clear: the light pink area in the screenshots is an above header campaign (the header's pretty small so it's not super clear). I'll update the issue above to include that. 

The [button block styles](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/sass/blocks/_blocks.scss#L528) are too specific right now to be applied there, so they're showing the default -- the group block is missing the background for the same reason. With the styles updated, they pick up the right fonts/colours/border radiuses. 

I'll dig into the issue with the light grey background not being applied now!

adekbadek:
> the light pink area in the screenshots is an above header campaign 

Ah, that makes sense :) 

Here are the differences I've found in the blocks' frontend rendering, using the provided post markup (left is before (`master`), right is after):

1. Table block – borders missing

image

2. File download block – text color

image

3. Author name in Carousel block - dimmed on `master`

image

Also, the File download block has a blue color in the editor:

image

while on the frontend it has the custom primary color:

image

laurelfulford:
Thanks @adekbadek!

> Table block – borders missing

Whoops, good catch! This should be fixed in 7b972ff.

> File download block – text color

This was an accidental bug fix -- the button originally wasn't changing the text colour for maximum contrast, it was always white. It's now picking up the dark text colour when it considers it to have better contrast -- you just happen to be using a perfect middle shade that works well with either, but still is better with dark text :) 

> Author name in Carousel block - dimmed on master

Another accidental bug fix -- the theme styles were too specific and overriding the block (the text should always be white). So this is a good thing!

> Also, the File download block has a blue color in the editor:

Yikes, thanks! I made some updates in 618012e to make sure the button style and block size were both consistent across the editor and front-end -- they should be closer in appearance now!

adekbadek (approved):
🚢

laurelfulford:
Tagging 'Do Not Merge' for now, just to get a chance to test this on a couple of the more customized sites before merging :)

laurelfulford:
@jeffersonrabb and I did some more intense testing on content/styles from live sites, and found a few more issues with this PR:

1. Buttons with a light, non-custom text colour set (like like grey), do not use the correct colour - below is a non-visited link (using the secondary colour) and a visited link (using dark grey):

image
image

2. Buttons with the outline style, background colour and text colour are not picking up the correct background colour -- they're using the defaults -- they look like this with the primary colour assigned:

image

... when they should pick up the site's custom primary colour:

image

3. Cover blocks that use a different colour + dimness setting will not have the custom colour applied - below is a cover block that should be the light grey from the palette, but is using the default black instead. Dimness should be tested with gradients, too: 

image

4. Newspack Scott is missing the top border on Pullquote Blocks -- they look like:

image

... but they should look like:

image


And a couple notes when it's rolled out:

Currently for sites with Custom CSS, it is going to be too specific in some cases to be applied to the above header campaign position. For example, if a site is customizing the button block, it will have to had used the `.entry .entry-content` selector, so the exact styles being applied to their content area won't be applied to the campaign. This can be done after this is rolled out, as it won't be an issue until the blocks -- and if the blocks -- are added to that campaign spot. 

There is also one site that has a different approach to custom colours that will need a plan before this is rolled out, because right now those style override some of the default styles, when they didn't used to. One known issue is link colour, getting inherited by things like buttons (different than the above issue); there may be more.

laurelfulford:
The button, cover block and Newspack Scott pullquote issues should now be fixed; now this just needs https://github.com/Automattic/newspack-issues/issues/203 resolved before it can be merged.

PR #1008, opened 2020-07-31
Feat: Adding sponsored content visuals

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR adds visuals for the Sponsored Content plugin, including:

* A 'flag', logo, customizeable byline and bio on single posts
* A 'flag', logo and byline on posts in archive/search.
* A flag, logo, byline and bio on sponsored category and tag archives.
* Customizer options for changing the sponsor flag background colour.
* Styles for the Homepage Posts and Post Carousel blocks -- the sponsored visuals for those will be coming from the plugin itself ([from this PR](https://github.com/Automattic/newspack-blocks/pull/563)).
 
**Questions/notes**
* Right now the sponsored content (flags and all) will appear on author archive pages. I'm not sure if those specific posts should be hidden there, or if that should be manually managed on a per-site basis (eg. always assign sponsored content to a guest author used only for that purpose if it shouldn't appear in author archives).
* On a sponsored category, each post still displays the 'sponsored' flag. I'm waffling a bit between whether this is okay, or overkill. Would appreciate feedback!

Closes #1006

### How to test the changes in this Pull Request:

1. Install and activate the Newspack Sponsors plugin (ping myself or Derrick if unsure where to find it).
2. Switch your site to use AMP transitional so you can test both the AMP and non-AMP versions; the main issues to check with both are a) logo sizing (without distortion or excessive white space), and b) the 'Sponsor Information' toggle.
3. Apply this PR and run `npm run build`.
4. Add at least four sponsors; make sure to include the following combinations; make sure to use at least one horizontal and one vertically-oriented logo:
      * Set scope to native content, attached to a category
      * Set scope to native content, attached to a specific story
      * Set scope to underwritten content, attached to a category
      * Set scope to underwritten content, attached to a specific story.
      * Make sure at least one story will end up with more than one sponsor. 
      * Make sure each story uses a different featured image setting (specifically make sure to incorporate behind and beside)
[...]

PR #1144, opened 2020-11-12
feat: optimize sponsors queries; fix phpcs lint rules

### LINKED ISSUE #1127: Improve efficiency of sponsors queries
Sponsor queries are working correctly, but there's some opportunity for improving performance by reducing the number of redundant queries that are executed to get sponsor info. For example in https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/archive.php, there are a number of conditionals that look like this:

```php
if ( ( is_category() || is_tag() ) && function_exists( 'newspack_get_all_sponsors' ) && newspack_get_all_sponsors( get_queried_object_id() ) ) {
	newspack_sponsor_label( get_queried_object_id(), true, 'native', 'archive' );
}
```

Each time the `newspack_get_all_sponsors` function is executed, the Sponsors plugin executes a query. Then if the condition is true, `newspack_sponsor_label` executes another sponsors query. In many cases we should be able to refactor these templates to run the `newspack_get_all_sponsors` once to get all sponsor info, store the sponsor info in a variable, and use the stored data to handle other display logic.

This optimization should be done everywhere we display sponsor info (post archives, tag/category archives, single posts).

dkoo:
### All Submissions:

I opened [this issue](https://github.com/Automattic/newspack-theme/issues/1127) after the systems team informed us that the sponsors plugin was executing a large number of queries per page. This PR attempts to minimize the number of queries executed by the sponsors plugin to at most once per template part.

I also added a phpcs.xml config file which should update the linting behavior to more recent WP standards (now matches other Newspack repos).

If all looks good to you, I'll do the same updates to the Blocks repo.

Closes #1127.

Front-end behavior shouldn't be any different. Testing instructions for sponsors are [here](https://github.com/Automattic/newspack-theme/pull/1008).

### Other information:

### DIFF
--- a/newspack-theme/archive.php
+++ b/newspack-theme/archive.php
@@ -51,7 +51,7 @@ $underwriter_sponsors = newspack_get_underwriter_sponsors( $all_sponsors );
 				<?php
 				if ( ( is_category() || is_tag() ) && ! empty( $native_sponsors ) ) {
 					// Get label for native archive sponsors.
-					newspack_sponsor_label( $native_sponsors );
+					newspack_sponsor_label( $native_sponsors, null, true );
 				}
 				?>
 
--- a/newspack-theme/archive.php
+++ b/newspack-theme/archive.php
@@ -7,6 +7,11 @@
  * @package Newspack
  */
 get_header();
+
+// Get sponsors for this taxonomy archive.
+$all_sponsors         = newspack_get_all_sponsors( get_queried_object_id() );
+$native_sponsors      = newspack_get_native_sponsors( $all_sponsors );
+$underwriter_sponsors = newspack_get_underwriter_sponsors( $all_sponsors );
 ?>
 
 	<section id="primary" class="content-area">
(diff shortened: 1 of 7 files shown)


laurelfulford (changes requested):
Thanks for tackling this @dkoo !

In my testing, I found two changes from how it works now:

1. On the author archive page, posts with Native sponsors used to display the 'flag' above the post content -- before:

image

After:

image

It looks like the flags are being displayed as expected on category and tag archives

The only case where the flags were not set to display is when you're on a sponsored category's archive, since it's at the very top of the archive page:

image

2. On single posts, the 'flag' now appears on underwritten posts; it should be displaying the category like on non-sponsored posts:

image

This is not working right on `master` either -- neither the category or flag displays, which no doubt caused confusion! It should work like it does in the search results, with the category being left in place for posts with underwritten sponsors:

image

Everything else seems to be working as expected -- please let me know if you have any questions about the above!

dkoo:
> 1. On the author archive page, posts with Native sponsors used to display the 'flag' above the post content

Good catch! I had an extra conditional that doesn't apply for author archives. Fixed in f393125.

> On single posts, the 'flag' now appears on underwritten posts; it should be displaying the category like on non-sponsored posts

I believe this should fix it, can you please confirm? On posts with at least one native sponsor, we show the sponsor flag. On posts with no sponsors or only underwriter sponsors, we show the regular category flags: d841204

laurelfulford (approved):
Thanks Derrick! 

Confirmed that the flags are displaying correctly on the author archive page, and posts with only an underwritten sponsor display their category -- but if they have both a native and underwritten sponsor, they'll display the flag.

LGTM!

PR #1154, opened 2020-11-26
Feature: Add option to display 'updated' date

### LINKED ISSUE #598: Add Update date option in the article
An enhancement is requested.

Add Last updated option for posts and also the option to hide published date and only display updated date.

It will be of great use for news sites.

laurelfulford:
### All Submissions:

This PR adds an option to the customizer to display each post's last updated date, if it has been updated more than a full day after its published.

When paired with the 'time ago' date format, it will automatically override the 'cut off' time for the time ago format, and switch it to 1 day. This is to serve two purposes: 1. Make sure you don't end up with a somewhat unclear date, like "November 20th Updated 2 days ago", and 2. to avoid having both use the time ago format without too many code acrobatics (5 days ago Update 2 days ago").

**Questions** 

* Do we need some kind of override for changes that happen during migration/launch? I noticed this in the CalMatters code I was working from.
* Right now, the updated dates only display on single posts; would it make sense to add them to other locations, too, like the archives, search, or blocks (which would, in part, need a separate PR at least to add the Updated string)? 
* Does the 'Time ago' override behaviour make sense, or would it be better to honour their settings and let them decide if it seems weird or not?

See #598

2. Start with a mix of posts that have been published recently, updated recently, and published and updated again in the same 24 hour period. 
3. Navigate to Customizer > Template Settings > Post Settings and check the 'Show "Last updated" date' checkbox.
4. Confirm that on posts that have been published and updated more than 24 hours later, a second 'Updated' date is displaying:

image

5. Confirm that on posts that have been updated less than 24 hours after publishing, the updated date does not show up.
6. Navigate back to Customizer > Template Settings > Post settings, and check 'Use "time ago" date format'; note that the 'Cut off' field is greyed out.
7. Try toggling on and off the 'Show "last updated" date' to make sure the cut off field is enabled/disabled. Leave both options checked and click 'Publish'.
8. Confirm that the time-ago date is only applied to the updated date, even when the normal 'cut-off' would apply it to both dates:

image

9. Confirm that posts that don't display the updated date (either because they haven't been updated, or that were updated in the 24 hours after publishing), don't show the updated date, and show the 'time ago' format:

image

### Other information:

### DIFF
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -605,20 +605,14 @@ function newspack_should_display_updated_date() {
 		return false;
 	}
 
-	$post = get_post();
-
+	$post          = get_post();
 	$publish_date  = $post->post_date;
 	$modified_date = $post->post_modified;
 
-	// Exclude posts updated during launch.
-	// if ( 0 === strpos( $modified_date, '2020-06-23' ) ) {
-	// return false;
-	// }
-
 	$publish_timestamp  = strtotime( $publish_date );
 	$modified_timestamp = strtotime( $modified_date );
+	$modified_cutoff    = strtotime( 'tomorrow midnight', $publish_timestamp );
 
-	$modified_cutoff = strtotime( 'tomorrow midnight', $publish_timestamp );
 	if ( $modified_timestamp > $modified_cutoff ) {
 		return true;
 	}
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -566,6 +566,7 @@ function newspack_math_to_time_ago( $post_time, $format, $post, $updated ) {
 		$cut_off_seconds = $cut_off * 86400;
 
 		if ( true === newspack_should_display_updated_date() ) {
+			// Switch cut off to one day if updated date is displayed.
 			$cut_off_seconds = 86400;
 		}
 
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -565,8 +565,8 @@ function newspack_math_to_time_ago( $post_time, $format, $post, $updated ) {
 		// Transform cut off from days to seconds.
 		$cut_off_seconds = $cut_off * 86400;
 
-		if ( true === newspack_should_display_updated_date() ) {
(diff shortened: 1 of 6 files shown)


SanjayKumarMonu:
One of the best and important feature needed by any news website.

Thanks @laurelfulford for bringing this 🙌

newspack-theme/inc/template-functions.php
```diff
@@ -562,6 +578,44 @@ function newspack_convert_to_time_ago( $post_time, $format, $post ) {
...
+	if ( is_single() && true === get_theme_mod( 'post_updated_date', false ) ) {
+		$post          = get_post();
+		$publish_date  = $post->post_date;
+		$modified_date = $post->post_modified;
+
+		$publish_timestamp  = strtotime( $publish_date );
+		$modified_timestamp = strtotime( $modified_date );
+		$modified_cutoff    = strtotime( 'tomorrow midnight', $publish_timestamp );
+
+		if ( $modified_timestamp > $modified_cutoff ) {
```
dkoo, line 614: Not a blocker at all, but this whole `if...else` block could be replaced by `return $modified_timestamp > $modified_cutoff;`.

dkoo (approved):
@laurelfulford Tested and works as described, so approving. Offering my opinions (just opinions) on the questions you posed:

> Do we need some kind of override for changes that happen during migration/launch? I noticed this in the CalMatters code I was working from.

Hmm, this does make sense to me, but I'm not sure what the best UI solution would be. Should be a per-post toggle (meta value)? Would it be on or off by default for new posts?

> Right now, the updated dates only display on single posts; would it make sense to add them to other locations, too, like the archives, search, or blocks (which would, in part, need a separate PR at least to add the Updated string)?

This seems like it would be a good option to add to the customizer, maybe?

> Does the 'Time ago' override behaviour make sense, or would it be better to honour their settings and let them decide if it seems weird or not?

What if we offered a second cut-off field that applies only to the "updated" time?

laurelfulford:
Thanks for the feedback, Derrick!

Going to merge this as is and move the suggestions into different issues, for implementing separately.

laurelfulford:
Documentation has been added here: https://newspack.pub/support/themes/date-formats/#last-updated

PR #1167, opened 2020-12-02
fix: correct issue with primary menu background control

laurelfulford:
### All Submissions:

Fixes issue where primary colour background control wasn't showing when expected.

1. Apply the PR and run`npm run build`
2. Try Header Settings > Appearance > Solid Background - toggle on and off the Header Background Colour and the primary background colour option should appear with the regular header background option.
3. Disable Header Settings > Appearance > Solid Background  - primary background colour should be gone
4. Enable Header Settings > Appearance > Solid Background - it should be back 
5. Enable Header Settings > Appearance > Short Header - it should be gone
6. Try toggling on and off Colors > Header Background Colour and make sure the primary menu option stays hidden. 
7. Turn Colors > Header Background Colour off. 
8. Turn Header Settings > Appearance > Short Header off.
9. Turn Colors > Header Background Colour on -- the primary menu option should be back. 

### Other information:

### DIFF
--- a/newspack-theme/js/src/customize-controls.js
+++ b/newspack-theme/js/src/customize-controls.js
@@ -153,24 +153,8 @@
 					if ( true === setting.get() ) {
 						if (
 							'custom' === wp.customize.value( 'header_color' )() &&
-							'custom' === wp.customize.value( 'theme_colors' )()
-						) {
-							control.container.slideDown( 180 );
-						}
-					} else {
-						control.container.slideUp( 180 );
-					}
-				};
-				visibility();
-				setting.bind( visibility );
-			} );
-
-			wp.customize.control( 'header_primary_menu_color_hex', function( control ) {
-				const visibility = function() {
-					if ( true === setting.get() ) {
-						if (
-							'custom' === wp.customize.value( 'header_color' )() &&
-							'custom' === wp.customize.value( 'theme_colors' )()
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
There are a lot of combinations here! I went through the steps and the only thing I saw that differed from what you specified was this:

> Enable Header Settings > Appearance > Short Header - it should be gone

With both "Solid Background" and "Short Header" options enabled in Header Settings > Appearance, the Header Background color option was still visible in Colors. This doesn't seem to be a problem to me, though, since it did change the short header color as expected.

laurelfulford:
> With both "Solid Background" and "Short Header" options enabled in Header Settings > Appearance, the Header Background color option was still visible in Colors.

We discussed a bit in Slack but just to close the loop here -- the menu background option should be the only one that's not visible in this case - the header background should still be there. Apologies for the confusing instructions for this very confusing control! 😬

dkoo (approved):
This explanation clears things up for me:

> It was the primary menu background that shouldn’t be there, not the header background itself — the primary menu background just shouldn’t display when it’s beside the logo, because it doesn’t have a different coloured background in that case.

Confirmed that this is the behavior. Looks good!

laurelfulford:
Thanks Derrick!

PR #1191, opened 2020-12-19
feat: add styles for the block 'read more' links

laurelfulford:
### All Submissions:

This PR adds some styles to the new 'Read More' link added to the Homepage Posts block. It needs to be tested with https://github.com/Automattic/newspack-blocks/pull/656 applied.

Edited to add: I'm still waffling a bit about how opinionated this is, so I'd definitely appreciate feedback on this either way; it's pretty subtle now, but could also be bolded, or have some kind of icon -- like an arrow -- to help it stand out more. 

1. Set your test site to use the default font and colours.
2. Apply this PR to the theme, and https://github.com/Automattic/newspack-blocks/pull/656 to the Newspack Blocks, and run `npm run build` for each. 
3. Add some homepage posts blocks with the Read More link enabled. 
4. View on the front-end and in the editor; confirm that the link is the same colour as the text, and doesn't have an underline unless hovered over. When the block type scale is 4 and higher, the link font size should be slightly smaller; for the type scale 1-3, the link should be the same font size:

image

image

5. Try changing the block text colour and confirm that the link picks it up:

image

6. Try changing the site's custom colours and fonts; confirm they're picked up as expected (with the link picking up the body font, and still the same colour as the text):

image

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -481,6 +481,7 @@ function newspack_custom_colors_css() {
 		}
 
 		/* Do not overwrite solid color pullquote or cover links */
+		.block-editor-block-list__layout .block-editor-block-list__block a.more-link,
 		.block-editor-block-list__layout .block-editor-block-list__block .has-text-color a,
 		.block-editor-block-list__layout .block-editor-block-list__block .has-text-color a:hover,
 		.block-editor-block-list__layout .block-editor-block-list__block .wp-block-pullquote.is-style-solid-color a,
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -179,6 +179,10 @@ p.has-background {
 		max-width: 100%;
 	}
 
+	.more-link + .entry-meta {
+		margin-top: #{0.75 * $size__spacing-unit};
+	}
+
 	article.format-aside {
 		.entry-meta > a,
 		.entry-meta .byline {
--- a/newspack-theme/sass/site/primary/_posts-and-pages.scss
+++ b/newspack-theme/sass/site/primary/_posts-and-pages.scss
@@ -192,19 +192,19 @@ amp-script .cat-links {
 	}
 
 	.more-link {
-		@include link-transition;
-		display: inline;
 		color: inherit;
+		font-size: 0.8em;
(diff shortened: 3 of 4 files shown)


laurelfulford:
Switching this PR to draft more to rework this a bit -- I think it's too opinionated, and after the benefit of some time away, the vertical spacing looks off. Should have an update shortly!

dkoo (changes requested):
Screen Shot 2021-01-05 at 11 14 53 AM

Code looks good to me, and I like the subtler style. My only question is, is it now *too* subtle? I find it hard to tell at a glance that these are links and not regular text since in many cases it looks identical to the excerpt when not in a hover state. If you think that's okay, I'll go ahead and approve, but it seems like a minor accessibility concern to me.

laurelfulford:
> My only question is, is it now too subtle?

😂 this honestly has been bugging me the entire break -- I don't disagree! 

Some other things I tried include:

* An arrow at the end, like "Keep Reading →", which I then started worrying was too "bloggy" or opinionated; you can also add an arrow to the actual text label of the button, so it seemed like "a lot" to put on all sites. 
* Making the link use a different font, but that made it seem to be part of the post meta rather than the excerpt (since it uses the header font -- it's less obvious with the settings on your test site, but when the body is serif and the headers is not, it feels... off). 
* Making it bolded, but that didn't totally read link-like, and competed visually with the post meta. 

I've pushed up an update that adds an underline in 990fc5e; I'd also be happy to re-try the arrow if that doesn't seem bad, or bolding? We're probably to be overriding this with Custom CSS on some sites no matter what, it's definitely better to make sure the default represents best practices!

I put together some screenshots to give better examples of what I'm talking about -- it's also very possible I'm overthinking these:

Underline (current):

image

Arrows:

image

Bold:

image

Header font + bolding:

image

dkoo (approved):
Oh boy, sounds like you've already considered all this and more. I'll go ahead and approve and let you decide which approach you want to go with. The underline looks fine to me, FWIW. 😁

laurelfulford:
Thanks @dkoo! Haha there's no point thinking about something if you can't _over_ think it! 

I think the underline is a good in-between option -- link-y and noticeable compared to how it looked before, but not too much, and easy to override should specific publishers want to see changes (it's just picking up the default link styles, at this point).

PR #1200, opened 2021-01-09
Add option to hide page title

### LINKED ISSUE #1048: Add option to hide page titles
Add an option to a theme to hide a page title on a specific page.

laurelfulford:
### All Submissions:

This PR adds an option to hide the page title on individual pages on the front-end.

Closes #1048.

2. Edit a page; confirm that you have the option to hide the title under Status & Visibility, and that it's off by default.

image

3. Enable 'Hide page title', and click 'Update'.
4. Confirm that the page title is not displaying: 

image

5. Toggle the Hide page title option back on, and confirm that the title displays again. 

### Other information:


newspack-theme/js/src/hide-page-title.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
```
dkoo, line 1: **Not a blocker:** most of this code is repeated from `post-date-updated.js`, right? What if we were to combine both files and nest both toggle controls under `PluginPostStatusInfo`? This would let you enqueue the file in both posts and pages, and manage all the `PluginPostStatusInfo` extensions in one place (in the JS, you would need to use the `getCurrentPostType` selector and conditionally show the options you want for posts vs. pages based on that). 

Totally not necessary if you prefer to keep each option separate, though.

dkoo (changes requested):
I've got one performance improvement suggestion which I think is a good idea to implement, and one organizational suggestion which is more of a suggestion but not a blocker if you don't like it.

Also wanted to confirm, did you mean to intentionally leave the `h1` title element in the markup, but hidden with CSS for SEO purposes? Because we could also just not show the title at all in the markup if the "hide title" option is enabled.

newspack-theme/functions.php
```diff
@@ -441,6 +441,11 @@ function newspack_enqueue_scripts() {
 			wp_set_script_translations( 'newspack-post-date-updated', 'newspack', get_parent_theme_file_path( '/languages' ) );
 		}
 	}
+
+	if ( 'page' === get_current_screen()->post_type ) {
```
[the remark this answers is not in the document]
laurelfulford: Thanks Derrick! I like this approach; I've updated this section in e42768e.

newspack-theme/js/src/hide-page-title.js
laurelfulford: I've made this update in e18a52e (along with some CSS tweaks); it's quite possible I've made it more convoluted than necessary, so just let me know if it could use some adjustments! :)

laurelfulford:
Thanks Derrick! This should be ready for another look through.

> Also wanted to confirm, did you mean to intentionally leave the h1 title element in the markup, but hidden with CSS for SEO purposes? Because we could also just not show the title at all in the markup if the "hide title" option is enabled.

I honestly wasn't really thinking when I did that! I think in most cases it's probably best to remove the title entirely (assuming it'll be replaced with whatever blocks are replacing the title), so I've made that update!

newspack-theme/js/src/hide-page-title.js
dkoo: It's hard to explain what I was thinking, so I pushed my suggestions related to the above in 060c81a. Feel free to revert if you don't want these changes! But this is what I was thinking:

* I like how you localized the post types that should show each toggle option. That makes it much more future proof. I turned those items into arrays so if we wanted, we could add more post types in the future.
* For the JS, I handled both toggles in a single sidebar plugin component. I think this way it's clear when each option should be displayed, and it also makes it easier if we have more custom options to add in the future.
* I tweaked some logic around the Updated date so that if `newspack_should_display_updated_date()` is false, we just don't render that second time string at all (instead of hiding it with CSS). This matches the new treatment of hiding the title.

Let me know your thoughts—if these changes look good and work for you, I'll approve!

dkoo (changes requested):
I like the new changes! Pushed some suggestions which aren't strictly blockers, so if you want to move ahead without them, we can revert those and merge.

laurelfulford:
Hey @dkoo! I think that looks great -- much more concise, and more "ready" for expanding if needed for different post types :) I didn't spot any issues in my testing with either the date or page title options 👍

dkoo (approved):

dkoo:
Thanks for testing! I'll give my approval but it's probably a good idea to have someone else look too. @adekbadek, can you review?

adekbadek (approved):
👌

laurelfulford:
Thanks Adam!

PR #1203, opened 2021-01-12
Fix editor width issue

adekbadek:
### All Submissions:

Partial fix of https://github.com/Automattic/newspack-blocks/issues/666. This fix will make the sidebar less clipped, but there is still clipping caused by the editor itself, reported here: https://github.com/WordPress/gutenberg/issues/28140

1. Observe https://github.com/Automattic/newspack-blocks/issues/666 is not reproducible

### Other information:

### DIFF
--- a/newspack-theme/sass/style-editor-base.scss
+++ b/newspack-theme/sass/style-editor-base.scss
@@ -8,12 +8,6 @@ body {
 	background-color: $color__background-body;
 }
 
-/** === Content Width === */
-
-.wp-block {
-	max-width: 810px; // 780px + 30px to offset padding
-}
-
 /** === Base Typography === */
 
 // stylelint-disable-next-line no-duplicate-selectors


laurelfulford (changes requested):
Unfortunately it looks like removing the `max-width` changes how wide the blocks preview in the editor.

Here is the one-column template before:

image

And after -- rather than using the theme's max-width of 810px, it's falling back to the editor's default max-width of 580px:

image

This update doesn't affect the 'wide' template, which has a max-width set [here](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/sass/style-editor-overrides.scss#L36), but if we remove that we get the same kind of preview issue:

With max-width: 

image

Without max-width:

image

**Note:** I was only able to reproduce this issue locally by deactivating Jetpack, which seems like a red herring because it's active on other sites where people are seeing this issue. So if the above doesn't mesh with anyone else's experience, it could have to do with that. 

It also only seems to affect some blocks -- heading blocks, for example, treat the max-width like the actual width, but for me the columns block seems to scale down as expected.

adekbadek:
Connecting Jetpack seems to solve the issue. It also appears that the issue in the editor, no our code (https://github.com/WordPress/gutenberg/issues/28140, https://github.com/WordPress/gutenberg/issues/27622), so I'm going to close this now.

PR #1205, opened 2021-01-12
Comments: Add optional space for a comment policy 

### LINKED ISSUE #1102: Commenting policy option
Some publishers have a commenting policy or rules for commenters to follow to help foster a respectful community of commenters and help make engaging a positive experience. To do this, the rules/policy need to be clear to commenters.

A space at the top of the commenting section where the publisher can add a notice, of their choosing, that can contain a link to their commenting policy/rules.

There doesn't appear to be a suitable plugin out there that will enable this. I considered the WP Comments Fields Manager plugin but it appears to be broken and failing to generate it's output, as well as not being updated in a while.

**Additional context**
We have one publisher looking for this, but there have been noises along these lines as folks try to better manage their online community.

laurelfulford:
### All Submissions:

This PR adds the ability to add a comment policy above the WordPress comment form. 

Closes #1102

2. Navigate to Customize > Comment Settings
3. Confirm that you now have a checkbox to display the Comment Policy, and a text field below that for the actual policy text: 
4. 
image

4. Enable the checkbox, and add some text to the Comment Policy field, including some simple HTML (like bolding and links).
5. Click Update.
6. Navigate to a post with comments open, and confirm that your text is appearing between the Comment Form title and fields:

image

7. Navigate back to the customizer and uncheck the option to display the comment policy.
8. Confirm that the comment policy is now hidden. 

### Other information:

### DIFF
--- a/newspack-theme/inc/template-tags.php
+++ b/newspack-theme/inc/template-tags.php
@@ -380,9 +380,10 @@ if ( ! function_exists( 'newspack_comment_form' ) ) :
 			);
 
 			$comment_policy = get_theme_mod( 'comment_policy', '' );
+			$display_policy = get_theme_mod( 'display_comment_policy', false );
 
 			// Check if there's a comment policy set in the Customizer.
-			if ( '' !== $comment_policy ) {
+			if ( $display_policy && '' !== $comment_policy ) {
 				$comment_attributes['title_reply_before'] = '<span class="comment-policy">' . wp_kses_post( $comment_policy ) . '</span>';
 			}
 
--- a/newspack-theme/inc/template-tags.php
+++ b/newspack-theme/inc/template-tags.php
@@ -384,7 +384,7 @@ if ( ! function_exists( 'newspack_comment_form' ) ) :
 
 			// Check if there's a comment policy set in the Customizer.
 			if ( $display_policy && '' !== $comment_policy ) {
-				$comment_attributes['title_reply_before'] = '<span class="comment-policy">' . wp_kses_post( $comment_policy ) . '</span>';
+				$comment_attributes['title_reply_before'] = '<div class="comment-policy">' . wp_kses_post( $comment_policy ) . '</div>';
 			}
 
 			comment_form( $comment_attributes );
--- a/newspack-theme/inc/template-tags.php
+++ b/newspack-theme/inc/template-tags.php
@@ -383,7 +383,7 @@ if ( ! function_exists( 'newspack_comment_form' ) ) :
 			$display_policy = get_theme_mod( 'display_comment_policy', false );
 
 			// Check if there's a comment policy set in the Customizer.
-			if ( $display_policy && '' !== $comment_policy ) {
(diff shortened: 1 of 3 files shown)


newspack-theme/inc/template-tags.php
```diff
@@ -374,12 +374,20 @@ function newspack_post_thumbnail() {
...
+			$comment_attributes = array(
+				'logged_in_as' => null,
+				'title_reply'  => null,
 			);
+
+			$comment_policy = get_theme_mod( 'comment_policy', '' );
+			$display_policy = get_theme_mod( 'display_comment_policy', false );
+
+			// Check if there's a comment policy set in the Customizer.
+			if ( $display_policy && '' !== $comment_policy ) {
```
dkoo, line 386: This will guard against if there's only whitespace in the Comment Policy field:

```suggestion
			if ( $display_policy && '' !== trim( $comment_policy ) ) {
```

newspack-theme/inc/template-tags.php
```diff
@@ -374,12 +374,20 @@ function newspack_post_thumbnail() {
...
+				$comment_attributes['title_reply_before'] = '<div class="comment-policy">' . wp_kses_post( $comment_policy ) . '</div>';
```
dkoo, line 387: This will guard against invalid HTML being entered into the Comment Policy field (source: I entered some by accident 😅):
```suggestion
				$comment_attributes['title_reply_before'] = '<div class="comment-policy">' . wp_kses_post( force_balance_tags( $comment_policy ) ) . '</div>';
```

dkoo (changes requested):
This works great, I just have two small suggestions to guard against bad user input.

newspack-theme/inc/template-tags.php
dkoo: Actually, I think this would be better implemented as part of a custom sanitize callback [here](https://github.com/Automattic/newspack-theme/pull/1205/files#diff-83cc0878126972a709ebefafd5bd74d757c79f98b7309fcc0bbe8845678dc656R954), where you could create a function that runs the input through `wp_kses_post( force_balance_tags( $input ) )`. This way `force_balance_tags` doesn't have to be run on every front-end page load. What do you think?

newspack-theme/inc/template-tags.php
laurelfulford: Oh good call! This has been added in c5829e2.

newspack-theme/inc/template-tags.php
laurelfulford: Great catch! I think you just saved me from troubleshooting exploded HTML in the future! I went with the custom sanitation function, added in c5829e2.

laurelfulford:
Thanks @dkoo! I've updated the PR with your feedback - just let me know if it needs anything else!

dkoo (approved):
Looks great—thanks for the updates!

laurelfulford:
Thanks @dkoo!

laurelfulford:
Documentation added here: https://newspack.pub/support/themes/comments/#comment-policy

PR #1208, opened 2021-01-13
feat: add a sticky ad slot to the Advertising wizard

### LINKED ISSUE #1199: Ads: Add a new "sticky ad" global placement
Some publishers want to be able to run sticky ads as they have very high viewability and as such can be very lucrative. At the moment, other than using an unsupported plugin to add arbitrary Javascript, there's no way to add stick ads to a Newspack site.

Under Newspack > Advertising > Global Settings there should be a new pre-defined ad placement. This would work in the same was as other ad placements, allowing the selection of an ad provider/unit to be inserted into the spot.

The advert should be rendered as a sticky element at the bottom of the screen, and show in all devices widths.

The theme may want to provide default styles for the sticky ad (such as the close button).

**Additional context**
AMP [explicitly supports sticky ads](https://amp.dev/documentation/components/amp-sticky-ad/), so there should be no concerns with AMP compatibility

dkoo:
### All Submissions:

Make sure to test alongside https://github.com/Automattic/newspack-plugin/pull/812.

Closes #1199.

Follow testing instructions in https://github.com/Automattic/newspack-plugin/pull/812.

### Other information:

### DIFF
--- a/newspack-theme/sass/plugins/newspack-ads.scss
+++ b/newspack-theme/sass/plugins/newspack-ads.scss
@@ -20,12 +20,14 @@
 		box-shadow: 0 0 5px 0 rgba( 0, 0, 0, 0.2 );
 		bottom: 0;
 		display: none;
+		max-height: 100px;
+		overflow-y: hidden;
 		position: fixed;
 		width: 100%;
 		z-index: 1;
 
 		&.active {
-			display: block;
+			display: flex;
 		}
 
 		button {
@@ -58,6 +60,10 @@
 .newspack_amp_sticky_ad {
 	* {
 		margin: auto;
+	}
+
+	.amp-sticky-ad-top-padding {
+		display: none;
 	}
 
 	button,
--- a/newspack-theme/sass/plugins/newspack-ads.scss
+++ b/newspack-theme/sass/plugins/newspack-ads.scss
@@ -19,9 +19,14 @@
 		background-color: $color__background-body;
 		box-shadow: 0 0 5px 0 rgba( 0, 0, 0, 0.2 );
 		bottom: 0;
+		display: none;
 		position: fixed;
 		width: 100%;
 		z-index: 1;
+
+		&.active {
+			display: block;
+		}
 
 		button {
 			align-items: center;
--- a/newspack-theme/sass/plugins/newspack-ads.scss
+++ b/newspack-theme/sass/plugins/newspack-ads.scss
@@ -14,6 +14,67 @@
 	&.global_above_header {
 		background-color: $color__background-pre;
 	}
+
+	&.sticky {
+		background-color: $color__background-body;
+		box-shadow: 0 0 5px 0 rgba( 0, 0, 0, 0.2 );
+		bottom: 0;
+		position: fixed;
+		width: 100%;
+		z-index: 1;
+
+		button {
+			align-items: center;
+			background-color: $color__background-body;
+			border-radius: 0.75rem 0 0 0;
+			box-shadow: 0 -1px 1px 0 rgba( 0, 0, 0, 0.2 );
+			color: $color__text-main;
+			display: flex;
+			font-size: 1.25rem;
+			height: 1.75rem;
+			justify-content: center;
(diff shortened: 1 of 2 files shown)


laurelfulford (changes requested):
Overall this is looking good, @dkoo, and working as described!

In my testing, I spotted a couple inconsistencies between the AMP and non-AMP versions - nothing major though:

**When AMP is disabled:**

* The sticky ad location is scaled up to fit taller ads:

image

* The ads are not centered: 

image

**When AMP is enabled:**

* The sticky space is not scaled up to fit taller ads (I don't thing we want to encourage very tall ads in this location, but I'm not sure what the intended behaviour should be):

image

* There is a bit of space above the ad, that doesn't appear when AMP is disabled, and seems unbalanced, since it's only at the top. It's coming from a separate element, `amp-sticky-ad-top-padding`, so might be mandatory, I'm not sure. If yes, it might be nice to add it to the bottom, just to balance it out a bit. 

image

Otherwise, things worked as expected, and displayed consistently across different browsers.

dkoo:
> The sticky space is not scaled up to fit taller ads (I don't thing we want to encourage very tall ads in this location, but I'm not sure what the intended behaviour should be):

I can answer this one! `amp-sticky-ad` has a max-height of 100px, per the [specs](https://amp.dev/documentation/components/amp-sticky-ad/). If we want to maintain parity across AMP vs. non-AMP, we can enforce this max height for non-AMP, too—should we?

laurelfulford:
> If we want to maintain parity across AMP vs. non-AMP, we can enforce this max height for non-AMP, too—should we?

Ah, that makes sense! 

My vote would be to add a max-height for the non-AMP version too -- that way they'll behave consistently, and we can use it as a way to discourage adding a huuuuge ad there, too :)

dkoo:
> The ads are not centered

Doh, the container needs `display: flex`, not `block`. Fixed in 644f93b.

> There is a bit of space above the ad, that doesn't appear when AMP is disabled, and seems unbalanced, since it's only at the top. It's coming from a separate element, amp-sticky-ad-top-padding, so might be mandatory, I'm not sure. If yes, it might be nice to add it to the bottom, just to balance it out a bit.

This is coming from the `amp-sticky-ad` element, but it looks like we can hide it using CSS. I've removed it so there's no padding either above or below the creative in both AMP and non-AMP. (Also 644f93b)

> My vote would be to add a max-height for the non-AMP version too -- that way they'll behave consistently, and we can use it as a way to discourage adding a huuuuge ad there, too :)

Done! 644f93b again.

newspack-theme/sass/plugins/newspack-ads.scss
```diff
@@ -14,6 +14,78 @@
...
+			top: -1.75rem;
+			width: 1.75rem;
+
+			&::before {
+				content: '×';
+			}
+		}
+
+		> * {
+			margin: 0;
```
laurelfulford, line 55: This is getting into super nitpicky territory since this space is not intended for tall ads, but the max-height on the non-AMP ads is making it possible to create an ad that can never be dismissed -- it's pushing the X out of view as well as keeping the ad contained:

image

If you remove the `overflow-y` and `max-height` from `.sticky` and move the max-height here, it makes sure the too-tall add is contained without pushing the X out of sight: 

```suggestion
			margin: 0;
			max-height: 100px;
```

image

laurelfulford (changes requested):
Thanks @dkoo! Everything looks good -- just added one more nitpicky thing around the too-tall ads behaviour. If it's not as easy to fix as I assumed, I don't think it's a blocker -- just a weird thing that would happen if you did accidentally assign a tall ad to this spot :)

newspack-theme/sass/plugins/newspack-ads.scss
dkoo: Ah, good suggestion—this works as you describe. Updated in 87af381.

laurelfulford (approved):
Thanks Derrick! Looks good 🚢

PR #1224, opened 2021-01-26
Archives: make hidden categories on archives more flexible

laurelfulford:
### All Submissions:

Right now, in the archive templates, the post categories aren't output at all. This PR switches the categories to be output, but hidden in the CSS instead; this is to allow them to be made visible with flexibility (being able to target all archives; post type archives or specific types of archives, like author or tag), without adding multiple toggles.

The categories were originally hidden on the archive pages because they didn't make sense on the category archives; however, there have been really specific cases -- like Reveal's podcast CPT archive -- where it would be handy to show them.  A styles-only approach seemed to be the most flexible -- especially as we get into some site-specific CPTs -- but I'm open to feedback on other approaches!

1. Start with a site with a few sponsored posts. 
3. View an archive page; note that the sponsor labels are still showing for the appropriate posts, but the categories are not. 
4. Edit the custom CSS and add the following (or some variation of it, targeting a specific archive type):

```
.archive .cat-links:not(.sponsor-label) {
    display: block;
}
```

5. Confirm that the categories are now displaying on the appropriate archive page. 

### Other information:

### DIFF
--- a/newspack-joseph/sass/style.scss
+++ b/newspack-joseph/sass/style.scss
@@ -193,7 +193,8 @@ input[type='submit'] {
 
 /* Posts & Pages */
 .search,
-.blog {
+.blog,
+.archive {
 	.cat-links::before {
 		display: none;
 	}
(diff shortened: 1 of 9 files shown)


dkoo (approved):
Looks good 👍

laurelfulford:
Thanks!

Issue #1247, opened 2021-02-17
How to hide post published Date


vinvitthal:
**Expected behavior**
How to Hide published date in WP posts? Please add 'Date Visible or Not visible' option in theme customisation. 

Old Published Date issue is related to SEO, Google ranking is based on the date for the page, Updated post date not showing in Google SERP.

laurelfulford:
Hi @vinvitthal! 

If you're using Jetpack, there is an option to hide the date on posts under Customizer > Content Options. It's part of the Newspack theme's support for [Jetpack's Content Options](https://jetpack.com/support/content-options/).

If you're not using Jetpack, you can hide the date using the Additional CSS field in the Customizer - the following will hide the date on single posts, archives, and search results:

```
.entry-meta span.posted-on {
	display: none;
}
```

If the Newspack theme you're using stacks the author name and date on single posts, adding this bit of CSS as well will help vertically center the author name with the avatar, once the date is gone:

```
.entry-header .entry-meta {
	display: flex;
	align-items: center;
}
```

If you wanted to show only the updated date, there is an option in the theme for `Show "last updated" date on single posts`, under Customizer > Template Settings > Post Settings; by default, it will show both the original published date, and the updated date, prefaced by the label 'Updated'. If you only wanted to show the updated date in these cases, you could turn that on, and use the Additional CSS field to hide the original published date using this CSS:

```
.show-updated .entry-date.published,
.show-updated .updated-label {
	display: none;
}
```
I found a bug when looking into this issue: originally the updated date was always output in the theme's HTML, but hidden; when we added the option to show the updated date on single posts, this stopped working. I'm working on a fix for that now; once it's in place, you should also be able to use CSS to show the updated date everywhere with something like this:

```
.posted-on .updated:not(.published) {
	display: inherit;
}
```

...or hide the published date and only show the updated date, without the `Show "last updated" date on single posts` option turned on:

```
.entry-date.published:not(.updated) {
	display: none;
}
.posted-on .updated:not(.published) {
	display: inherit;
}
```

Since we do have an option in the Customizer to hide the date when Jetpack is enabled, and a few work arounds with Custom CSS to hide the date if you'd rather not use Jetpack, I'm going to close this issue. But if you have any questions about the above, please let me know!

vinvitthal:
Thanks for your support

And *How to Hide Author Avatar image in post??*

laurelfulford:
[...]

PR #1248, opened 2021-02-17
Make sure the updated date is still output in the HTML

laurelfulford:
### All Submissions:

When the option was added to 'Show updated date' on single posts in #1154, the theme's original behaviour -- always outputting and hiding the updated date -- stopped working. 

This is because the check for whether the published date matched the modified date got moved into the `newspack_should_display_updated_date()` check, so it was only run when the post would display the updated date when `Show "last updated" date on single posts` was enabled. 

This PR makes sure that functionality still works, while also allowing the 'updated date' option.

This came up when digging into #1247

1. Make sure `Show "last updated" date on single posts` is turned off under Customizer > Template Settings > Post Settings.
2. Update an older post, and publish a brand new post.
3. Inspect the published dates for both; you should just have one `<time>` in the source in both cases:

image

4. Apply the PR.
5. Confirm that the new post still has one `<time>` in the source, while the updated one now has two:

image

6. Confirm that this is the case wherever the posts appear from the theme -- a single post, archives, search results.
7. Navigate to Customizer > Template Settings > Post Settings and enable  `Show "last updated" date on single posts`.
8. Confirm that your updated date is now displaying on single posts. 
9. Confirm that nothing funky happened to the date in the archives or search, which should not have changed. 

### Other information:

### DIFF
--- a/newspack-theme/inc/template-tags.php
+++ b/newspack-theme/inc/template-tags.php
@@ -11,11 +11,11 @@ if ( ! function_exists( 'newspack_posted_on' ) ) :
 	 */
 	function newspack_posted_on() {
 		$time_string = '<time class="entry-date published updated" datetime="%1$s">%2$s</time>';
+		if ( get_the_time( 'U' ) !== get_the_modified_time( 'U' ) ) {
+			$time_string = '<time class="entry-date published" datetime="%1$s">%2$s</time>%3$s<time class="updated" datetime="%4$s">%5$s</time>';
+		}
 
 		if ( newspack_should_display_updated_date() ) {
-			if ( get_the_time( 'U' ) !== get_the_modified_time( 'U' ) ) {
-				$time_string = '<time class="entry-date published" datetime="%1$s">%2$s</time>%3$s<time class="updated" datetime="%4$s">%5$s</time>';
-			}
 			add_filter( 'get_the_modified_date', 'newspack_convert_modified_to_time_ago', 10, 3 );
 
 			$time_string = sprintf(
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Sorry I changed the logic back in #1154—didn't realize that both time strings were still needed, even if not shown!

This works as described except:

> Confirm that the new post still has one <time> in the source, while the updated one now has two

Both the brand new post and the updated post had two `time` elements (the brand new one with a modified time that was just a few seconds after the publish time—maybe from an auto-save?), but this didn't affect how things looked or behaved either with or without the `Show "last updated" date on single posts` option enabled.

laurelfulford:
> Sorry I changed the logic back in #1154—didn't realize that both time strings were still needed, even if not shown!

No worries! I should've caught that :) 

> Both the brand new post and the updated post had two time elements (the brand new one with a modified time that was just a few seconds after the publish time—maybe from an auto-save?), but this didn't affect how things looked or behaved either with or without the Show "last updated" date on single posts option enabled.

I can recreate this too -- first I tried creating two posts back to back, one really quickly, and the second allowing it to autosave while I messed around. At first both only had one date... but when I went to the archives and back, they both had two -- one example was 14 seconds apart:

```
Published: 2021-02-17T11:01:20-08:00
Modified:  2021-02-17T11:01:34-08:00
```

I double-checked this against the most-recent default theme that handles the updated date this way (Twenty Nineteen), and I get the same result for the same posts. So maybe it's some kind of delay in the full publishing process? 

This PR does restore it to the original behaviour -- and matches Twenty Nineteen -- so I'm going to go ahead and merge it, but if it does turn out to be an issue, we could try switching the `if ( get_the_time( 'U' ) !== get_the_modified_time( 'U' ) )` check to use a different date format that doesn't have seconds :)

PR #1250, opened 2021-02-17
Switch Donate block to use the header font

### LINKED ISSUE #1174: Donate Block: Use header font
Typically, things like buttons use the theme's header font, but the Donate block is the outlier: 

image

I think it'd look better if the donation amounts, and tabs, used the header font instead of the body font, which is often serif.

laurelfulford:
### All Submissions:

This PR switches the bulk of the Donate block to use the header font (either set by the theme or the custom fonts). With how the header font is otherwise used, these more button-like elements (tabs and toggles) make more sense styled like other buttons in the theme than not. 

Closes #1174

1. Navigate to Customize > Typography and unset any custom fonts.
2. Add a donate block to a post or page; note that it uses the body font for all but the button.

image

4. Confirm that the tabs and dollar amounts are now using the header font (the 'thank you message should still be the body font):

image

5. Edit the post and confirm that the block is using the correct fonts there as well.
6. Navigate to Customize > Typography and set a custom header and body font.
7. Verify that your custom fonts are being applied to the Donate block correctly, both in the editor and on the front end - the header font should be applied to all but the thank you message, which should use the body font:

image

### Other information:

### DIFF
--- a/newspack-theme/inc/typography.php
+++ b/newspack-theme/inc/typography.php
@@ -111,6 +111,8 @@ function newspack_custom_typography_css() {
 		.wp-block-latest-posts time,
 		.wp-block-file,
 		.wp-block-file .wp-block-file__button,
+		.wpbnbd .freq-label,
+		.wpbnbd .tier-label,
 
 		/* _widgets.scss */
 		.widget,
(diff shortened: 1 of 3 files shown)


dkoo (approved):
🚢

laurelfulford:
Thanks Derrick! 🙌

PR #1251, opened 2021-02-18
WooCommerce: make sure WC pages are picking up custom fonts/colour

laurelfulford:
### All Submissions:

This PR removes the Sass variables being set in the WooCommerce.scss file, and moves the typography selectors to the main theme SCSS files instead, to help the WooCommerce pages pick up the correct colours/fonts. As part of that, it removes a function that should have been applying the theme's custom colours, but it wasn't really working.

It also corrects a `.button` hover colour issue that was affecting all of the buttons on the WooCommerce pages (once a link was visited, it hovered dark text on a dark background, instead of white text on a dark background). 

**Future improvement:** Right now, the CSS generated for each child theme uses a combination of the parent theme's Sass, with some variables overwritten in each child theme (ones you can technically override include fonts face and style, colours, spacing, though we just do font face). The WooCommerce file is built a bit differently, and only uses the Sass variables from the parent theme. This is not a big deal right now -- only the colour variables are used in the WooCommerce Sass file, and they aren't changed from child theme to child theme yet, but it would be good to get this working if that changed down the road. The WooCommerce pages should still pick up the custom colours set in the Customizer.  

1. On a test site using custom fonts and colours, click through the WooCommerce purchase flow (shop -> product -> adding to cart -> checking out); while doing that, note the spots where the pages use the default fonts or colours. Some examples include:
* **Shop:** breadcrumb, title, prices use the wrong font; 'Sale' tags use the wrong font and colour
* **Product:** breadcrumb, price, tabs, description use the wrong fonts; 'Sale' tags use the wrong font and colour
* **Checkout:** coupon message uses the wrong fonts
2. Apply the PR and run `npm run build`
3. Confirm that the above pages are now using your custom fonts and colours

**Shop - before:**

image

**Shop - after:**

image

**Product - before**

image

**Product - after**

image

**Checkout - before**

image

**Checkout - after**

image

4. Try removing your custom fonts and colours, and pick a child theme (anything but Newspack default); confirm that the child theme's correct fonts (and not the default Newspack's theme's fonts). Good ones to check are Newspack Nelson, Katharine or Joseph, because they don't use the same body font (Georgia). 

### Other information:


dkoo (changes requested):
The colors and typography settings are picked up as described on all the WC pages I checked. I did notice two tiny unrelated things that we could fix here too:

1. On the product page, if you show Related Products, the section could use a `clear: both` to clear the float:

Screen Shot 2021-02-18 at 11 13 32 AM

The element to be cleared is `<section class="related products">` (note two separate class names).

2. On the checkout page, the × in the remove button is ever-so-slightly off-center:

Screen Shot 2021-02-18 at 11 15 00 AM

laurelfulford:
Thanks @dkoo! 

There is a LOT to be fixed on these pages, and I didn't explain that very well in this PR :) I'm trying to address the things listed in [these issues](https://github.com/Automattic/newspack-theme/issues?q=is%3Aissue+is%3Aopen+label%3AWooCommerce) one by one, but opted to start with a couple blanket PRs (colours & fonts, then tables), before digging into the per-page messes. 

Let me know if that doesn't make sense as an approach, review-wise! I was mostly trying to avoid creating a giant PR, but that might be easier than carefully reviewing XYZ, while trying to ignore the jankiness of ABC.

dkoo (approved):
You're right, my mistake! I saw the roadmap you've laid out but didn't put two and two together. With that understood now, I'll approve this so you can move ahead with the other items.

laurelfulford:
No worried, Derrick! 😄 It would have helped if I added some of that context to this PR, too! 

Thanks for the speedy review, as always!

PR #1253, opened 2021-02-19
fix: only show sticky ad at mobile viewports

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-plugin/pull/873, ensures that sticky ads only get shown at mobile viewports (less than 600px). Since the `amp-sticky-ad` element doesn't currently support responsive units, only the smallest sizes can get shown, which looks weird at larger viewports.

1. Check out this branch and run `npm run build`. Also make sure your main plugin is on https://github.com/Automattic/newspack-plugin/pull/873.
2. Set up an ad unit via GAM and set it to the Sticky predefined placement.
3. View the site front-end. Confirm that the sticky ad is shown only when the viewport is smaller than 600px.

### Other information:

### DIFF
--- a/newspack-theme/sass/plugins/newspack-ads.scss
+++ b/newspack-theme/sass/plugins/newspack-ads.scss
@@ -57,6 +57,12 @@
 }
 
 .newspack_amp_sticky_ad {
+	&__container {
+		@include media( mobile ) {
+			display: none;
+		}
+	}
+
 	* {
 		margin: auto;
 	}


newspack-theme/sass/plugins/newspack-ads.scss
```diff
@@ -57,6 +57,12 @@
 }
 
 .newspack_amp_sticky_ad {
+	&__container {
```
claudiulodro, line 60: This is cool. I didn't realize you can combine selectors like this!

claudiulodro (approved):

PR #1302, opened 2021-03-13
feat: add option to set footer logo size

### LINKED ISSUE #1218: Footer Logo: Add logo sizer for the footer logo
Right now, the site logo will be displayed in the footer, and there's an option to upload a second version of the logo for there, in case it needs to contrast against a different colour.

Unlike the header logo, the footer logo always displays at one fixed maximum width and height, which doesn't always work with all logo shapes. It would be helpful to have a similar resizing slider for this logo, that would change the logo size there independent of the header, regardless if it uses the same logo file as the header, or its own file.

laurelfulford:
### All Submissions:

Adds an option to switch the footer logo to use a larger size.

Note: The footer logo currently only displays if you have footer widgets assigned. This is probably something we want to untangle separately, to make the design wizard less confusing. 

Closes #1218

2. If you haven't already, navigate to Cusotmizer > Widgets > Footer and add at least one footer widget.
3. Navigate to Customizer > Site Identity, and confirm that you have a dropdown underneath the Footer logo setting, that defaults to 'Small':

image

4. Try changing the logo size to large -- this should change the logo max width and height from 200px x 75px, to 350px x 150px on screen sizes larger than tablet. 
5. Confirm that your logo is now sized up:

**Small:**

image

**Large:**

image

6. If you have a Footer Logo file assigned, unassign it so it falls back to the main logo. (If you don't, assign a Footer Logo, so you can test the opposite of what you've already tested).
7. Test both the Small and Large footer logo sizes, and confirm they still work. 

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -65,7 +65,7 @@ if ( ! function_exists( 'newspack_setup' ) ) :
 		add_image_size( 'newspack-featured-image', 1200, 9999 );
 		add_image_size( 'newspack-archive-image', 800, 600, true );
 		add_image_size( 'newspack-archive-image-large', 1200, 900, true );
-		add_image_size( 'newspack-footer-logo', 400, 9999, true );
+		add_image_size( 'newspack-footer-logo', 400, 9999 );
 
 		// This theme uses wp_nav_menu() in two locations.
 		register_nav_menus(
--- a/newspack-theme/inc/customizer.php
+++ b/newspack-theme/inc/customizer.php
@@ -584,7 +584,7 @@ function newspack_customize_register( $wp_customize ) {
 			$wp_customize,
 			'newspack_footer_logo',
 			array(
-				'label'       => esc_html__( 'Footer logo', 'newspack' ),
+				'label'       => esc_html__( 'Footer Logo', 'newspack' ),
 				'description' => esc_html__( 'Optional alternative logo to be displayed in the footer.', 'newspack' ),
 				'section'     => 'title_tagline',
 				'settings'    => 'newspack_footer_logo',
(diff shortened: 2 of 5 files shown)


thomasguillot:
Works well. I added 2 more sizes in https://github.com/Automattic/newspack-theme/pull/1302/commits/81589f5de1e6da35bb956b83235db0fc7dbf0d4a

I've also changed the sizes a bit, to follow a 8px baseline. This means that the default size is now 76px instead of 75px. I'm pretty certain no one will notice 😅

dkoo (changes requested):
The image display is working well, but I have one request: is it possible to skip the crop step (or allow skipping it) when selecting a footer image? It seems there's no way to select an existing image without cropping it first, and even if I set the crop boundaries to the full image, it creates a duplicate of the selected image.

Also, when cropping I frequently ran into the following error. Not sure if it's related to anything codewise or if it's just my local environment, though:

Screen Shot 2021-03-16 at 10 43 37 AM

thomasguillot:
@dkoo done in https://github.com/Automattic/newspack-theme/pull/1302/commits/299e84fcb47952048bc6bb18c3f8938bcd6b5fd1

Note: It should affect current sites because `newspack-footer-logo` has already been generated and it would require to use something like [Regenerate Thumbnail](https://wordpress.org/plugins/regenerate-thumbnails/) to resize old images.

If you upload a new image (and/or use Regenerate Thumbnail to get the correct size) then the skip cropping should work 😊

dkoo (approved):
Exactly what I wanted, thanks! Question, though:

> If you upload a new image (and/or use Regenerate Thumbnail to get the correct size) then the skip cropping should work 😊

Does this mean if an existing site sets the footer image to a previously existing image (before the new image size) AND skips cropping, it'll show the full-size image there? That's not really a bug IMO, but maybe something we should let publishers know about when rolling out the new feature.

thomasguillot:
> Does this mean if an existing site sets the footer image to a previously existing image (before the new image size) AND skips cropping, it'll show the full-size image there? That's not really a bug IMO, but maybe something we should let publishers know about when rolling out the new feature.

If they don't run  Regenerate Thumbnail, then no, the existing image will have the previously forced ratio.
If they upload a new image, it will be fine.

Tbh, it sounds like an edge case that would be easily fixed by the publication/TAM.

PR #1323, opened 2021-04-01
Add custom style to subscribe block and patterns

thomasguillot:
### All Submissions:

Test this along side https://github.com/Automattic/newspack-blocks/pull/730

This PR slightly tweaks the CSS of the Mailchimp block and add custom CSS specific to Newspack Pattern styles 5 and 6. 

1. Switch to this branch (make sure you also have https://github.com/Automattic/newspack-blocks/pull/730)
2. Add style 5 and/or 6 to your page

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -966,25 +966,35 @@ hr {
 
 	.newspack-pattern.subscribe__style-5 &,
 	.newspack-pattern.subscribe__style-6 & {
-		form {
-			align-items: center;
-			display: flex;
-			flex-wrap: wrap;
+		.wp-block-button__link {
+			width: 100%;
 
-			> * {
-				flex: 0 0 100%;
+			@include media( mobile ) {
+				width: auto;
 			}
+		}
 
-			> p:first-child {
-				flex: 1 1 auto;
-				margin-bottom: $size__spacing-unit / 2;
-				margin-right: $size__spacing-unit / 2;
-			}
+		@include media( tablet ) {
+			form {
+				align-items: center;
+				display: flex;
+				flex-wrap: wrap;
 
-			> .wp-block-jetpack-button.wp-block-button {
-				flex: 0 0 auto;
-				margin-left: auto;
-				margin-top: 0;
+				> * {
+					flex: 0 0 100%;
+				}
+
+				> p:first-child {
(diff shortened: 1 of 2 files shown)


adekbadek (commented):
The only difference I see (at least in the editor) is in the paragraph alignment in subscribe pattern 5. What else should change?

thomasguillot:
> The only difference I see (at least in the editor) is in the paragraph alignment in subscribe pattern 5. What else should change?

Correct, in the editor that's the only change. I didn't change the aligment of the submit buttn like on the front end because it's an inner block and I think it's ok if it's not exactly 1:1.

Front-end, style 5 and 6 should have the submit button alongside the email input

Screenshot 2021-04-01 at 10 41 15

_Note: I couldn't do it for all the patterns because some live sites are already using patterns 1 to 4_

adekbadek:
> Correct, in the editor that's the only change. I didn't change the aligment of the submit buttn like on the front end because it's an inner block and I think it's ok if it's not exactly 1:1.

Hm, I have to disagree here. I think small discrepancies (e.g. spacing, font sizes) between editor and frontend are ok, but different layout of the buttons looks like a bug. 

Here's a comparison (editor on the left):

image


> I couldn't do it for all the patterns because some live sites are already using patterns 1 to 4

Not sure if I follow here. Does this mean that patterns cannot be updated?

thomasguillot:
> Hm, I have to disagree here. I think small discrepancies (e.g. spacing, font sizes) between editor and frontend are ok, but different layout of the buttons looks like a bug.

Pushed some changes: https://github.com/Automattic/newspack-theme/pull/1323/commits/5f7fea50243f76d6f9b845007910993c10c8125c

> Not sure if I follow here. Does this mean that patterns cannot be updated?

Correct -- otherwise it could "affect" live sites that might already have some sort of custom css.

dkoo:
> Correct -- otherwise it could "affect" live sites that might already have some sort of custom css.

I didn't think patterns that are already placed would "update" if the original patterns change—when you add a pattern to a post, doesn't it just place the pattern as static blocks in the editor? There's no connection maintained to the pattern after it's placed.

thomasguillot:
@dkoo the group that contains the pattern has some specific classes `newspack-pattern subscribe__style-*` which is what I'm using the style the patterns

adekbadek (approved):

PR #1355, opened 2021-05-14
feat: allow limited text-formatting html elements in subtitles

### LINKED ISSUE #1136: Allow italic and bold tags in subtitles
Need to allow italic and bold tags in subtitles, with corresponding editor preview

This will probably need its own `newspack_subtitle_kses` function or something, and should allow `<i>`, `<em>`, `<b>`, and `<strong>` tags.

dkoo:
### All Submissions:

Allows a limited set of text formatting HTML elements in post subtitles. The list of allowed tags is swiped from [this reference](https://www.w3schools.com/html/html_formatting.asp), plus `u` tags for underlining.

Some potential improvements:

* Use a `RichText` field instead of a simple `TextAreaControl` field to edit the subtitle. I couldn't get the `RichText` component to render with a formatting toolbar outside of a block context, though, so I'm not sure this would be supported by Gutenberg.
* Sanitize the HTML string in the editor preview before rendering as `innerHTML`, so that the output matches the front-end output more exactly. See step 3 below for more details.

Note that `em` and `i` tags are rendered with _non-italic_ styling because the subtitle as a whole is italicized. This is the generally accepted way to handle italics-within-italics in written English, but noting here just in case.

Closes #1136.

1. Edit a post, expand the "Article Subtitle" sidebar panel.
2. Enter a string with some of the allowed HTML tags, plus some disallowed tags. Example string you could use:

```
Regular text <b>bold</b> <i>italic</i> <small>small</small> <mark>marked</mark> <del>deleted</del> <sup>superscript</sup> <sub>subscript</sub> <u>underlined</u> <div>disallowed (appears as regular text)</div>
```

3. Confirm that the editor preview updates and renders the string with HTML formatting. **Note** that the editor preview will render _all_ HTML, not just the allowed ones—not sure the best way to prevent this without some very complicated regex-ing and am open to suggestions on this!
4. Save the post, view on the front-end. Confirm that the subtitle is rendered with the allowed HTML formatting, with any disallowed tags stripped.

Screen Shot 2021-05-14 at 4 40 13 PM

### Other information:

### DIFF
--- a/newspack-theme/js/src/post-subtitle/utils.js
+++ b/newspack-theme/js/src/post-subtitle/utils.js
@@ -25,7 +25,7 @@ export const appendSubtitleToTitleDOMElement = ( subtitle, isInCodeEditor ) => {
 			}
 			titleEl.appendChild( subtitleEl );
 		}
-		subtitleEl.innerText = subtitle;
+		subtitleEl.innerHTML = subtitle;
 	}
 };
 
--- a/newspack-theme/sass/site/primary/_posts-and-pages.scss
+++ b/newspack-theme/sass/site/primary/_posts-and-pages.scss
@@ -51,6 +51,11 @@
 	@include media( mobile ) {
 		margin-bottom: 2.3em;
 	}
+
+	em,
+	i {
+		font-style: normal;
+	}
 }
 
 body.page {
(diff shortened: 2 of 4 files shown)


adekbadek (approved):
🙌

PR #1368, opened 2021-06-11
feat: enable jetpack share buttons on pages, too

dkoo:
### All Submissions:

Allows Jetpack's share buttons to appear on pages. Currently they're only allowed to display on singular posts of any post type. This comes from a request by a few publishers who want to display these buttons on their About pages, so please let me know if this was an intentional decision.

1. In WP admin, go to **Jetpack > Settings > Sharing**. In the "Sharing buttons" section, turn on the "Add sharing buttons to your posts and pages" option and configure sharing buttons so at least a few are enabled.
2. On `master`, view posts of various types using all the available templates and note how the buttons appear. Then view a page and observe that no share buttons are displayed.
3. Check out this branch and refresh the posts and the page. Confirm that a.) the share buttons appear on posts exactly as they did while on `master`, and b.) the share buttons now appear on pages as well.

### Other information:


laurelfulford (changes requested):
Thanks for creating this PR, @dkoo! 

It works as described (though I'm not 100% sure about the placement... though it's hard to match how the single posts look, since pages are missing a lot of the same elements).

One concern I have is in my testing, share buttons are on by default on both posts and pages if you've turned them on from the site. You can turn them off on pages in your WP.com marketing settings, but since they haven't been showing up on pages yet, I'm not sure how many publishers would have done that - it's very possible we'll have a bunch of unwanted sharing buttons suddenly appear. 

My second concern is that if publishers have only asked for these on one page for their site (the about page), this might be overkill -- it'll turn them on for every single page. 

I don't know the ins and outs of this request, but if not, what do you think about finding a plugin/block that would do this instead? Or perhaps build an option into the theme that will let people opt into the page share buttons on a per-page basis on top of the Jetpack control, so it can be finer tuned? 

Just let me know if you have any questions about the above!

dkoo:
> One concern I have is in my testing, share buttons are on by default on both posts and pages if you've turned them on from the site. You can turn them off on pages in your WP.com marketing settings, but since they haven't been showing up on pages yet, I'm not sure how many publishers would have done that - it's very possible we'll have a bunch of unwanted sharing buttons suddenly appear.

> My second concern is that if publishers have only asked for these on one page for their site (the about page), this might be overkill -- it'll turn them on for every single page.

Those are excellent points. Definitely don't want to cause unintended changes for publishers who are happy with the current behavior.

> Or perhaps build an option into the theme that will let people opt into the page share buttons on a per-page basis on top of the Jetpack control, so it can be finer tuned?

This might be a good idea. What if we add a meta option to pages only that is displayed only if the Jetpack sharing module is enabled? The option could be a toggle switch called something like "Show Jetpack share buttons" and is turned off by default, but if turned on, it will display the share buttons on that page only?

laurelfulford:
> What if we add a meta option to pages only that is displayed only if the Jetpack sharing module is enabled? The option could be a toggle switch called something like "Show Jetpack share buttons" and is turned off by default, but if turned on, it will display the share buttons on that page only?

I think this might be the best way to have consistency across posts and pages -- both will use the Jetpack sharing function, but it won't suddenly appear where it shouldn't, and it'll allow a per-page control. It feels a bit hacky, but that fits in with a lot of other functionality in the theme! :) 

Just let me know if you'd like me to pick this one up, or I'm happy to re-review if you'd rather take care of it!

dkoo:
@laurelfulford I added the post meta option for pages as described above: b7ef72c

To test:

1. Edit a page, observe a new "Show Jetpack share buttons" option under the post status sidebar:
Screen Shot 2021-07-05 at 3 03 31 PM

2. Confirm that the option is OFF by default.
3. View the page's front-end while it's off, confirm that no share buttons are shown.
4. Turn it ON, refresh the front-end, confirm that the share buttons are now shown.
5. Edit a non-page post, confirm that the option isn't shown. View the post's front-end, and confirm that the share buttons are shown (the option only applies to pages).
6. Disable Jetpack or the sharing functionality in **Jetpack > Settings > Sharing**. Confirm that with that off, the option isn't shown in the page editor, and the share buttons are never shown in posts or pages.

laurelfulford (changes requested):
Thanks for updating this, @dkoo! 

In my testing it mostly works; the only issue I ran into is with using `wp_kses_post` -- it strips out the share buttons entirely when AMP is enabled (due to the non-standard markup), and seems to strip out random other buttons depending on the settings (eg. if you use the official buttons, for example, LinkedIn goes AWOL). Apologies for not catching this before, I don't think it's newly introduced with this update! 

In theory we could allow all of the necessary tags through a filter, but that could turn into a lot between the AMP-specific tags, and whatever markup the different Jetpack button options use (not to mention if anything ever changes! 😰). 

My very inelegant work around would be something like...

```
if ( '' !== $sharing_enabled ) {
    sharing_display( '', true );
}
```

... which seems to work, but is also pretty clunky compared to what's there now. There's probably a better way - let me know if you have any questions about the above!

dkoo:
> the only issue I ran into is with using wp_kses_post -- it strips out the share buttons entirely when AMP is enabled (due to the non-standard markup)

Ah, good catch! I happened to have AMP off on the page I was testing with. 54d3347 should fix this. I notice that if you use the "official buttons" they're rendered as icons-only in AMP mode, but that's probably expected behavior from Jetpack, right?

laurelfulford (approved):
Thanks Derrick! This looks good to me. 

> I notice that if you use the "official buttons" they're rendered as icons-only in AMP mode, but that's probably expected behavior from Jetpack, right?

Yup, that's expected! It's definitely caused some confusion when folks have picked a different visual option through Jetpack, though. Some buttons also aren't supported by AMP, so they'll get stripped out in the AMP version of the page.

PR #1382, opened 2021-06-23
fix: figcaption rendering bug in Safari

### LINKED ISSUE #1132: Image Block: Caption placement in Safari
It looks like #1055 is also happening on Safari on desktop.

Unfortunately the fix there (switching the image block from display table to display block) causes other issues (eg. narrow floated images with captions let the caption be wider than the image, rather than matching the width of the image). So we probably need to find a different fix for all cases.

dkoo:
### All Submissions:

Fixes an [obscure Safari rendering bug](https://github.com/ampproject/amp-wp/issues/6035) which causes captions inside image blocks with an alignment set to render in the wrong position.

Closes #1132

1. Add an image block to a post and choose an image that has a caption. Set an alignment on the block (left, center, or right).
2. On `master`, view the post in Safari. Observe that the caption appears in the wrong place (if you don't see the bug, try hard-refreshing the post):

Screen Shot 2021-06-23 at 3 44 30 PM

3. Check out this branch and hard-refresh the post. Confirm that the caption now renders as expected.
4. Re-test with all alignment options and no alignment options, and in another browser to ensure that the fix doesn't change anything outside Safari.

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -769,14 +769,6 @@ p.has-background {
 	}
 
 	@include media( mobileonly ) {
-		.alignleft,
-		.aligncenter,
-		.alignright {
-			figcaption {
-				display: block;
-			}
-		}
-
 		.aligncenter {
 			display: block;
 			img {
--- a/newspack-theme/sass/media/_captions.scss
+++ b/newspack-theme/sass/media/_captions.scss
@@ -24,8 +24,11 @@ figcaption,
 	line-height: $font__line-height-pre;
 }
 
-.wp-block-image [class^='align'] > figcaption {
-	display: block;
+/* Hack to address a bug that affects caption positioning in Safari only. Targets Safari 9+ only. */
+@supports ( -webkit-hyphens: none ) {
+	.wp-block-image [class^='align'] > figcaption {
+		display: block;
+	}
 }
 
 .entry-content {
--- a/newspack-theme/sass/media/_captions.scss
+++ b/newspack-theme/sass/media/_captions.scss
@@ -24,6 +24,10 @@ figcaption,
 	line-height: $font__line-height-pre;
 }
 
+.wp-block-image [class^='align'] > figcaption {
+	display: block;
+}
+
 .entry-content {
 	.wp-caption-text,
 	figcaption {


laurelfulford (changes requested):
There's a version of this fix [that's only applied to mobile in the _blocks.scss](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/sass/blocks/_blocks.scss#L772-L778) file though this location makes more sense! That one should probably be removed, too as part of this PR. 

The theme did include this fix for screen sizes at one point, but it causes smaller floated images to have their captions run wider than the images themselves. There's some more information and screenshots for that here:

https://github.com/Automattic/newspack-theme/pull/1109

IIRC the too-small-images-and-wide-captions issue was only reported once while this fix was in the theme, so I think that might be an acceptable trade-off at this point, since the caption gap issue has come up pretty often. I've been manually adding CSS fixes to the sites that report it, but it's probably better to go the other way around and manually undo it on any sites that report the small-image-wide-caption issue instead (I remember the site that originally reported the issue, so I can take care of that ahead of this being merged). 

Let me know what you think about the above! If it sounds okay, I think this will be good to go once the near-duplicate styles are removed from _blocks.scss.

dkoo:
Wow, thanks for the historical context around this bug! That's a pretty infuriating trade-off—either we fix the bug in Safari at the expense of introducing another bug with smaller images in all browsers, or we leave the bug in Safari...

As discussed off GitHub, maybe the best hybrid approach is to fix the bug only in Safari and leave core styles active for all other browsers. [This StackOverflow thread](https://stackoverflow.com/questions/16348489/is-there-a-css-hack-for-safari-only-not-chrome) discusses some CSS hacks that supposedly work to target CSS in specific versions of Safari. I tried the `@supports ( -webkit-hyphens: none )` hack that targets Safari 9+ (introduced 2015) and it seems to do the trick. I also removed the duplicate styles with the media query targeting only mobile viewports in `_blocks.scss` as you requested. 64c9596

laurelfulford (approved):
Thanks @dkoo!

I took this for a spin both locally and using Browserstack; I can confirm this fix works, and, by using a smaller image, confirm that it's only being applied to Safari 9+ (on Safari 8, and other browsers, the smaller image's caption matches its width). 

Thanks for tackling this! I'm looking forward to never hearing about this one again, it's been an annoying one 😄

laurelfulford:
@dkoo I'm on a bit of a merging spree so I'm going to go ahead and merge this one! :)

dkoo:
Oh, thanks for merging!

PR #1396, opened 2021-07-01
feat: customizing svg icon for social media with a hook

uptimizt:
### All Submissions:

I added a hook to add SVG icons for the custom social media.

For example, I need Telegram or MoyKrug now, and tomorrow I might want to add new social media and need a hook for fast customization.

usage example
```

add_filter('newspack_get_social_link_svg', function ($svg, $uri, $size, $title){
    if(!empty($svg)){
        return $svg;
    }
    if(strpos($uri, 'telegram') === false ){
        return $svg;
    }

    return <<<SVG
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none">
<g clip-path="url(#clip0)">
<path d="M9.41717 15.1809L9.02017 20.7649C9.58817 20.7649 9.83417 20.5209 10.1292 20.2279L12.7922 17.6829L18.3102 21.7239C19.3222 22.2879 20.0352 21.9909 20.3082 20.7929L23.9302 3.82092L23.9312 3.81992C24.2522 2.32392 23.3902 1.73892 22.4042 2.10592L1.11418 10.2569C-0.338822 10.8209 -0.316822 11.6309 0.867177 11.9979L6.31017 13.6909L18.9532 5.77992C19.5482 5.38592 20.0892 5.60392 19.6442 5.99792L9.41717 15.1809Z" fill="white"/>
</g>
<defs>
<clipPath id="clip0">
<rect width="24" height="24" fill="white"/>
</clipPath>
</defs>
</svg>
SVG;
}, 11, 4);
```

example result:

Mobile People News – Connecting people   mobile technologies 2021-07-01 13-03-16

### Other information:


laurelfulford (changes requested):
Thanks for submitting this PR, @uptimizt!

One thing I noticed while testing is that you can pass any markup to the menu, not just an SVG -- for example:

```
add_filter('newspack_get_social_link_svg', function ( $svg, $uri, $size, $title ) {
    if ( ! empty( $svg ) ) {
        return $svg;
    }
    if ( strpos( $uri, 'telegram' ) === false ) {
        return $svg;
    }
    return '<script>alert("hello world!!");</script>';
}, 11, 4);
```

... will fire an alert if you include a link with telegram in it in the social menu. 

With opening this up to be filter-able, what do you think about adding some kind of sanitation to make sure only SVG-related elements are passed through? The theme [already has a function](https://github.com/Automattic/newspack-theme/blob/c2cd40f731c6c2173f8f4346281823f94d9b95cb/newspack-theme/functions.php#L745-L770) that returns allowed SVG tags that's used [in other parts of the theme](https://github.com/Automattic/newspack-theme/blob/5fa724dddf7cd0e3b3ace5ee52b509bfa3199945/newspack-theme/header.php#L57) with `wp_kses` -- it may need expanding over time to make sure it allows all of the SVG tags someone might use, but so far it covers everything used by all of the SVGs in the theme.

Let me know what you think, and if you have any questions at all - thanks!

uptimizt:
@laurelfulford thank you for the feedback.

I added wp_kses.

But I guess what if users want to add a PNG icon? or WebP?

laurelfulford:
Thanks for the update, @uptimizt!

> But I guess what if users want to add a PNG icon? or WebP?

The theme only uses SVGs in that menu, so it seems like overkill for the filter to support other image types when the theme itself doesn't use them. 

By keeping everything SVG, it makes it much easier to standardize the icons appearance with what's included in the theme, and the icons will also automatically change colour to have sufficient contrast when you use the different header background options. They can also be customized in a much more standard fashion (eg. changing the fill in the Custom CSS will be applied to all of the icons). 

I think for now it keeps things much more straight forward just to support the one format. But please let me know if you have any questions at all about the above!

laurelfulford (approved):

Issue #1372, opened 2021-06-21
Newspack Blocks: Update block styles in widget area


laurelfulford:
Block widgets will be included in WP 5.8, and our Newspack blocks may need some visual tweaking when placed in the theme's sidebar, slide out sidebar, or footer areas. 

Issues to fix:

- [x] Homepage Posts - issues fixed in #1407
- [x] Donate - issues fixed in #1398
- [x] Author Profile - issue will be fixed in https://github.com/Automattic/newspack-theme/pull/1422
- [x] Carousel - issues fixed in https://github.com/Automattic/newspack-theme/pull/1411

Some potential tweaks include:

### Homepage posts block:
* Reduce/adjust font sizes (? I'm a bit torn on this one, since on one hand you can already change font sizes, but on the other hand, the upper range of font sizes will be useless in the sidebar area, unless some really bad decisions are being made). 
* Reduce avatar size
* Make sure there's sufficient contrast when added to the slide-out sidebar & mobile menu areas
* The 'image behind' placement isn't filling the space in the mobile and slide-out sidebars due to all links being set to `display: inline-block`; there's a general padding and margin on links in both that's affecting the spacing too -- all of these styles assume all links are navigation items, and need to be updated to be more specific. 
* Newspack Katharine's overlap, and all-caps in the site info section don't work in areas with background colours, like the widget areas.

![image](https://user-images.githubusercontent.com/177561/123150762-9390aa80-d417-11eb-8b3b-743c7e943058.png)

![image](https://user-images.githubusercontent.com/177561/123844350-c6c7b380-d8c7-11eb-8eca-4f25d5d203f9.png)

![image](https://user-images.githubusercontent.com/177561/124515225-bf0a8200-dd93-11eb-8922-9b3a78967e80.png)

![image](https://user-images.githubusercontent.com/177561/124521679-4e6c6100-dda5-11eb-9a63-bc9a95bbedda.png)

### Donate block:
* Adjust layout to mobile view when in sidebar, slide-out-sidebar, at all times. 
* Potentially adjust layout to mobile view in footer widget area if there are multiple widgets. 

Example of block in sidebar area; longer number values, like on the Annually tab, don't have enough room - tightening up some of the space between elements here, and the padding, should help: 

![image](https://user-images.githubusercontent.com/177561/123142512-5e338f00-d40e-11eb-8199-4581ac35b921.png)

[...]

PR #1398, opened 2021-07-02
Donate Block: improve donate block appearance for WP 5.8

laurelfulford:
### All Submissions:

This PR cleans up some small issues in the Donate Block when placed in the widget block spaces in WordPress 5.8, specifically making sure it's still picking up the custom colours, and reducing the padding a bit when it's in narrower spaces.

See #1372

1. Start on a test site running the latest WordPress 5.8 RC; you can do this using the [Beta Tester](https://wordpress.org/plugins/wordpress-beta-tester/) plugin.
2. Enable the slide-out sidebar, and set those widgets to also appear in the mobile menu.
3. Set up your site to use custom colours and fonts; set the footer to use a darker custom footer colour - this will help us test contrast, and that the block is picking up the custom colours when appropriate. 
4. Add the Donate widget to every widget space: sidebar, footer, slide-out sidebar, above copyright, and above and below the content area. 
5. View on the front end; note that the block is not always picking up the custom colours, and in some cases there's a bit more padding than needed:

Sidebar - spacing:
image

Slide-out Sidebar - spacing and colour:
image

Slide-out Sidebar widgets in the mobile area - spacing and colour:
image

Footer and Above Copyright - colour; weird 'thank you' spacing:
image

7. Confirm that the above issues are resolved:

Sidebar:

image

Slide-out Sidebar: 

image

Slide-out Sidebar widgets in the mobile area:

image

Footer and Above Copyright:

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -232,6 +232,21 @@ p.has-background {
 	}
 }
 
+//! Newspack Donate Block
+#secondary .wpbnbd,
+.desktop-sidebar .wpbnbd,
+.mobile-sidebar .wpbnbd {
+	&.tiered .tiers {
+		margin-left: #{0.5 * $size__spacing-unit};
+		margin-right: #{0.5 * $size__spacing-unit};
+	}
+	.thanks,
+	button {
+		margin-left: $size__spacing-unit;
+		margin-right: $size__spacing-unit;
+	}
+}
+
 //! Columns
 .wp-block-columns {
 	.wp-block-cover,
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -233,17 +233,23 @@ p.has-background {
 }
 
 //! Newspack Donate Block
-#secondary .wpbnbd,
-.desktop-sidebar .wpbnbd,
-.mobile-sidebar .wpbnbd {
-	&.tiered .tiers {
-		margin-left: #{0.5 * $size__spacing-unit};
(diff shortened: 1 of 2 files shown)


newspack-theme/sass/blocks/_blocks.scss
```diff
@@ -232,6 +232,27 @@ p.has-background {
 	}
 }
 
+//! Newspack Donate Block
```
adekbadek, line 235: More of a general remark, but I find it a bit weird that there are block styles in the theme. Why are these not in `newspack-blocks`?

adekbadek (approved):

newspack-theme/sass/blocks/_blocks.scss
laurelfulford: That's a good question! I opted to put these styles in the theme instead of the plugin because they require the theme, if that makes sense. If it was a more general style that was always applied to the Donate block I agree it should go in the plugin. In this case, the block needs to be in certain areas of the theme (the sidebar, the mobile menu) for the styles to be applied, and with any other theme, these styles wouldn't be used. It's kind of similar to the custom colour/typography styles living in the theme instead of the plugin - even though those styles are applied to the block, it's only in the context of this theme, and wouldn't be useful otherwise. 

Let me know if this doesn't make sense as an approach!

laurelfulford:
Thanks Adam! I'm going to go ahead and merge this but I'm open to discussing the location of these styles further if my logic doesn't make sense :)

newspack-theme/sass/blocks/_blocks.scss
adekbadek: Ah, that makes sense, thanks for the clarification!

PR #1402, opened 2021-07-05
fix: ads media queries

adekbadek:
### All Submissions:

Fixes a small issue.

1. On `master`,
1. Create an ad unit with only mobile (<780) sizes
2. Insert it in a post using an ad block or SCAIP
3. Observe a PHP notice logged when viewing the post on desktop screen size: `PHP Notice:  Trying to access array offset on value of type null`
4. Switch to this branch, refresh the page, observe no PHP notice logged

### Other information:


dkoo (approved):
I was only able to replicate the PHP notice when placing an ad via SCAIP (not via the ad unit block), but can confirm that this fixes it.

PR #1410, opened 2021-07-12
Add option to change the footer layout

### LINKED ISSUE #1393: Footer widget area + blocks
Right now, the footer widget area attempts to smartly format the widgets into up to four columns.

With blocks becoming available in the widget area in WP 5.8, this may cause issues. At the very least, we need to test to confirm that adding a group or column block around the block area will let you lay things out freely. 

It's possible we can also use the block-related classes on the widgets to override some of the theme's baked in behaviour. Worst case, we can add some kind of override or check to see if this layout behaviour is desired.

laurelfulford:
### All Submissions:

By default, the Newspack theme uses flexbox for the footer widgets, dividing them into up to four columns automatically.

With WordPress 5.8's widget blocks, this isn't a necessary tool for layouts -- the columns block is available, on top of other blocks that you may want to stack or have behave a different way than the default.

One work around is to wrap everything in the widget area in a single block, like a group or column block. Since that might be a bit unintuitive -- and a bit fragile, since you always have to remember to add new widgets inside of that containing block -- I've also added a toggle to the Customizer so the column behaviour can be 'turned off'. 

The CSS it uses is a bit clunky - it adds a class when you chose to stack the columns, and then uses that to override the flexbox styles. This is to avoid introducing more specific styles to create the current layout, since some Newspack sites already use Custom CSS to override that, and I don't want to break those sites. 

Closes #1393.

1. Add 2-4 widgets to the footer area; note that they sort into columns.
2. Navigate to Customizer > Footer Settings, and confirm there is now two radio options for the Footer layouts, either Columns (default) or Stacked. (I went with radio buttons instead of a single checkbox option, on the off-off-off chance we need to add a third layout down the road). 

image

3. Switch the Footer Layout to 'Stacked'. 
4. Confirm the widgets are now stacked on top of one another, and take up the full content width (1200px) of the widget area. 

### Other information:

### DIFF
--- a/newspack-theme/inc/customizer.php
+++ b/newspack-theme/inc/customizer.php
@@ -1100,6 +1100,28 @@ function newspack_customize_register( $wp_customize ) {
 		)
 	);
 
+	// Add option to toggle off the default footer layout.
+	$wp_customize->add_setting(
+		'footer_widget_layout',
+		array(
+			'default'           => 'columns',
+			'sanitize_callback' => 'newspack_sanitize_radio',
+		)
+	);
+	$wp_customize->add_control(
+		'footer_widget_layout',
+		array(
+			'type'        => 'radio',
+			'label'       => esc_html__( 'Footer widget layout', 'newspack' ),
+			'description' => esc_html__( 'Stack the footer widgets, or have them automatically divide into even columns.', 'newspack' ),
+			'choices'     => array(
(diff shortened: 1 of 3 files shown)


dkoo (approved):
This is a pretty clean solution to the problem without affecting existing sites' footers. LGTM!

laurelfulford:
Thanks Derrick!

PR #1416, opened 2021-07-14
Add support for custom spacing

### LINKED ISSUE #1391: Add support for custom spacing?
WP themes can opt into Custom Spacing for some blocks, which could be a huge help with spacing weirdness we have with some block combinations.

`add_theme_support('custom-spacing')` needs to be tested to see if it will work well with the theme and Newspack blocks, to see if it's something that could be worth adding.

laurelfulford:
### All Submissions:

This PR adds support for Gutenberg's 'custom spacing', which will add an option to manually add padding or spacing around on a per block basis to blocks which have added support for spacing. 

In Core, these blocks include:

* Column
* Group
* Cover
* Verse
* Site Tagline (5.8+)
* Site Title (5.8+)
* Template Part (5.8+ - but won't appear in a site running the theme, since we need to opt-in to see it)

In WP 5.8, at least one of the example Query Loop patterns includes padding on a Group block, so without this control visible in the theme, it's not possible to remove that padding (without flipping to the code editor and picking the padding out of the block that way... not great). The option should work in 5.7.2 though. 

This control will be helpful for one-off block adjusting, but we should discourage publishers from using this to control padding _everywhere_ -- like, if they always want the column block to have more spacing site wide, or always want the group block with a background to have less, it's much better for us to add it to their Custom CSS. Otherwise, all of the lines of inline styles really add to the CSS page load, and can make it hit the AMP limit prematurely. 

Closes #1391

1. Apply the PR.
2. Try adding each of the above blocks; and confirm a section labelled 'Spacing' appears on the right when you select each block. (Note: some blocks are 5.8 specific but they're also not ones that are really useful outside of FSE).

image

3. Try adjusting the block's padding; you should have the option to add padding evenly to all four sides by entering a pixel amount in the text field, or you can click the link icon to get fields for each the top right bottom and left.
4. Publish your post; confirm your changes appear on the front-end. 

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -252,6 +252,9 @@ if ( ! function_exists( 'newspack_setup' ) ) :
 			)
 		);
 
+		// Add block custom spacing support.
+		add_theme_support( 'custom-spacing' );
+
 		// Add support for responsive embedded content.
 		add_theme_support( 'responsive-embeds' );
 


miguelpeixe (approved):
What a nice feature!

laurelfulford:
Thanks Miguel! 🙌

PR #1453, opened 2021-07-30
fix: update event calendar shortcode custom fonts

laurelfulford:
### All Submissions:

While the theme's fonts are getting picked up by the Events pages, there's something about the way the styles are loaded for The Events Calendar Pro shortcodes that causes the per-site custom typography not to override the theme styles.

This PR makes the custom typography styles for TEC a bit more specific, to override the theme styles. 

Note: I'm seeing a mix of the custom Accent Colour I set in the Events Calendar settings and the default blue that comes from the plugin in my shortcodes. I'm not sure if this is a bug with the plugin, or an issue with me adding so many of these shortcodes on one page -- the custom colours from the plugin are applied via JavaScript so I'm not sure if that would trip them up. 

1. Start with a test site running The Events Calendar PRO plugin, and a few events added. Make sure under your settings that:
     * That you're running the v2 views (under WP Admin > Events > Settings > Display, the Enable updated designs for all calendar views option should be checked).
     * That your site is set to use 'Tribe Events Styles' (under WP Admin > Events > Settings > Display)
     * That your site is set to use the 'Default Events Template' (under WP Admin > Events > Settings > Display).
2. Pick one of the child themes, and set some custom fonts that are noticeably different than that child theme's defaults. I recommend not using the default Newspack theme because its fonts are really similar to what the Events Calendar uses by default -- not using it will help make it easier to spot issues, if those plugin fonts are getting picked up.
3. Add a page; turn off AMP and add some shortcodes -- some examples to test:
     * `[tribe_mini_calendar]`
     * `[tribe_events_list]`
     * `[tribe_featured_venue slug="replace-slug"]` -- you need to replace the slug value with one of a venue you've created on your site
     * `[tribe_event_countdown slug="replace-slug"]` -- you need to replace the slug value with one of an event you've created on your site
     * `[tribe_this_week]`
     * `[tribe_events]`
     * `[tribe_events view="month"]`
     * `[tribe_events view="list"]`
     * `[tribe_events view="day"]`
     * `[tribe_events view="[photo"]`
     * `[tribe_events view="week"]`
     * `[tribe_events view="map"]`
     * `[tribe_events view="summary"]`
  

4. View on the front-end; note that the shortcodes are using your theme's fonts, not your custom fonts. On my test site, I'm using Newspack Scott, so I have a lot of Fira Sans Condensed, rather than the Merriweather header I picked:

image

image

image

6. Confirm that your shortcodes are now picking up your custom fonts (on my site, Merriweather and Arial):

image

image

image

### Other information:


miguelpeixe (approved):
💯

laurelfulford:
Thanks! 😊

Issue #1265, opened 2021-02-25
Simplified Subpage Header: Doesn't show anything if no logo


laurelfulford:
If you are using the site title, rather than the logo, the simplified subpage header doesn't show it.

Though it should hide the site title if you have a logo in that case, it should show it if there's no logo assigned.

PR #1459, opened 2021-08-03
Default theme: fix tertiary menu appearance in slide-out sidebar

### LINKED ISSUE #1406: Subpage Slide-out Menu: Tertiary menu display incorrectly
With the default Newspack theme, the tertiary menu in the Subpage Slideout menu doesn't pick up the correct colours, and the spacing is wrong:

image

laurelfulford:
### All Submissions:

When using the default Newspack theme + slideout sidebar header option + tertiary menu, the correct styles (padding, colours) aren't being picked up in all cases. 

This PR fixes those styles.

Closes #1406

1. Start with a site running the Default Newspack theme.
2. Under Customizer > Header Settings > Subpage Header, turn on the Simplified Subpage Header
3. Add a menu and assign it to the Tertiary Menu location; add at least two links, and add the CSS class `menu-highlight` to one of them.
4. View on the front-end and check the appearance of this menu on the homepage and on a subpage (where it's inside of the slideout sidebar), both with a solid header background colour (Customizer > Header Settings > Appearance; check Solid Background), and without. Note the following issues:

**Homepage with default header background:**
image
* Correctly uses primary colour for `menu-highlight` button

**Homepage with solid header background:**
image

* Correctly uses secondary colour** for `menu-highlight` button, and a slightly darker version of the header colour for regular buttons

_** Note: this button switches to the secondary colour as a work-around for the original solid header bg option, which would always use the primary colour._

**Slide-out sidebar with default header background:**
image
* Wrong button colour on the `menu-highlight` link (it should be using my custom primary colour but is using Newspack blue)
* Wrong padding on the standard link

**Slide-out sidebar with solid header background:**
image
* Wrong button colour on the `menu-highlight` link (it should be using the primary colour)
* Wrong button colour on the regular link (it should be using the regular grey)

6. Confirm the menu is now displaying as expected from the homepage and on subpages:

**Homepage with default header background:**
image

**Homepage with solid header background:**
image

**Slide-out sidebar with default header background:**
image

**Slide-out sidebar with solid header background:**
image

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -335,7 +335,8 @@ function newspack_custom_colors_css() {
 				}
 				.cat-links a,
 				.cat-links a:visited,
-				.site-header .nav3 .menu-highlight a {
+				.site-header .nav3 .menu-highlight a,
+				.subpage-sidebar .nav3 .menu-highlight a {
 					background-color: ' . esc_html( $primary_color ) . ';
 					color: ' . esc_html( $primary_color_contrast ) . ';
 				}
--- a/newspack-theme/sass/styles/style-default/style-default.scss
+++ b/newspack-theme/sass/styles/style-default/style-default.scss
@@ -35,7 +35,7 @@
 	}
 }
 
-.h-sb .nav3 {
+.h-sb .site-header .nav3 {
 	a {
 		background-color: darken( $color__primary, 10% );
 		color: #fff;
@@ -68,6 +68,10 @@
 	.menu-highlight a {
 		background-color: darken( $color__primary, 10% );
 	}
+}
+
+.subpage-sidebar .nav3 a {
+	padding: #{$size__spacing-unit * 0.5} #{$size__spacing-unit * 0.75};
 }
 
 /* Posts and pages */


adekbadek (approved):
Works, but the issue was only reproducible on a single post page, where the header behaves differently than on the homepage. Approving since that's not tied to this PR, but should it be so?
On homepage, the header has the logo, but it does not on a single post:

image

Also, on a single post when the resolution is ~840px, header layout looks weird:

image

If these situations are unintended, I'll open an issue.

laurelfulford:
Thanks @adekbadek!

> On homepage, the header has the logo, but it does not on a single post:

There is a known issue with using the subpage header and _not_ having a logo -- the subpage header styles are set to hide the site title, so if that's all that's displaying, you don't have anything. It's been a lower priority because so far we don't have any sites that don't have logos:

https://github.com/Automattic/newspack-theme/issues/1265

Can you confirm if your test site has a logo, or just the site title? There might be some other weird case that's not working that I'm missing here! 

> Also, on a single post when the resolution is ~840px, header layout looks weird:

I've added an issue for the weird gappy mobile CTA here (thanks for spotting that!):

https://github.com/Automattic/newspack-theme/issues/1466

PR #1461, opened 2021-08-04
Embeds: remove float on screens smaller than desktop

### LINKED ISSUE #1349: Fix Responsive Display of Crowdsignal Block
**Describe the bug**

When adding an embed via the Crowdsignal block, choosing left or right-aligned causes the Crowdsignal item to sit on top of the text. 

**To Reproduce**
1. Create a Crowdsignal poll (or you can use mine: https://poll.fm/10491859)
2. Create a new post with several blocks of text.
3. Add the Crowdsignal block and embed the link.
4. Set the Crowdsignal block to left align. 
5. Notice that on desktop it appears as expected - the poll displays to the left of the text. 
6. View the page on mobile and observe the poll on top of the text. 

This works as expected on a site with the Twenty Twenty One theme. The poll moves inline with the text and the text displays above and below. 

**Screenshots**

Screen Shot on 2021-05-04 at 18-16-59

**Expected behavior**
The poll should display as it does for non-Newspack sites.

laurelfulford:
### All Submissions:

This PR removes the float/max-width on embeds floated left and right on screens narrower than `1168px`. This is to prevent the embeds from getting cut off; though some behave responsively, since they're in iframes there's no way for us to "know" if content is being cut off, so this is a way to try to avoid that.

Closes #1349

1. Add or edit a post; add a couple embeds (Crowdsignal (demo poll https://poll.fm/10491859), Twitter, etc). Align these blocks left and right.
2. View on the front-end; shrink the browser window and note the embeds get cut off eventually (some, like Twitter, do react responsively first, but on very small screens they do get cut off):

image

image

4. Recheck the embeds; they should stop floating at pretty large screen sizes now, which should keep them from getting cut off on smaller screens:

image

image

### Other information:

### DIFF
--- a/newspack-theme/sass/mixins/_utilities.scss
+++ b/newspack-theme/sass/mixins/_utilities.scss
@@ -29,7 +29,7 @@
 		}
 	}
 
-	@if desktoponly == $res {
+	@if notdesktop == $res {
 		@media only screen and ( max-width: #{ $desktop_width - 1 } ) {
 			@content;
 		}
--- a/newspack-theme/sass/mixins/_utilities.scss
+++ b/newspack-theme/sass/mixins/_utilities.scss
@@ -29,6 +29,12 @@
 		}
 	}
 
+	@if desktoponly == $res {
+		@media only screen and ( max-width: #{ $desktop_width - 1 } ) {
+			@content;
+		}
+	}
+
 	@if desktop == $res {
 		@media only screen and ( min-width: $desktop_width ) {
 			@content;
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -46,13 +46,14 @@
 		}
 	}
 
-	&.wp-block-embed {
-		&.alignleft,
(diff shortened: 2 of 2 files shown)


newspack-theme/sass/mixins/_utilities.scss
```diff
@@ -29,6 +29,12 @@
 		}
 	}
 
+	@if desktoponly == $res {
```
adekbadek, line 32: Naming nit: this media query targets screen widths up to 1167px. I'm used to calling "desktop" sizes _above_ a certain threshold. For me this media query is rather "not-desktop".

adekbadek (approved):

newspack-theme/sass/mixins/_utilities.scss
laurelfulford: That's very true! I was kind of mindlessly following the pattern I used for the other two `min-size` queries (`mobileonly` and `tabletonly`); honestly the tablet one is even a bit of a stretch, but this one makes zero sense. 

I've updated it to `notdesktop` in fe0e808!

PR #1475, opened 2021-08-11
Add 'above' and 'below' widget areas to the theme

### LINKED ISSUE #1419: Add above & below header widget areas.
We've been installing our custom Header Widget Areas plugin on sites for about a year now, and there are a handful of these sites floating around. Especially with WP 5.8's widget blocks feature, I think it makes sense to officially add these widget areas into the theme. They would be widely useful. One main use would be: it would easily enable header ads for non-GAM ad providers, like Broadstreet, where the user can just add an ad block or widget in the header widget area.

laurelfulford:
### All Submissions:

This PR adds a new widget area above and below the header in the Newspack theme, with a couple exceptions:
* The 'below header' widget area is not included in Newspack Nelson. With the overlap, it's tricky to get the styles working nicely. I'm happy to revisit if we do see a lot of demand for it, but I until then, I'd like to keep things as simple as possible.
* Similar to the below header ads, if you use the simplified subpage header + the behind/beside image placements, the 'below header' widget is hidden via CSS, since the header overlaps the featured image.

For styles, I've tried to maximize the stuff we can do with the widget block editor, so I've done things like removed the top margin from the first width and bottom margin from the last widget - this way, it's easier to do things like add full-width group block with a background, and it will fill the space above the header. I would really appreciate some feedback whether or not this seems like a good idea, though, as it also means if you just add something like an ad block or image, you'll need to add spacer blocks, or a group block with a background as well, to make sure it doesn't touch the top/bottom of the screen. 😬 

**Note:** these widgets have different IDs than the ones added via plugin to a handful of Newspack sites. If we want to migrate folks off of that, we'll need to move the widget from the plugin-generated area to one of the theme areas, then deactivate the plugin. 

Closes #1419

1. Start with a theme other than Newspack Nelson. 
3. Navigate to Customizer > Widgets, and confirm you now have two new widget areas: Above Header and Below Header. 
4. Try adding a widget to the 'Above Header' widget area; I'd recommend using an image, or starting with a group block with a background. 
5. Make sure you can align the group block wide and full, and that if you do it takes up all available space; an image or cover block should work the same way:

image

6. Add a widget to the 'Below Header' widget space; view and make sure things look good -- it should touch the header:

image

7. Enable the subpage header  under Customizer > Header Settings > Simplified Subpage Header, and view posts with the featured image behind and beside; confirm that the below widget doesn't display:

image

8. Disable the subpage header, and confirm the below header widget area does display on posts using the behind and beside featured image placement:

image

9. Switch to Newspack Nelson; make sure the 'Below Header' widget space is not displayed, and is not available in the Customizer, either. 

### Other information:

### DIFF
--- a/newspack-nelson/functions.php
+++ b/newspack-nelson/functions.php
@@ -26,6 +26,15 @@ endif;
 add_action( 'after_setup_theme', 'newspack_nelson_setup', 12 );
 
 /**
+ * Function to remove the 'below header' widget area.
+ */
+function newspack_nelson_remove_widget() {
+	// Unregister the 'below header' widget area.
+	unregister_sidebar( 'header-3' );
+}
+add_action( 'widgets_init', 'newspack_nelson_remove_widget', 11 );
+
+/**
  * Function to load child theme's Google Fonts.
  */
 function newspack_nelson_fonts_url() {
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -308,6 +308,30 @@ function newspack_widgets_init() {
 
 	register_sidebar(
 		array(
+			'name'          => __( 'Above Header', 'newspack' ),
+			'id'            => 'header-2',
+			'description'   => esc_html__( 'Add widgets here to appear above the site header.', 'newspack' ),
+			'before_widget' => '<section id="%1$s" class="below-content widget %2$s">',
+			'after_widget'  => '</section>',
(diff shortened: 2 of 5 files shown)


miguelpeixe (approved):
It's looking great! Thank you for the detailed instructions!

laurelfulford:
Thanks for all the testing, @miguelpeixe! 🙌

PR #1489, opened 2021-08-19
Add a custom login screen based on theme's settings

thomasguillot:
### All Submissions:

This PR customise the login screen `wp-login.php` to kind of match the front-end and have a better branding.

1. It checks if the site has a custom logo (if not, the Newspack logo will replace the WordPress logo)
2. It checks if the site is using a custom background colour for the header. If so, it will use it to set the background colour of the login screen.

Does it make sense? 😅

default
Default

custom
Custom

1. Switch to this branch
2. Play with colours and logos
3. In a private window check `wp-login.php` and see if the correct changes are applied
4. Test with various child themes

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -1039,6 +1039,11 @@ require get_template_directory() . '/inc/customizer.php';
 require get_template_directory() . '/inc/logo-resizer.php';
 
 /**
+ * Custom Login Screen.
+ */
+require get_template_directory() . '/inc/login-screen.php';
+
+/**
  * Load Jetpack compatibility file.
  */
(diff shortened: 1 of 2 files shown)


laurelfulford (approved):
This is a great little customization bonus! 

I think the colour logic makes sense (or as much sense as it makes in the theme!): if it's based off of the header colour, there should be sufficient contrast against the logo. 

I looped through the different themes, and tried some different header/branding settings (default, custom, no header background; with and without logos), and it worked well! 👍

thomasguillot:
> if it's based off of the header colour, there should be sufficient contrast against the logo.

My thoughts exactly 😊

PR #1553, opened 2021-10-15
fix: move sticky ad down when sticky header is enabled

### LINKED ISSUE #1551: Remove `overflow: hidden` from the two-column template.
The theme currently uses `overflow: hidden` to make sure full-width blocks don't cause a slight side scroll (though I'm not sure if that's even still needed - see #1530).

This is interfering with sticky ads added to the sidebar; since these pages can't have full-width blocks, the overflow should be safe to remove at least from the two-column template.

laurelfulford:
### All Submissions:

This PR removes the `overflow: hidden` from the `.site-content` when you're using a two column template, so the sticky ad option will actually stick.

It also adds a bit of space on top of the sticky ad when the sticky header is enabled, so they don't run over each other :) 

Closes #1551

1. Start with a test site with Newspack Ads, and add an Ad Widget to the bottom of the sidebar. Check "Stick to the top"
2. View on the front-end and scroll down; note the widget doesn't stick.
4. Repeat step two; confirm that the ad now moves down the page with you.
5. Navigate to Customizer > Header Settings > Appearance and check 'Sticky Header'.
6. View the front-end again; confirm that both the ad and header stick, and the ad is an appropriate distance from the header. 

### Other information:

### DIFF
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -23,6 +23,16 @@
 	}
 }
 
+.post-template-default,
+.page-template-default:not( .newspack-front-page ),
+.archive,
+.search,
+.blog {
+	.site-content {
+		overflow: unset;
+	}
+}
+
 #primary {
 	margin: auto;
 	max-width: 90%;
--- a/newspack-theme/sass/plugins/newspack-ads.scss
+++ b/newspack-theme/sass/plugins/newspack-ads.scss
@@ -143,3 +143,8 @@ body {
 		}
 	}
 }
+
+// Move the sidebar sticky ad down a bit further when the header is sticky.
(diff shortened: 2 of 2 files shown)


laurelfulford:
Switching to draft briefly -- just realized I didn't take into account all of the heights the sticky header could be 😐

newspack-theme/sass/layout/_layout.scss
```diff
@@ -23,6 +23,16 @@
 	}
 }
 
+.post-template-default,
+.page-template-default:not( .newspack-front-page ),
+.archive,
+.search,
+.blog {
```
adekbadek, line 30: When is a `blog` class applied? I'm wondering is the `overflow` should rather be _set_ for specific cases, than _unset_.

Issue #1549, opened 2021-10-13
Remove unneeded CSS that AMP isn't stripping out


laurelfulford:
When working on a recent site with a lot of Custom CSS, I noticed there are a number of things not getting stripped out by AMP, even though they're not used. This is because they include things like `@keyframes`, `@font-face`, or CSS Variables. 

Specifically on this site, this list included: 

- [x]  mediaelement-css~~ - (304b) - to be removed in #1548.
- [x]  wc-blocks-vendors-style-css (198b) - optionally removed from home in #1555
- [x]  wc-blocks-style-css (1973b) - optionally removed from home in #1555
- [ ]  newspack-listings-styles-css (460b)
- [x]  woocommerce-layout-css (285b) - optionally removed from home in #1555
- [x]  woocommerce-smallscreen-css (328b) - optionally removed from home in #1555
- [x]  newspack-woocommerce-style-css (1381b) - to be simplified in #1550, and optionally removed from homepage in #1559
- [ ]  social-logos-css (921b) - Jetpack (widget not used on site)

Other sites have included 
- [ ] CSS from Gravity Forms (amount TK)

Total from this site: 4607b, or a little over 6% of the allowed CSS -- it's not a ton, but it adds up when the site also has a lot! 

I imagine this varies from site to site, depending on what plugins they have installed.

This is a known issue in the [AMP plugin repo](https://github.com/ampproject/amp-wp/issues/4095),  but I think we might be able to do some things in the theme that could help for the short term -- things like disabling WooCommerce styles in all but the WooCommerce pages, and optionally turning off the WC block styles if a site isn't using them (especially on the homepage). We may also be able to check if the Jetpack Social Links widget is in use before enqueuing those styles.

laurelfulford:
Closing this task for now; the AMP plugin is now allowing CSS overages, and for most sites, the problematic CSS varies, and is coming from different plugins.

PR #1555, opened 2021-10-18
feat: add option to dequeue some WooCommerce styles on home

laurelfulford:
### All Submissions:

This PR adds an option to dequeue the WooCommerce block, layout and small screen CSS from the site's homepage -- AMP removes most of these styles through tree-shaking, but a bit of each file is left over because they contain things like animations and CSS variables, or classes that exist on the homepage (like `.screen-reader-text`).

The purpose is to buy a little more wiggle room for CSS on the homepage. It's a little hacky/clunky but I also wanted to avoid possibly removing styles when they may be needed on subpages -- I don't want to introduce a weird edge case, like removing styles when they may be 'expected' by third-party WC plugins. That being said, I'm definitely open to suggestions for other approaches! 

In my testing, this should free up around 3.5-4% of the available CSS on the homepage. 

See #1549

1. Start with a test site with AMP and WooCommerce enabled.
2. When viewing the homepage, navigate to AMP > CSS Usage. 
3. Click though the CSS being loaded; note that the WC Blocks, Small Screen and Layout styles are loaded, even though they aren't needed on the homepage: 

image

image

4. Apply the PR.
5. Navigate to Customizer > WooCommerce; there should now be a new Advanced Settings panel. 
6. Open the panel and check the two options to dequeue styles on the homepage, then click 'Publish'.

image

7. Navigate back to AMP > CSS Usage from the homepage, and click 'Recheck'.
8. Confirm that the block, small screen and layout stylesheets are no longer loaded on the homepage. 

### Other information:

### DIFF
--- a/newspack-theme/inc/customizer.php
+++ b/newspack-theme/inc/customizer.php
@@ -1317,6 +1317,51 @@ function newspack_customize_register( $wp_customize ) {
 			'section'     => 'woocommerce_thank_you',
 		)
 	);
+
+	/**
+	 * WooCommerce Advanced Settings
+	 */
+	$wp_customize->add_section(
+		'woocommerce_advanced',
+		array(
+			'title' => esc_html__( 'Advanced Settings', 'newspack' ),
+			'panel' => 'woocommerce',
+		)
+	);
+
+	// Dequeue WooCommerce block CSS on the homepage.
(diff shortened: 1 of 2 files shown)


laurelfulford:
Thanks @dkoo! 🥳

PR #1579, opened 2021-11-13
Update es_ES.po

SergioRodriguezSanchez:
Suggestion of spanish translation for the byline if the post have two or more author. ' and ' to ' y '.

### All Submissions:

### Changes proposed in this Pull Request: Update translation in spanish for more than one author.

Closes # .

### Other information:

### DIFF
--- a/newspack-theme/languages/es_ES.po
+++ b/newspack-theme/languages/es_ES.po
@@ -1060,6 +1060,10 @@ msgstr "Última actualización"
 msgid "by"
 msgstr "por"
 
+#: inc/template-tags.php:121
+msgid " and "
+msgstr " y "
+
 #. translators: %s: Name of current post. Only visible to screen readers.
 #: inc/template-tags.php:229
 msgid "Leave a comment<span class=\"screen-reader-text\"> on %s</span>"
--- a/newspack-theme/languages/es_ES.po
+++ b/newspack-theme/languages/es_ES.po
@@ -974,7 +974,7 @@ msgstr "Cerrar"
 #. translators: separates last two names; needs a space on either side.
 #: inc/newspack-sponsors.php:137 inc/template-tags.php:121
 msgid " and "
-msgstr ""
+msgstr " y "
 
 #. translators: separates all but the last two names; needs a space at the end.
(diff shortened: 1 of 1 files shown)


newspack-theme/languages/es_ES.po
```diff
@@ -1060,6 +1060,10 @@ msgstr "Última actualización"
 msgid "by"
 msgstr "por"
 
+#: inc/template-tags.php:121
```
laurelfulford, line 1063: The `" and "` string already exists further up the file, [here](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/languages/es_ES.po#L974-L977), so there's no need to add the file path and msgid again, only the translation needs to be added on line [977](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/languages/es_ES.po#L977).

laurelfulford (changes requested):
Thanks for submitting this fix, @SergioRodriguezSanchez -- great catch!

It looks like a spot for this translation already exists in the es_ES.po file, it just has an empty string for the translation -- I've added a note for where that is! 

Also, to have the translation work correctly, we'll need to [generate a new es_ES.mo file](https://stackoverflow.com/questions/4499622/how-can-i-create-an-mo-file-for-wordpress) from your new .po file -- do you have ability to do that? If not, I'm happy to take care of that step and loop in another member of the team to review. 

Just let me know if you have any questions at all -- thanks again!

newspack-theme/languages/es_ES.po
SergioRodriguezSanchez: Hi @laurelfulford thanks for the review. It is already changed in line 977 and deleted the previous one, as you requested. If you could generate the 'mo' file will be great as I am not used to do that. Thanks a lot!

laurelfulford (approved):
Thank you, @SergioRodriguezSanchez - this looks good! 👍 

I'll approve and merge this change, and then follow up with a separate PR to update the es_ES.mo file.

SergioRodriguezSanchez:
> Thank you, @SergioRodriguezSanchez - this looks good! 👍
> 
> I'll approve and merge this change, and then follow up with a separate PR to update the es_ES.mo file.

Thank you so much!

Issue #1138, opened 2020-11-10
French translation


philipjohn:
**Is your feature request related to a problem? Please describe.**
We have a Morrocan publisher who has their site set to Français de Belgique but we don't have a translation for that so some items like bylines aren't translated.

**Describe the solution you'd like**
The theme should provide a Français de Belgique translation. 

**Describe alternatives you've considered**
Not sure there's another way to be honest!

PR #1596, opened 2021-11-29
feat(i18n): add polish translations

adekbadek:
### All Submissions:

Adds polish translations. 

1. Switch your site's language to polish, observe the Newspack Theme strings are in polish

### Other information:


laurelfulford (changes requested):
Thanks for adding these translation, @adekbadek! 

One weird issue with the theme is that the translation files aren't picked up by the editor unless the MD5 hash format is changed to a specific naming convention -- I don't know why this differs from how the block translations work, but checking things like the Featured Image panel, it looks like the English labels are still being used. 

The format that I found that works is `newspack-[language]-newspack-[original-file-name].json`

For the translating extend-featured-image-editor.js, it's currently called `pl_PL-3b89b516cbaada3f50d73c34c0c88e3e.json` -- to get the editor to pick it up, it should be changed to `newspack-pl_PL-newspack-extend-featured-image-script.json`

Just let me know if you have any questions about this! I figured out how to get it to work early on though Googling and trial and error, but it does seem strange that it works differently.

adekbadek:
> One weird issue with the theme is that the translation files aren't picked up by the editor unless the MD5 hash format is changed to a specific naming convention

I noticed that in the editor, but forgot to mention it. Fixed in 8e26c12b401277c410219573c84fc3aa9062d486

I've also noticed that some JSON files have `-script` suffix, and some use the file from `src`, some the file from `dist` as the `source`. I've found it to be a problem with `fr_BE` translation, fixed in 26b052ca49426e5f3d86784f43f493800b850c0d.

As a sidenote, why do we have Belgian French translation and not French? AFAIK the difference is small and a `fr_FR` translation would serve a larger audience. 

> I noticed the by on single posts has been translated to -  […] though if it's very unusual to have byline prefaced by something like 'by' in Polish, swapping it with a symbol or even an empty string seems like a good alternative!

Exactly – there's no good Polish equivalent for a short "by".

laurelfulford (approved):
Thanks for fixing the editor translation, and for catching that issue in the fr_BE files, @adekbadek! The translations are now loading as expected in the editor :shipit: 

Thank you also for entertaining my curiosity about the byline translation :) 

> As a sidenote, why do we have Belgian French translation and not French? AFAIK the difference is small and a fr_FR translation would serve a larger audience.

That's a good question! It was a [specific request from a publisher](https://github.com/Automattic/newspack-theme/issues/1138), but IIRC @thomasguillot mentioned if we did ever need a French translation, we could use the exact same strings (please correct me if I'm wrong, Thomas!). 

I can spin up a new issue to look into switching the file names if that is the case -- as I understand it, WordPress should fall back to the fr_FR translation if a fr_BE translation isn't available, but I can confirm for sure before making any changes!

thomasguillot:
> That's a good question! It was a [specific request from a publisher](https://github.com/Automattic/newspack-theme/issues/1138), but IIRC @thomasguillot mentioned if we did ever need a French translation, we could use the exact same strings (please correct me if I'm wrong, Thomas!).
> 
> I can spin up a new issue to look into switching the file names if that is the case -- as I understand it, WordPress should fall back to the fr_FR translation if a fr_BE translation isn't available, but I can confirm for sure before making any changes!

Yep, I guess we could simply rename the file.

laurelfulford:
Thanks for confirming, @thomasguillot! I spun up issues for both the [theme](https://github.com/Automattic/newspack-theme/issues/1600) and [blocks](https://github.com/Automattic/newspack-blocks/issues/951) to make this change.

PR #1598, opened 2021-11-30
Correct checks before adding the `no-sidebar` and `has-sidebar CSS classes

### LINKED ISSUE #1595: Add a class or ID to the main-content <div> or <main> related to the template that's being used
There isn't a way to add or remove or modify items via CSS based on the template a post is using (default, one column, one column wide)

A simple but effective way, is to add a class or ID to the main-content <div> that describes the template that's being used. 

Since there is no way I can identify which template the post is being used from a CSS point of view, there is no alternative.

laurelfulford:
### All Submissions:

Right now the theme adds the CSS classes `has-sidebar` or `no-sidebar`, depending on whether or not widgets are assigned to the sidebar widget space. 

However, the logic adding this class doesn't check if the current post or page is using a template that _shows_ the sidebar, so the class `has-sidebar` could be applied when none are present. 

This PR improves the logic around adding the `has-sidebar` class, so it's only added when sidebar widgets have been added AND the current post/page/archive has the right settings to actually display the sidebar. Otherwise, if there are no sidebar widgets assigned and/or the sidebar widgets are not visible, the class `no-sidebar` should be added. 

Closes #1595

1. Apply the PR.
2. Navigate to Customizer > Widgets > Sidebar, and add at least one widget to the sidebar area if there aren't any already.
3. View the static front page and inspect elements; find the body tag and confirm it has the CSS class `no-sidebar`.
4. View a post or page with the default template; confirm that it has the body class `has-sidebar`. 
5. View a post or page with the one column and/or the one column wide template. Confirm that it has the body class `no-sidebar`. 
6. View an archive page; confirm that is has the body class `has-sidebar`.
7. Navigate to Customizer > Template Settings > Archive Settings; under 'Archive Layout' select either One Column or One Column wide.
8. Return to an archive page and inspect elements; find the body tag and confirm it now has the CSS class `no-sidebar`.
9. Navigate back to Customizer > Widgets > Sidebar and remove your sidebar widget. 
10. Double check a post or page with the default template, and confirm that it now has the CSS class `no-sidebar`. 

### Other information:

### DIFF
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -148,9 +148,14 @@ function newspack_body_classes( $classes ) {
 		$classes[] = 'has-highlight-menu';
 	}
 
-	// Adds a class of has-sidebar when there is a sidebar present.
-	if ( is_active_sidebar( 'sidebar-1' ) && ! ( is_front_page() && 'posts' !== get_option( 'show_on_front' ) ) ) {
+	// Adds a class of has-sidebar when there is a sidebar present and populated.
+	if ( is_active_sidebar( 'sidebar-1' )
+		&& ( ( ! is_archive() && ! is_page_template() && ! ( is_front_page() && 'posts' !== get_option( 'show_on_front' ) ) )
+		|| ( is_archive() && 'default' === get_theme_mod( 'archive_layout', 'default' ) ) )
+	) {
 		$classes[] = 'has-sidebar';
+	} else {
+		$classes[] = 'no-sidebar';
 	}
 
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works as described! This sent me down a rabbit hole of seeing whether there's a way to programmatically check whether a sidebar is rendered on-screen (not just whether it's active and/or has widget content). Apparently there isn't, so this is the most elegant solution.

laurelfulford:
Thanks for the testing @dkoo -- and for digging around a bit more with this, I had a few runs at it trying to find a nicer if/else statement, so I appreciate the second set of eyes on that bit!

PR #1661, opened 2022-01-15
Update figcaption styles for WordPress 5.9

### LINKED ISSUE #1657: Figcaptions not correct width in WP 5.9 in the editor
In the editor, the captions are no longer displaying at 100% on their own; they are as wide as they need to be to match their contents: 

image

Adding a width of 100% to the editor styles should fix.

laurelfulford:
### All Submissions:

The Image block figcaptions styles are changing in WP 5.9, and they are causing a minor display error in the editor preview. This PR tweaks the editor styles to fix that.

Closes #1657

1. Start with a test site running the latest 5.9 RC - you can do this using the [WordPress Beta Tester](https://en-ca.wordpress.org/plugins/wordpress-beta-tester/) plugin.
2. Create a new post and add some image blocks with different alignments (none, centred, left, right, wide, full). Add short captions to each block.
3. Note when the block has no alignment, the caption appears to be centred (if you keep adding text to make it longer, it will eventually fill the whole space and appear left-aligned). When the block has no alignment, or a wide or full alignment, the caption seems extra close to the image:

image

5. Confirm that the captions now preview as expected in 5.9:

image

6. Test the PR in the same fashion on a site still running 5.8.X and confirm it causes no major issues -- it _will_ cause the space above the caption to increase in some cases (like when the image is not aligned), but this seems like an acceptable trade-off for the short-term:

image

### Other information:

### DIFF
--- a/newspack-theme/sass/style-editor-base.scss
+++ b/newspack-theme/sass/style-editor-base.scss
@@ -126,6 +126,8 @@ figcaption,
 	line-height: $font__line-height-pre;
 	margin: 0 auto #{0.5 * $size__spacing-unit};
 	max-width: 780px;
+	padding-left: 0;
+	padding-right: 0;
 }
 
 /** === Post Title === */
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Looks good in both 5.8 and 5.9!

laurelfulford:
Thanks @dkoo! 🙌

PR #1689, opened 2022-01-27
Fix: prevent odd wrapping with multiple categories

### LINKED ISSUE #1638: Categories can wrap oddly due to `&nbsp`
[The theme uses `&nbsp;` in the category separators](https://github.com/Automattic/newspack-theme/blob/d1af3271d899a5a7183c65c7bd54a3b4060b74eb/newspack-theme/inc/template-tags.php#L258), which can cause odd display issues when you have several categories applied to a post -- they can't "figure out" where they wrap, and breaks end up in multi-word categories if any exist: 

image

It should be safe to replace this with a regular space, but testing is required to make sure that doesn't introduce new issues.

laurelfulford:
### All Submissions:

Right now the category list markup includes `&nbsp;` after the comma separators, which causes an odd wrapping issue when you have a lot of categories displaying.

There isn't really a need for a non-breaking space there, so this PR replaces that with a regular space. 

Closes #1638

1. Switch your site to a theme other than the Newspack default theme (which hides the separators, thereby hiding this issue).
2. Navigate to Customizer > Template Settings > Post Settings, and turn off "Use Yoast's primary category functionality"
3. Create a post and add several categories -- make sure at least one is more than one word, so it contains its own space. 
4. Publish; view the post on the front-end, and make your browser window narrow; note that your categories list has some funky line breaks, possibly running off the screen. This is because the breaks are only happening inside of categories that have spaces, not between the different categories:

image

5. Apply the PR.
6. Confirm that the funky line breaks are gone, and that your categories seem to be breaking over multiple lines in a logical fashion:

image

### Other information:

### DIFF
--- a/newspack-theme/inc/template-tags.php
+++ b/newspack-theme/inc/template-tags.php
@@ -255,7 +255,7 @@ if ( ! function_exists( 'newspack_categories' ) ) :
 
 		if ( ! $categories_list ) {
 			/* translators: used between list items; followed by a space. */
-			$categories_list = get_the_category_list( '<span class="sep">' . esc_html__( ',', 'newspack' ) . '&nbsp;</span>' );
+			$categories_list = get_the_category_list( '<span class="sep">' . esc_html__( ',', 'newspack' ) . ' </span>' );
 		}
 
 		if ( $categories_list ) {


adekbadek (approved):

laurelfulford:
Thanks @adekbadek!

PR #1694, opened 2022-01-28
fix: update editor gradient styles

### LINKED ISSUE #1673: Theme-driven gradients preview incorrectly in the editor
Discovered while testing 5.9, but it happens in 5.8 as well:

If you pick either of the gradients generated from the Primary or Secondary colours, they will preview using the Newspack defaults in the editor (blue, and grey), regardless what custom colours you've picked in the Customizer.

laurelfulford:
### All Submissions:

This PR corrects the CSS selectors used in the editor preview for the theme's baked-in gradient options -- as is, it works on some blocks (buttons) but not others (groups). 

Closes #1673

1. If your test site isn't already using custom colours, navigate to Customizer > Colours, and change both the Primary and Secondary colours.
2. Edit a post or a page and add two group blocks, and two buttons; set each to use gradients as backgrounds, applying each of the theme's baked in options created from the primary colour and its variation, and the secondary colour and its variation:

image

3. Note that the preview of the buttons is correct, but the group blocks use the theme's default blue and grey in the generated gradients, instead of your chosen colours; if you publish you page, though, the gradients will display correctly:

image

4. Apply this PR.
5. Refresh the editor and confirm that your gradients are now previewing correctly on the group block, without negatively impacting the buttons:

image

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -619,10 +619,10 @@ function newspack_custom_colors_css() {
 		}
 
 		/* Set gradients */
-		.edit-post-visual-editor.editor-styles-wrapper .has-grad-1-gradient-background {
+		.editor-styles-wrapper .wp-block.has-grad-1-gradient-background {
 			background-image: linear-gradient( 135deg, ' . esc_html( $primary_color ) . ' 0%, ' . esc_html( newspack_adjust_brightness( $primary_color, -40 ) ) . ' 100% );
 		}
-		.edit-post-visual-editor.editor-styles-wrapper .has-grad-2-gradient-background {
+		.editor-styles-wrapper .wp-block.has-grad-2-gradient-background {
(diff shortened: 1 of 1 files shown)


dkoo:
Would this warrant rebasing against `release` as a hotfix?

laurelfulford:
@dkoo I don't think this is something that requires a hotfix (unless I'm misunderstanding the cases where we do do hotfixes, which I totally might be!).

It's an odd editor preview issue, so it wouldn't affect the front-end -- it's also something I noticed on my own (rather than it being reported) so I'm not sure if anyone's actually run across this yet.

adekbadek (commented):
In my testing, this resolves the issue for group block, but the buttons are affected both on `master` and on this branch:

image

laurelfulford:
Thanks @adekbadek! I'm at a bit of a loss as to how I got my original results -- when I tested this again, I also saw the same issue with the buttons 😕 

The issue happening with both blocks definitely makes more sense, though -- I've updated the selector and it seems to fix this for both now!

adekbadek (approved):

laurelfulford:
Thanks Adam! 🙌

PR #1696, opened 2022-01-29
Make sure full-width group block contents can't touch edges

### LINKED ISSUE #815: Align-full Group + no background: Contents hit edges on mobile
When you have a group block that's full width, with no background, the blocks inside can hit the edges on mobile.

This is because the Group block only gets padding when there's a background colour, but the styles (for full width blocks) expect it to always be there.

Adding a padding for smaller screen sizes could help, or there might be another way to fix (margins on the nested blocks?). The trick will be making sure the nested blocks line up on the edges with the rest of the content.

laurelfulford:
### All Submissions:

If you have a full-width group block with no background or padding set, its content will hit the edge of the browser window on smaller screens. 

This PR adds some padding in that case, that shouldn't affect when the block has a background or borders, and that should be over-ride-able with the block's padding options.

Closes #815.

1. Create a new post or page and switch it to the One Column or One Column Wide template.
2. Add four group blocks, each containing a paragraph, and set them all to full-width; give one padding, one a background, one a border style and leave one as the default. Alternatively, copy [this example code](https://github.com/Automattic/newspack-theme/files/7962717/group-block.txt) into the code editor.
3. View on the front-end, and shrink down the browser window; note how the paragraph in the default group block hits the edge of the window: 

image

image

5. Confirm the paragraph no longer hits the edge and is roughly lined up with the post title. The other group blocks are not affected (they pick up their inline padding, or the padding attached to their background/border styles):

image

image

6. Switch to the other one column template and confirm there are no issues.

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1680,6 +1680,14 @@ hr {
 		}
 	}
 
+	// Make sure content in a full-width group block doesn't touch edges
+	.entry .entry-content > .wp-block-group {
+		&.alignfull:not( .has-background ):not( .is-style-border ) {
+			padding-left: 5.5%;
+			padding-right: 5.5%;
+		}
+	}
+
 	// keep nested align-full elements from overflowing the container.
 	.entry .entry-content {
 		.wp-block-column,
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
👌

PR #1699, opened 2022-01-31
feat: full-width scaip ad

miguelpeixe:
### All Submissions:

Make SCAIP ads full-width on one-column and one-column wide templates and apply a default background color. This background color should be overwritten by the **Colors -> Ads Background Color** in the Customizer.

image

Closes https://github.com/Automattic/newspack-ads/issues/295

1. With SCAIP installed and configured with ad units, create a One-column wide post with multiple paragraphs
2. Check out this branch and confirm the ad placement area is full-width and with a light gray background
3. Through the Customizer, visit **Colors -> Ads Background Color** and change it to a random color
4. Confirm the custom color is applied to the SCAIP ad as well

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1703,6 +1703,13 @@ hr {
 			margin-left: calc( 50% - 50vw );
 			margin-right: calc( 50% - 50vw );
 			max-width: 100vw;
+			> * {
+				margin-top: $size__vertical-rhythm * 2;
+				margin-bottom: $size__vertical-rhythm * 2;
+			}
+			amp-ad {
+				display: block;
+			}
 		}
 	}
 }
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1698,7 +1698,7 @@ hr {
 	// Make SCAIP placements full-width.
 	.entry .entry-content .scaip {
 		position: relative;
-		background: #eee;
+		background: rgba( 0, 0, 0, 0.05 );
 		max-width: none;
 		width: 100vw;
 		margin-left: -50vw;
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1694,6 +1694,16 @@ hr {
 			}
 		}
 	}
+
+	// Make SCAIP placements full-width.
+	.entry .entry-content .scaip {
+		position: relative;
+		background: #eee;
+		max-width: none;
+		width: 100vw;
+		margin-left: -50vw;
+		left: 50%;
+	}
 }
 
 //! Social Links block
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1697,12 +1697,13 @@ hr {
 
 	// Make SCAIP placements full-width.
 	.entry .entry-content .scaip {
-		position: relative;
-		background: rgba( 0, 0, 0, 0.05 );
-		max-width: none;
-		width: 100vw;
-		margin-left: -50vw;
-		left: 50%;
+		.newspack_global_ad,
+		.widget_newspack-ads-widget {
+			background-color: $color__background-pre;
+			margin-left: calc( 50% - 50vw );
+			margin-right: calc( 50% - 50vw );
+			max-width: 100vw;
+		}
 	}
 }
 
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -459,6 +459,8 @@ function newspack_custom_colors_css() {
 	// Set ads background color
 	if ( 'default' !== get_theme_mod( 'ads_color', 'default' ) ) {
 		$theme_css .= '
+			.site .entry .entry-content .scaip .newspack_global_ad,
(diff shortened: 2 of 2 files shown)


newspack-theme/sass/blocks/_blocks.scss
```diff
@@ -1694,6 +1694,16 @@ hr {
...
 	}
+
+	// Make SCAIP placements full-width.
+	.entry .entry-content .scaip {
+		position: relative;
+		background: rgba( 0, 0, 0, 0.05 );
+		max-width: none;
+		width: 100vw;
+		margin-left: -50vw;
+		left: 50%;
```
laurelfulford, line 1705: These styles look like they get applied to all SCAIP-inserted widgets, not just ads - I know this used to be a think more publishers used prior to campaigns taking off, but we still might want to avoid non-ad widgets. In that case, something like this should work:

```suggestion
	.entry .entry-content .scaip .widget_newspack-ads-widget {
		background: $color__background-pre;
		margin-left: calc( 50% - 50vw );
		margin-right: calc( 50% - 50vw );
		max-width: 100vw;
	}
```

(I also made a couple tweaks for consistency -- using the same light grey Sass variable as in other spots, and the same styles to make the ad full width as we use for the full-width blocks!). 

This _will_ break the custom colours styles -- they'll also have to be updated to be more specific in the inc/color-patterns.php file.

laurelfulford (changes requested):
Thanks for this update, @miguelpeixe!

I added some feedback around the element to target -- I'm not sure that these styles should be applied to everything in a SCAIP space, or just ads, but my suggestion targets just ad widgets. Just let me know if I'm misunderstanding how this should work! 

Also out of curiosity, should ads in the default template also get the same light grey background? They do use the custom ad backgrounds if applied, but have a white background otherwise. 

Just let me know if you have any questions about my comments!

miguelpeixe:
Thank you for the review, @laurelfulford! 

> Also out of curiosity, should ads in the default template also get the same light grey background?

I don't think so, I believe the gray background is suitable for full-width ad placements because it creates a subtle separator to allow an ad with any width to live inside it without disrupting the surrounding grid.

> I know this used to be a think more publishers used prior to campaigns taking off, but we still might want to avoid non-ad widgets.

Good call! I've updated your suggestion to also include the non-legacy SCAIP widgets that are created through the Ads Placements system. 2945093806fc79b783a74874a5c2866acc0edcbf

> This will break the custom colours styles -- they'll also have to be updated to be more specific in the inc/color-patterns.php file.

6bc040400cca5e2c039b05186ae02c90a3e64065 adjusts the rules specificity in `inc/color-patterns.php`

newspack-theme/sass/blocks/_blocks.scss
```diff
@@ -1694,6 +1694,17 @@ hr {
...
+
+	// Make SCAIP placements full-width.
+	.entry .entry-content .scaip {
+		.newspack_global_ad,
+		.widget_newspack-ads-widget {
+			background-color: $color__background-pre;
+			margin-left: calc( 50% - 50vw );
+			margin-right: calc( 50% - 50vw );
+			max-width: 100vw;
+		}
```
laurelfulford, line 1706: One more nitpick (sorry I missed this before 🤦‍♀️): I noticed when the default colour is used, the ad touches the top and bottom of the space -- this could be because of the suggestions I made before! 

AMP: 

image

No AMP:

image


This tweak will add the same padding as in the custom colour settings for the ads, and the `vertical-align: middle;` will remove a little extra space that's getting added to the bottom of the `amp-ad` (it's what's adding the space in the above screenshot):

```suggestion
                        > * {
                               padding: 8px;
                       }

                       amp-ad {
                               vertical-align: middle;
                       }
		}
```

Just let me know if you have questions about either of these!

miguelpeixe:
I didn't notice that too! It appears to be exclusive to the legacy mode through widgets. When using Newspack Ads placements, it gets a margin defined through [a rule at `.newspack_global_ad > *`](https://github.com/Automattic/newspack-theme/blob/feat/full-width-ad/newspack-theme/sass/plugins/newspack-ads.scss#L4-L12).

3907ed7e59cbd9b7122c34874b275b6d4cba2a0b replicates the same styles to the new selectors and adds `display: block;` to `<amp-ad />` so it doesn't render that extra margin at the bottom. WDYT?

laurelfulford (approved):
Thanks @miguelpeixe! 

The reason I'd suggested the padding instead of margins was to match the styles that get added when you have an ad with a background colour -- with both a margin and a padding being added by that, you get more space with a custom background than without (16px with the default colour and 24px with  a custom colour):

image

image

That being said, I find the [ad colour styles overall a bit confusing](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/inc/color-patterns.php#L469-L483), and it's really the only spot where we're adding any non-colour-related styles to that color-patterns.php, so it might make more sense to address making the styles more consistent overall in a separate PR where that can be addressed, too :) 

I don't think the discrepancy in spacing is a blocker for this landing, especially since in either case the spacing around the ads would be consistent across a site based on their settings, just not consistent between sites with different settings. 

Marking as 👍 from me, but just let me know if you have any questions!

miguelpeixe:
I see what you mean. Indeed it is a bit confusing, the styles are applying `margin` to `.newspack_global_ad` and `padding` to `.widget_newspack-ads-widget`. They are likely to behave a bit differently until they are normalized.

Thank you for the review! I'll merge this PR so this can be tackled in a different PR.

PR #1707, opened 2022-02-13
Make sure styles are inherited correctly when using Pico

### LINKED ISSUE #1656: Aligned block's specific styles break with Pico
Right now the theme styles a number of blocks as direct decedents of `.entry-content` -- specifically when they have particular alignments, like left, right, or wide. 

This breaks when Pico is enabled on a site, because the plugins wrap the content in an additional div.

These styles should be updated so they work when Pico is used on a site.

laurelfulford:
### All Submissions:

A lot of the block styles in the theme rely on the theme's structure -- they 'need' blocks to be direct decendents of the `.entry-content` class to apply styles like `.alignfull` and `.alignleft`. The structure also helps set the vertical alignment. 

Pico wraps the content in an extra div with the ID `#pico`, which causes these styles not to work, and some blocks -- like cover and group -- don't have the correct vertical spacing. 

This PR adds the `#pico` selector to the mix, making sure the styles work as expected when that plugin is used. 

Closes #1656

1. Start on a test site running Pico (we have one internally -- please ping me if you need a hand tracking it down to test!), and create a page using the One Column template. 
2. Add a mix of blocks to the page that have a mix of alignments: full, wide, left, right...
3. View on the front-end -- note how the blocks are not respecting their alignments. Wide and Full aligned blocks are no wider than the content area, and while the left and right-aligned blocks work, they don't "pop out" of the content like they should using the one column template:

image

image

4. Apply this PR to the Pico test site. To do this, I applied this PR to my local test site, and ran `npm run release:archive` to generate a zipped copy of the theme. Then I unzipped the Newspack parent theme, renamed it in the style.css file to distinguish it from the default on the Pico test site, rezipped it and uploaded it. 
5. View your test post and confirm that the blocks are using the correct alignments:

image

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1,7 +1,8 @@
 /* !Block styles */
 
 .entry .entry-content > *,
-.newspack-inline-popup > * {
+.newspack-inline-popup > *,
+#pico > * {
 	margin: 32px 0;
 	max-width: 100%;
 
@@ -373,7 +374,8 @@ p.has-background {
 }
 
 // Temporarily make some column styles more specific
-.entry .entry-content > .wp-block-columns {
+.entry .entry-content > .wp-block-columns,
+#pico > .wp-block-columns {
 	&.is-not-stacked-on-mobile {
 		margin-left: -16px;
 		margin-right: -16px;
@@ -1010,9 +1012,12 @@ hr {
 	}
 }
 
-.entry .entry-content > .wp-block-separator,
-.entry .entry-content > hr {
-	margin: ( 2 * $size__spacing-unit ) auto;
+.entry .entry-content,
+#pico {
+	> .wp-block-separator,
+	> hr {
+		margin: ( 2 * $size__spacing-unit ) auto;
+	}
 }
 
 //! Twitter Embed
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
Works as described! 👍 Also tested inline prompts since I see some styles for those here.

One question: I saw another instance of a `.entry-content >` selector in the [Nelson theme](https://github.com/Automattic/newspack-theme/blob/master/newspack-nelson/sass/style.scss#L211-L219). Do we need to also add the `#pico` selector here? The styles seem to prevent a group block with `.alignfull` from being full-width if it's the very first block on a homepage with Nelson.

laurelfulford:
> One question: I saw another instance of a .entry-content > selector in the [Nelson theme](https://github.com/Automattic/newspack-theme/blob/master/newspack-nelson/sass/style.scss#L211-L219). Do we need to also add the #pico selector here? The styles seem to prevent a group block with .alignfull from being full-width if it's the very first block on a homepage with Nelson.

Oh, good catch @dkoo! This should now be fixed. 

When I started looking at this on the testing site, I realized NONE of the wide styles were working on the front-page :| It seems the original CSS is just structured in a weird way, and using and ID, rather than a class (like `.entry-content`) the selector for Pico became stronger and it wasn't inheriting the same way as the existing styles. 

Rather than trying to rejig existing CSS (and potentially introducing issues), I also swapped the `#pico` selector with `[id='pico']`, and that seems to have the same 'power' as a CSS class, instead of overpowering like an ID. 

Just let me know if you have any questions about this!

dkoo (approved):
> Rather than trying to rejig existing CSS (and potentially introducing issues), I also swapped the `#pico` selector with `[id='pico']`, and that seems to have the same 'power' as a CSS class, instead of overpowering like an ID. 

Good call! The `#` ID selector does tend to be a bit overly strong sometimes. The homepage is looking good, as well as the test page and similar content on a post, too.

laurelfulford:
Thanks @dkoo!

PR #1717, opened 2022-02-24
fix: remove ads custom media queries

miguelpeixe:
### All Submissions:

With the recent implementation of #1699, this custom media query control of legacy SCAIP ads is preventing wider units to be displayed.

I couldn't identify a reason to keep this filter as it is only being applied to legacy (widget) SCAIP ads and its purpose is not clear to me. The native implementation of media queries should handle the different viewports without the need for this customization.

1. On the master branch make sure you have SCAIP in legacy mode, AMP active and AMP Plus off
2. Add a valid ad unit to the first SCAIP position containing the following sizes: `300x250`, `728x90`, `970x250`
3. Visit a post with a wide template and observe you are not able to render the `970x250` size
4. Check out this branch, refresh the post page and confirm you are able to render the `970x250` size

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -1015,63 +1015,6 @@ function newspack_coauthors_in_rss( $the_author ) {
 add_filter( 'the_author', 'newspack_coauthors_in_rss' );
 
 /**
- * Determine minimum theme breakpoint, below which responsive Ads media queries should not have a min-width.
- *
- * @param array  $media_queries An array of objects, one for each size, with width|height|min_width|min_height.
- * @param string $placement ID of the ad placement.
- * @param string $context Optional second string describing the ad placement. For Widget placements, the ID of the Widget.
- * @return array The correct array of media query data.
- */
-function newspack_theme_newspack_ads_media_queries( $media_queries, $placement, $context ) {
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
I dug through the `blame` for this function and found this context: https://github.com/Automattic/newspack-ads/pull/31

I think now that we have full-viewport-width SCAIP ads on one-column and one-column-wide templates, plus true responsive ads in SCAIP placements, this works well for those templates. However, does it also work for the default template? I couldn't verify because I couldn't get a 970x250 or 728x90 ad unit to actually render in my test site, but I want to make sure that the responsive ads are intelligent enough to render an appropriate size based on the ad's container, not just the viewport.

dkoo (approved):
Update: I was able to get this working on a fresh site, and the correct size rendered in the default/one-column (729x90) and one-column-wide (970x250) templates.

miguelpeixe:
> I was able to get this working on a fresh site, and the correct size rendered in the default/one-column (729x90) and one-column-wide (970x250) templates.

This was probably by chance. We are not taking the container width into account for determining which ad should be rendered. This can be achieved by using the recently implemented `template` targeting key values to target creatives for wide templates.

miguelpeixe:
The `template` targeting key has already been tackled at https://github.com/Automattic/newspack-ads/pull/297

dkoo (approved):

Issue #1718, opened 2022-02-25
Default font sizes from WP 5.9+ overriding theme


laurelfulford:
As of WP 5.9, the default sizes for the small, medium, large and huge fonts include `!important`, which is overriding the styles coming from the theme.

It looks like it's related to a change in WP 5.9; the theme is affected because it doesn't have a theme.json file.

laurelfulford:
This issue appears to be tied to whether or not the Newsletters plugin is activated; I've created an issue in that repo, too, in case there's something to be collectively untangled to get this working: 

https://github.com/Automattic/newspack-newsletters/issues/755

PR #1719, opened 2022-02-25
fix: button hover state background color #1713

### LINKED ISSUE #1713: Button block: `!important` in WP 5.9+ CSS variables overrides hover styles
**Describe the bug**
As of WP 5.9, the block CSS includes an `!important` when assigning preset background colours:

```
.has-white-background-color {
    background-color: var(--wp--preset--color--white) !important;
}
```

Because the theme also uses the `.has-white-background-color` in its baked in colours, this ends up getting used, and for the Button block, it overrides the default dark hover background. Instead of having white text on a dark background on hover, you get white text on a still-white background.

**To Reproduce**
1. Add a button block to a post; give the button a white background and black text.
2. Publish.
3. Hover over the button block.

**Screenshots**
Button with white background:

image

Button with white background on hover:

image

**Expected behavior**

Correct button hover colours -- white text, and the correct `#111` background:

image

9ete:
### All Submissions:

Closes #1713.

1. See https://github.com/Automattic/newspack-theme/issues/1713

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -757,7 +757,7 @@ p.has-background {
 		&:active,
 		&:focus {
 			color: #fff;
-			background: $color__background-button-hover;
+			background: $color__background-button-hover !important;
 		}
 
 		&:focus {


laurelfulford:
Thanks for this PR, @9ete! 

I'm going to hold off reviewing this for now -- I'm following a thread for #1718, which seems like it may be tangentially related -- they both involve WP global styles and  `!important` at least :) I'd like to try to nail down a fix for that issue first in case some part of it does end up overlapping with what we should do here. Just let me know if you have any questions in the meantime!

laurelfulford:
Just as a quick update, https://github.com/Automattic/newspack-theme/pull/1724 may also close the same button issue by removing the new global styles, which is where the `!important` is coming from -- depending on the feedback that change gets, anyway :) I'll leave this PR open until a decision is made there!

laurelfulford:
We've gone ahead with #1724, so I'm going to close this PR as that also resolves the same issue.

Thanks again for taking the time to look at this, @9ete!

Issue #111, opened 2019-07-18
Customizer custom colors preview overriden by style.css


claudiulodro:
**Describe the bug**
In the Customizer, when I have a custom color selected, I get a flash of the color when the page is first loaded for .5 seconds, and then it all gets replaced with the default Newspack color. Selecting a custom color doesn't show the color in the Customizer, and it's always mighty blue all the time. On the site frontend, the custom color works great.

In the inspector, if I examine e.g. a Search button I see the default color is active:
`style.css`
```
.button, button, input[type="button"], input[type="reset"], input[type="submit"] {
    background: #2A7DE1;
    ...
}
```
At a lower priority (striked through in the inspector), my custom color is inactive:
`http://{site}/?customize_changeset_uuid={uuid}&customize_theme=newspack-theme&customize_messenger_channel=preview-0`
```
.main-navigation .sub-menu, .sticky-post, .entry .entry-content .wp-block-button .wp-block-button__link:not(.has-background), .entry .button, button, input[type="button"], input[type="reset"], input[type="submit"], .entry .entry-content > .has-primary-background-color, .entry .entry-content > *[class^="wp-block-"].has-primary-background-color, .entry .entry-content > *[class^="wp-block-"] .has-primary-background-color, .entry .entry-content > *[class^="wp-block-"].is-style-solid-color, .entry .entry-content > *[class^="wp-block-"].is-style-solid-color.has-primary-background-color, .entry .entry-content .wp-block-file .wp-block-file__button {
    background-color: #eab700; /* My custom color */
}
```

Potentially related to #52? (The issue happens without a style pack active in my testing, though)

**To Reproduce**
1. Go to Customizer
2. Set a custom color. Observe it doesn't preview.
3. Save. Refresh the page. Observe custom color is overriden by style.css.

**Expected behavior**
I expected the custom color preview to preview normally.

claudiulodro:
Not sure. I should be getting my new computer today, so I'll give it a test on a clean slate and see.

laurelfulford:
Thanks! (And yay new computer!! 🎉 )

claudiulodro:
Just tested on the cleanest of new sites, and the issue still happens. It's not because I had a style pack and then removed it.

laurelfulford:
Thanks for double-checking! 

Dang, this could be a bigger issue than I thought.

claudiulodro:
[...]

PR #1724, opened 2022-03-04
Dequeue FSE global styles

### LINKED ISSUE #1713: Button block: `!important` in WP 5.9+ CSS variables overrides hover styles

laurelfulford:
### All Submissions:

WordPress 5.9 introduced a Global Styles file, which is meant to be populated by a theme's theme.json file. It's still enqueued on non-FSE themes, and is causing a couple small issues in the Newspack theme:

* It shares a `white` background colour class with the theme, and the `!important` it uses overrides the theme's button hover styles (#1713).
* They're part of an issue between the theme, Newsletters plugin and how non-FSE theme editor font styles are handled. Dequeuing them is not a complete fix, but it does correct the font-sizes on the front-end. As part of this, I've also added a temporary x-large font size class to the theme so sites experiencing the editor font sizing mismatch will still have the correct font sizes applied (#1718, https://github.com/Automattic/newspack-newsletters/issues/755).
* AMP is not tree-shaking these styles because they're CSS variables. The styles add 3,231b (4.3% of 75kb) to a site's front-end, or more if any of the overlapping classes are used (like the colour white, and the small, large, and x-large font sizes). 

Closes #1713; see #1718 and https://github.com/Automattic/newspack-newsletters/issues/755.

1. Start on a test site running the Newsletter plugin.
2. Create and publish a blank page. Either view the non-AMP version of the page and confirm the presence of a `global-styles-inline-css` style tag in the source, or confirm those styles are there in AMP's CSS usage. 
3. In the editor, view source and confirm the presence of a `global-styles-css-custom-properties-inline-css` style tag.
4. To test out the issue in #1713, add a button block and make the background white and the text colour one of the other colours in the editor palette. On the front end, try hovering over the button; note that it turns white-on-white. 
5. To test out the issue in #1718, add multiple Heading blocks, and make one small, one large, and one x-large; note how they're smaller than the theme's defaults (small: 16px, large: 36px, huge: 44px), and are actually picking up the sizes set in the Newsletter (small: 12px, large: 24px, x-large: 36px). They should be too small in the editor and on the front-end. 
7. Confirm that the button hover is now working as expected -- rather than being white on white, it is now white on `#111`. 
8. Confirm that the font sizes on the front end line up with the theme's settings (small: 16px, large: 36px, x-large/huge: 44px).
9. With AMP disabled, view the source and confirm `global-styles-inline-css` is no longer loaded, or confirm those styles are no longer there in AMP's CSS usage. 
10. In the editor, view source and confirm that the `global-styles-css-custom-properties-inline-css` styles are no longer getting loaded.
11. Out of an abundance of caution, confirm that theme/block styles that did appear in the global styles still work as expected. 
    * The theme-based custom colours (primary, secondary, and the greyscale)
    * Gradients available in the editor palette
    * Font sizes (I recommend deactivating Newsletters before testing, to untangle https://github.com/Automattic/newspack-newsletters/issues/755 from the mix). 
    * Duo-tones (can be tested by adding an image block, and clicking on the circular/starburst icon in the toolbar, then selecting one of the options):

image

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -275,6 +275,7 @@ if ( ! function_exists( 'newspack_setup' ) ) :
 endif;
 add_action( 'after_setup_theme', 'newspack_setup' );
 
+
 /**
  * Register widget area.
  *
@@ -596,6 +597,28 @@ function newspack_enqueue_scripts() {
 	wp_enqueue_script( 'newspack-post-meta-toggles' );
 }
 add_action( 'enqueue_block_editor_assets', 'newspack_enqueue_scripts' );
+
+
+/**
+ * Dequeue Gutenberg global styles.
+ * 
+ * These styles aren't currently used because Newspack is not a Full-Site editing theme.
+ */
+function newspack_dequeue_global_styles() {
+	wp_dequeue_style( 'global-styles' );
+}
+add_action( 'wp_enqueue_scripts', 'newspack_dequeue_global_styles' );
+
+/**
+ * Dequeue Gutenberg global editor styles.
+ * 
+ * These styles aren't currently used because Newspack is not a Full-Site editing theme.
+ */
+function newspack_dequeue_global_editor_styles() {
+	wp_dequeue_style( 'global-styles-css-custom-properties' );
+}
+add_action( 'enqueue_block_editor_assets', 'newspack_dequeue_global_editor_styles', 99 );
+
 
 /**
  * Fix skip link focus in IE11.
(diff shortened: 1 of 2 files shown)


dkoo (approved):
This fixes the issues described and I'm still able to use the dutone and gradient options in the editor. 👍

laurelfulford:
Thanks @dkoo!

PR #1738, opened 2022-03-22
Add `--wp--style--block-gap` default for the button block

### LINKED ISSUE #1736: `--wp--style--block-gap` default value affected by button block font size
The Button Blocks uses the CSS variable `--wp--style--block-gap` to calculate the space between the button blocks; because Newspack is not a FSE theme with a theme.json, it uses the fallback value 0.5em, to calculate the gap.

For the button block, this space becomes too small when you reduce the font size of the buttons, so if you set them to say, 50% of the available width, they will be the correct width but stack instead of sitting side by side. 

Steps to replicate:

1. Set up a button block with two buttons.
2. Make them each 50% wide.
3. Decrease the font size of the first button to `15px`. 
4. The second button will wrap to a new line:

Default font sizes:

image

Smaller font sizes: 

image

**Temporary work around**
Setting a non-em value for the CSS variable in the Custom CSS will fix this issue on the front-end -- example: 

```
.wp-block-buttons {
    --wp--style--block-gap: 0.5rem;
}
```

laurelfulford:
### All Submissions:

Blocks now use the `--wp--style--block-gap` CSS variable to set the space between different block elements. Since the Newspack theme doesn't have a theme.json, it doesn't set its own values for these variables, but it does use the WordPress fallbacks, which look like they're often measured in `ems`.

For the buttons block specifically, this `em` value is affected when you change the font size of the individual buttons -- decreasing the font size can decrease the gap to the point where there is no longer space for the buttons to sit side by side. This PR adds a gap value fallback that no longer is affected by changing the button block font size, but without affecting other blocks that use the value that don't have this same issue (like the Columns block).

So far I haven't run into this issue for other blocks --  possibly because where the font styles are applied -- but it seems like something we'll want to fix on a case by case basis if there are others, just to make sure we don't add a spacing default that works in some cases, but not others!

Closes #1736

1. Add a Buttons block to the editor, and add at least two buttons.
2. Make each button 50% of the available space; note how they site side by side on the front-end and in the editor:

image

4. Decrease the font size of the first button to `15px`.
5. Note that the buttons now stack:

image

7. Confirm that the buttons now sit side-by-side as expected on the front-end and in the editor:

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1,5 +1,4 @@
 /* !Block styles */
-
 .entry .entry-content > *,
 .newspack-inline-popup > *,
 [id='pico'] > * {
@@ -698,6 +697,10 @@ p.has-background {
 }
 
 //! Button
+.wp-block-buttons {
+	--wp--style--block-gap: 0.5rem;
+}
+
 .wp-block-button__link {
 	background-color: $color__background-button;
 	color: #fff;
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1,4 +1,5 @@
 /* !Block styles */
+
 .entry .entry-content > *,
 .newspack-inline-popup > *,
 [id='pico'] > * {
--- a/newspack-theme/sass/style-editor-base.scss
+++ b/newspack-theme/sass/style-editor-base.scss
@@ -423,6 +423,9 @@ div[data-align='full'] {
 }
 
 /** === Button === */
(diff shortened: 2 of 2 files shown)


dkoo (approved):
This fixes the issue of the gap. I tested with various combinations of font sizes and didn't see any problems with stacking.

Perhaps not totally related to this, but I notice that the discrepancy between editor font sizes (e.g. "normal" is 16px in the editor but 20px on the front-end) persists even if you set the font size as a pixel value. In other words, if you set the font size to 15px in the editor, it shows up as 15px on the front-end, but set it to 16px and it renders as 20px. A bit disconcerting but perhaps already a known bug?

PR #1742, opened 2022-03-24
Update ad spacing and move ad background styles

### LINKED ISSUE #1739: Featured Image behind + below header ad gets cut off slightly
**Describe the bug**
When you pair the featured image behind image placement with the below header global ad position, the ad gets cut off ever so slightly (2-3px) due to the theme's negative bottom margin, and it seems a bit odd that the image + ad touch.

**To Reproduce**
1. Publish a single post using the image behind featured image placement.
2. Add an ad to the below header global space.

**Screenshots**

image

**Expected behavior**

That the entire ad would be visible, and that there would be at least a slight white gap to match the top.

laurelfulford:
### All Submissions:

This PR makes the following changes to the ads styles:

* Moves the non-colours styles out of the color-patterns.php file and into the ads Sass file. For the styles that only should be applied when the ads have a background image, the CSS class `.ads-bg` is added and used.
* Adjusts the gap underneath below-header ads when paired with the behind featured image placement so the ad isn't covered (#1739).

Note: In testing this I found a similar issue is happening with the footer ad location that I have to file/dig into further. I suspect the styles that are there worked originally, but when the `overflow: hidden` was moved to get the sticky header working, the ad started getting cut off 😕 

Closes #1739.

1. Start on a test site running ads; add ads to the below header global spot, and the above footer global spot.
2. Turn on "Ads Background Color" under Customizer > Colors.
3. Create a post; set the featured image to behind. 
4. View on the front end and note that the featured image seems to be touching your ad; if you inspect it, they're actually overlapping just a hair (2-3px):

image

6. Recheck your post with the 'behind featured image' setting and confirm that the image and ad have a small space between them:

image

7. Confirm that the `ads-bg` class is being added so the other spacing styles originally added through the color-patterns.php file should be applied.
8. Turn off the ad background colour and confirm that the top and bottom ad spacing still seem appropriate. 

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -471,21 +471,6 @@ function newspack_custom_colors_css() {
 			div[class*="newspack-ads-blocks-ad-unit"] {
 				background-color: ' . esc_html( get_theme_mod( 'ads_color_hex', '#ffffff' ) ) . ';
 			}
-			.single-featured-image-behind .newspack_global_ad.global_below_header,
-			.newspack_global_ad.global_above_footer {
-				margin-bottom: -2rem;
-			}
-			.newspack_global_ad.global_above_footer {
-				margin-top: 2rem;
-			}
-			.newspack_global_ad > * {
-				margin-bottom: 8px;
-				margin-top: 8px;
-			}
-			.widget_newspack-ads-widget .textwidget,
-			div[class*="newspack-ads-blocks-ad-unit"] {
-				padding: 8px;
-			}
 		';
 	}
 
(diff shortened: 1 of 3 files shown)


dkoo (approved):
Fixes the issue with the Above Header placement (not with the Above Footer, but that sounds like it'll need a separate PR?).

laurelfulford:
Thanks @dkoo!

> Fixes the issue with the Above Header placement (not with the Above Footer, but that sounds like it'll need a separate PR?).

Unfortunately yes -- I spun up a fresh issue for it here: #1743. I'd hoped to knock them both out with one, but it looks like that one will use a bit more experimenting (and maybe not just CSS) to fix 😕

PR #1744, opened 2022-03-28
fix: remove bottom space from last block in cover block

### LINKED ISSUE #1210: Columns Block + Cover Block: Extra space on bottom
When a columns block is nested inside of a cover block, there is extra space on the bottom due to the default bottom margin of the columns block. 

Removing the top margin from the first element in the Cover block, and the bottom margin from the last element in the Cover block, should help fix.

laurelfulford:
### All Submissions:

This PR tweaks the Cover block styles, to make sure the first and last blocks nested inside don't have a top and bottom margin, respectively. This style already existed in the theme, but was only being applied when the cover block was aligned left or right -- it makes much more sense for it to be applied all the time. 

Closes #1210.

1. Add a cover block to a post or page, and nest a columns block inside. Add enough content to the columns block so that it's taller than the cover block's minimum height (this is 300px by default, but you can also change the minimum height to as low as 50px using the options in the right sidebar).
2. Publish and view on front-end.
3. Note that the column block has more space underneath it than above:

image

6. Confirm that the space is now gone:

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -439,15 +439,15 @@ p.has-background {
 		.wp-block-pullquote:not( .is-style-solid-color ) blockquote {
 			padding-left: 0;
 		}
+	}
 
-		.wp-block-cover__inner-container {
-			[class*='wp-block-']:first-child {
-				margin-top: 0;
-			}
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
👌

laurelfulford:
Thanks Miguel! 🙇‍♀️

PR #1745, opened 2022-03-28
fix: make cover block link styles more specific

### LINKED ISSUE #1741: Cover Block's styles can cause button inheritance weirdness
**Describe the bug**
The theme includes some styles on the cover block   

**To Reproduce**
1. Create two cover blocks -- one with a really light background image, and one with a darker background image. (In both cases you want to use images light enough and dark enough for Gutenberg to determine if they're light or dark -- you can tell if this works because the placeholder text will be dark if Gutenberg thinks the image is light, or white if it thinks the image is dark) 
2. Add a button block to each cover block.
3. In the 'light' cover block, make the button have a dark background and white text (using the theme's greyscale colour palette). 
3. In the 'dark' cover block, make the button have a white background and dark text (using the theme's greyscale colour palette). 
4. Publish.
5. Note on the front-end the text in both cases is illegible. 

**Screenshots**

image

**Expected behavior**
That the button colour settings would be respected; instead, they're being overridden by these styles here: 

https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/sass/blocks/_blocks.scss#L455-L461

```
.wp-block-cover .wp-block-cover__inner-container {
	a,
	a:hover,
	a:visited {
		color: inherit;
	}
}
```

It's entirely possible those styles are no longer required (some of the theme's cover block styles are so old they reference the original `.wp-block-cover-image` class); worst case, they shouldn't override any colours set using the block colour tools.

laurelfulford:
### All Submissions:

This PR fixes an issue where the button blocks were picking up the wrong link colour when nested in a cover block.

Closes #1741

1. Add two cover blocks -- one with a lighter image and one with a darker image. This is to test with both the light and dark text styles the block uses -- the lighter image should automatically use darker text, and the darker image should automatically use white text.
2. Add a buttons block to each cover block, and add multiple buttons with different settings - eg. a white button with black text, and black button with white text, and buttons using the primary and secondary colours. 
3. Publish and view on the front end. 
4. Note that the buttons are picking up the cover block's text colour, rather than the one assigned in the settings -- so you can have black text on a darker background, light text on a lighter background, and styles that just don't line up with the settings: 

image

7. Refresh your test page and confirm that the buttons are using the correct colours, and are white text on dark grey on hover:

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -453,9 +453,9 @@ p.has-background {
 }
 
 .wp-block-cover .wp-block-cover__inner-container {
-	a,
-	a:hover,
-	a:visited {
+	a:not( .wp-block-button__link ),
+	a:not( .wp-block-button__link ):hover,
+	a:not( .wp-block-button__link ):visited {
 		color: inherit;
 	}
 }


miguelpeixe (approved):
Can confirm it fixes the issue for button colors!

laurelfulford:
Thank you! 🙌

PR #1746, opened 2022-03-28
fix: update approach to remove space with above footer ad

### LINKED ISSUE #1743: Above footer ad is getting cut off when background colour is set.
**Describe the bug**

Discovered while testing #1742.

The `margin-bottom: -2rem` on the above footer ad space is causing the ad to get cut off -- I suspect it was originally added to close the gap between the ad and the footer.

**To Reproduce**
1. Add an above footer ad to your site.
2. Under Customizer > Colours, enable the Ad background colour.
3. View on front end; note the ad is getting cut off, and the background is not touching the footer.

**Screenshots**

image

**Expected behavior**

That the full ad would display, and that the gap between the background and footer would be gone.

laurelfulford:
### All Submissions:

This PR changes the approach to removing the gap between the above footer ad space when it has a background, and the footer. 

Originally this was done by adding a negative margin to the ad space; I believe that stopped working as expected when the `overflow: hidden` was moved in #1254-- now, the negative margin causes the ad to get cut off. 

This PR checks for an ad in that location, and if there is one, adds a CSS that's then used to remove the margin above the footer. 

This approach seemed the least disruptive (it doesn't try to move the ad; it doesn't remove the margin on top of the footer outright), but I'm open to feedback if it seems like it's over-complicating things. I also may not have used the best approach for checking for the Above Footer ad space!

Closes #1743

1. Navigate to Customizer > Colors, and add a background colour to your ads.
2. Add an ad to the 'Above Footer' location.
3. View on the front end -- there is a gap between your ad background colour, and part of your ad is getting cut off: 

image

6. Confirm that the ad is no longer cut off, and there is no longer a gap between the ad background and the footer. 

image

### Other information:

### DIFF
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -214,16 +214,18 @@ function newspack_body_classes( $classes ) {
 	}
 
 	// Add a class when there's an ad background color, and another when there's an ad above the footer to remove the space.
-	$ads_background_color = get_theme_mod( 'ads_color', 'default' );
-	$above_footer_ad = get_option( '_newspack_ads_placement_global_above_footer' );
-	if ( 'custom' === $ads_background_color ) {
-		$classes[] = 'ads-bg';
+	if ( is_plugin_active( 'newspack-ads/newspack-ads.php' ) ) {
+		$ads_background_color = get_theme_mod( 'ads_color', 'default' );
+		$above_footer_ad = get_option( '_newspack_ads_placement_global_above_footer' );
+		if ( 'custom' === $ads_background_color ) {
+			$classes[] = 'ads-bg';
 
-		if ( str_contains( '"enabled":true', $above_footer_ad ) ) {
-			$classes[] = 'ad-above-footer';
+			if ( str_contains( '"enabled":true', $above_footer_ad ) ) {
+				$classes[] = 'ad-above-footer';
+			}
 		}
 	}
-
+	
 	// Add a class for the footer logo size.
 	$footer_logo_size = get_theme_mod( 'footer_logo_size', 'medium' );
 	if ( 'medium' !== $footer_logo_size ) {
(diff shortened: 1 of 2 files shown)


newspack-theme/inc/template-functions.php
```diff
@@ -213,12 +213,17 @@ function newspack_body_classes( $classes ) {
 		$classes[] = 'feature-latest';
 	}
 
-	// Add a class when there's an ad background color.
+	// Add a class when there's an ad background color, and another when there's an ad above the footer to remove the space.
 	$ads_background_color = get_theme_mod( 'ads_color', 'default' );
+	$above_footer_ad = get_option( '_newspack_ads_placement_global_above_footer' );
```
miguelpeixe, line 218:
```suggestion
	$above_footer_ad      = method_exists( 'Newspack_Ads\Placements', 'can_display_ad_unit' ) && \Newspack_Ads\Placements::can_display_ad_unit( 'global_above_footer' );
```

The placements system provides a public method with more advanced checks to verify if the placement will render an ad unit which I believe is more suitable.

newspack-theme/inc/template-functions.php
```diff
@@ -213,12 +213,17 @@ function newspack_body_classes( $classes ) {
...
 	if ( 'custom' === $ads_background_color ) {
 		$classes[] = 'ads-bg';
-	}
 
+		if ( str_contains( $above_footer_ad, '"enabled":true' ) ) {
```
miguelpeixe, line 222:
```suggestion
		if ( true === $above_footer_ad ) {
```

The suggested method above returns a boolean value.

miguelpeixe (changes requested):
Working as described! One suggestion below:

newspack-theme/inc/template-functions.php
laurelfulford: 🙌 Awesome, thanks Miguel! I made this update in [a2e5486](https://github.com/Automattic/newspack-theme/pull/1746/commits/a2e548679d56567b2ca57d2f2880abb2490dbcb8).

laurelfulford:
Woohoo! Thanks for the review! 🙌

PR #1747, opened 2022-03-29
Fix alignment of square logos when paired with the mobile CTA

### LINKED ISSUE #1618: Square-shaped logos appear to be 'floating' in some cases with AMP enabled
Moved from #1594: 

> I also notice that on the homepage at the smallest breakpoint, my default/square logo seems to have a bit of extra side padding that doesn't appear at other breakpoints (this seems to happen any combination of header/logo settings). This also happens with the subpage header but only on AMP pages:

image

AMP sets images to `object-fit: contain`; if the image has a max-width that's wider than the actual image, you can get this effect of the image floating in the middle of the space. I'm fairly confident this is what's happening here -- it could just be that we need to adjust the logo sizing, or add an `object-position` when the image should be flush left (it doesn't work when there should be something next to the image, since the image "space" is still there, but for mobile-only it could be the best bet!).

laurelfulford:
### All Submissions:

This PR tidies up some weird logo alignment issues when you have the mobile CTA and AMP enabled. 

Closes #1618

1. Start with a test site running AMP. 
2. Change your logo to a small square logo, and make it smaller than 140px wide (preferably more like 80-100px wide, so the issue is easier to see). 
3. Under Customizer > Header Settings, turn on the Mobile CTA, and turn off the Sticky Header (since the styles there "fix" the issue). 
4. View on the front-end, and shrink down your browser window to a mobile-sized screen.
5. Note that on screens smaller than 600px wide, the square logo seems to float slightly  away from the left. This is due to a max-width on the logo (to make sure it doesn't run into the mobile CTA), and how AMP handles max widths on images:

image

7. Refresh the page; confirm that the logo is flush to the left:

image

8. Confirm no new issues are introduced when you increase the screen size. 
9. Try switching the header height and alignment under Customizer > Header Settings > Appearance (toggling on or off the Short Header and Centred Logo settings) and confirm no issues are present. 

### Other information:

### DIFF
--- a/newspack-theme/sass/site/header/_site-header.scss
+++ b/newspack-theme/sass/site/header/_site-header.scss
@@ -28,6 +28,7 @@
 
 		.custom-logo {
 			max-width: 100%;
+			object-position: left center;
 		}
 	}
 }


miguelpeixe (approved):
👌

laurelfulford:
Thanks Miguel! 🥳

PR #1749, opened 2022-03-29
Define some of the global styles in the theme

### LINKED ISSUE #1748: Missing Global Styles cause colours like #000 not to work on the front end
**Describe the bug**
The theme [dequeues](https://github.com/Automattic/newspack-theme/pull/1724) WP 5.9's Global Styles to resolve some other style issues, but it looks like any colours matching those classes -- like `has-black-color` and `has-black-background-color` -- will be used if you manually pick their matching hex value in the editor. This means the theme should define these colours -- at least the most basic ones -- for them to work on the front end. 

A set of global styles classes can be found ~~[here](https://github.com/WordPress/gutenberg/blob/6d25339faeee5cb9beeed427c9f15bd8104eb202/packages/edit-site/src/components/global-styles/test/use-global-styles-output.js#L380)~~ actually that's not really right, here's the class name formats and colours. We probably won't run into issues with all, but it could be a good safety net: 

**Class name formats** 
* `.has-[colour]-color`
* `.has-[colour]-background-color`
* `.has-[colour]-border-color`

**Colours**
* `black: #000000;`
* `cyan-bluish-gray: #abb8c3;`
* `pale-pink: #f78da7;`
* `vivid-red: #cf2e2e;`
* `luminous-vivid-orange: #ff6900;`
* `luminous-vivid-amber: #fcb900;`
* `light-green-cyan: #7bdcb5;`
* `vivid-green-cyan: #00d084;`
* `pale-cyan-blue: #8ed1fc;`
* `vivid-cyan-blue: #0693e3;`
* `vivid-purple: #9b51e0;`

**To Reproduce**
1. Add a group block and give it a background colour of `#000`
2. Note that it works in the editor. 
3. Publish the post or page.
4. Note that the colour is not applied on the front-end.

laurelfulford:
### All Submissions:

As of #1724, the theme has dequeued WP 5.9's global styles, because the `!important` they include were causing unexpected issues. 

If you happen to enter one of the hex values from the colours used in the Global Styles in the block colour picker, it looks like the standard inline hex value is replaced with one of the Global Styles CSS classes on the front end. Because these classes are not defined, the colour is not applied on the front end. 

Though the odds are very low that this would happen for most of the hex values, but we've already had reports of issues with folks trying to use `#000`. I have included the full set of colour, background colour and border colour classes, on the very unlikely chance some stumbles upon a "luminous-vivid-orange" exact match. 

Closes #1748 

1. Add multiple group blocks to your editor, and give them different background colours -- #000, and a couple of the other 'unlikely' defaults -- like `#9b51e0` (vivid purple) and `#7bdcb5` (light green cyan):

image

2. Publish your post; view on the front-end and note your colours aren't applied:

image

5. Confirm the colours are now applied:

image

6. Test out the colours (primarily black) on other blocks (headers, buttons, pullquotes...) just to make sure it displays as expected. 

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -130,7 +130,7 @@ function newspack_custom_colors_css() {
 			.wp-block-quote:not(.is-style-large),
 			.woocommerce-tabs ul li.active a,
 			.has-primary-border-color,
-			.wp-block-pullquote .has-primary-border-color {
+			.wp-block-pullquote.has-primary-border-color {
 				border-color: ' . esc_html( $primary_color ) . '; /* base: #0073a8; */
 			}
 
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1353,6 +1353,7 @@ $colors: (
 	.has-#{$name}-background-color,
 	.has-#{$name}-background-color.has-background-dim,
 	.wp-block-pullquote.is-style-solid-color.has-#{$name}-background-color,
+	.wp-block-pullquote.has-#{$name}-background-color,
 	.is-style-outline .wp-block-button__link.has-#{$name}-background-color:not( :hover ) {
 		background-color: $hex;
 	}
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1383,6 +1383,83 @@ hr {
 	background-color: #fff;
 }
 
+.has-black-background-color,
+.has-black-background-color.has-background-dim,
+.wp-block-pullquote.is-style-solid-color.has-black-background-color,
+.is-style-outline .wp-block-button__link.has-black-background-color:not( :hover ) {
+	background-color: #000;
+}
+
+.has-cyan-bluish-gray-background-color,
+.has-cyan-bluish-gray-background-color.has-background-dim,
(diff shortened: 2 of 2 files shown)


adekbadek (commented):
I took the liberty of refactoring the code to reduce duplication in 959e520aeb37ddf765f892250b2561b55a620a51, let me know if that's ok and if I haven't messed it up while refactoring!

laurelfulford:
Ahhhh, that's so much better @adekbadek - thank you! 

I did a bunch of testing, and found two issues, but neither would have been introduced with the refactor: 

* There was a typo in the color-patterns.php file, so the primary border colour was always blue - before and after, and fixed in [305bc20](https://github.com/Automattic/newspack-theme/pull/1749/commits/305bc20fdacac56e6deabef8ccc4b119629181ae): 

image

image


* The selector for the pullquote backgrounds wasn't specific enough to override a style from WordPress Core setting the `.wp-block-pullquote:not( .is-style-solid-color )` to `background: none` - before and after and fixed in [b52b9d5](https://github.com/Automattic/newspack-theme/pull/1749/commits/b52b9d5c8354ddddcb002e35b8075f140ea9c545). In WP 5.9 the global styles colours all use `!important`, which would explain why this wouldn't be an issue normally (unfortunately the `!important`s were causing us other issues!):

image

image

Could you take it for one last spin just to make sure things look okay?

adekbadek (approved):
Looks good after latest changes!

PR #1750, opened 2022-03-30
Improve behaviour of floated groups, and fix display issues with iframe blocks nested inside

### LINKED ISSUE #1710: Style `<iframe` to be `display: block`
AMP in many cases will make `amp-iframe` display block (like if it has a fixed width or height) -- does it makes sense to style the `iframe` tag the same way in the theme? 

It shouldn't cause any major issues, and it will definitely help make sure things display consistently across AMP and non-AMP pages as the iframes pick up styles from the theme ( for example, it would help make sure the iframe is centred if it's the child of a full-width group block). 

The only thing that should be confirmed prior to this change is that it is the most "likely" way for AMP to style `amp-frame`, for consistency's sake.

### LINKED ISSUE #1723: Give iframes in floated group blocks a min-width
If you combine an iFrame block with a 100% width in a floated group block, the iframe collapses completely when AMP is enabled due to how all of the styles come together. 

We should be able to work-around this by giving iframes in floated group blocks a min-width -- while they'll likely need a px width to truly match the iframe's contents, this way, they will at least display!

laurelfulford:
### All Submissions:

This PR cleans up some behaviour of the floated group block and the iframe block:

* It adds a `display: block` to all `iframe` elements so they behave consistently across AMP and non-AMP (`amp-frame` appears to be largely styled as `display: block`).
* Adds a minimum width to the iframe block when nested inside of a floated group block when its width is 100%; otherwise, it collapses. 
* Adds a maximum width to the iframe block, so it cannot exceed the available space when it has a `px` width set. 
* Makes floated group blocks stack on mobile; we've done this in Custom CSS on a couple sites already; it seems like the most logical default, as having the group block squished to 50% of the screen space would rarely make sense. 

Closes #1710; closes #1723.

1. Create a page and add a couple of floated group blocks, containing:
   * An iframe block with the default width (`100%) 
   * An iframe block with a higher fixed `px` width, like `500px`
   * Some text content -- a paragraph block or two, or a list. 
2. View on the front end; note that:
   * The floated group block + iframe block with the default width disappears.

image

   * The floated group block + iframe block with a wider fixed width is too wide when you start scaling down the browser.

image

   * The floated group block + regular content seems cramped on smaller screens.

image

4. Confirm the following:
    * The floated group block + iframe block with the default width is now visible.

image

    * The floated group block + iframe block with a wider fixed width doesn't exceed the available space (it is smaller than your selected width, but I think that makes sense in this case -- if you need a wider iframe, it shouldn't be floated).

image

    * The floated group block stacks on mobile.

image

5. Double-check the page with AMP disabled as well, to make sure there are no unexpected issues!

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1140,11 +1140,17 @@ hr {
 	}
 }
 
-@include media( mobileonly ) {
-	.entry .entry-content > .wp-block-group.alignleft,
-	[id='pico'] > .wp-block-group.alignleft,
-	.entry .entry-content > .wp-block-group.alignright,
-	[id='pico'] > .wp-block-group.alignright {
+.entry .entry-content > .wp-block-group.alignleft,
+[id='pico'] > .wp-block-group.alignleft,
+.entry .entry-content > .wp-block-group.alignright,
+[id='pico'] > .wp-block-group.alignright {
+	@include media( mobile ) {
+		&:not( :first-child ) {
+			margin-top: 0;
+		}
+	}
+
+	@include media( mobileonly ) {
 		float: none;
 		margin-left: 0;
 		margin-right: 0;
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1129,6 +1129,27 @@ hr {
 			margin-bottom: 0;
 		}
 	}
+
+	@include media( mobile ) {
+		&.alignleft,
+		&.alignright {
+			iframe[style*='width: 100%'] {
+				width: 400px !important; // !important to override inline style.
+			}
+		}
+	}
+}
+
+@include media( mobileonly ) {
+	.entry .entry-content > .wp-block-group.alignleft,
+	[id='pico'] > .wp-block-group.alignleft,
+	.entry .entry-content > .wp-block-group.alignright,
+	[id='pico'] > .wp-block-group.alignright {
+		float: none;
+		margin-left: 0;
+		margin-right: 0;
+		max-width: 100%;
+	}
 }
 
 .wp-block-group.has-background + .wp-block-group.has-background,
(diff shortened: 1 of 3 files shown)


dkoo (changes requested):
Looking great at mobile and desktop viewports (the stacking on mobile in particular looks much better), but at the tablet breakpoint I'm seeing all three test Group blocks break outside of the viewport:

Screen Shot 2022-03-31 at 10 47 24 AM

Screen Shot 2022-03-31 at 10 47 33 AM

In the inspector, I'm seeing some negative margins on these that I think need to be zeroed out at the tablet breakpoint:

Screen Shot 2022-03-31 at 10 50 43 AM

laurelfulford:
Thanks @dkoo! 

I've made a few adjustments:

* I've tweaked the styles that pull the aligned blocks out to the left and right -- when scaling down the browser window, they should still just touch the edge at some point, but no longer get cut off like they do now. 
* I added the same iframe width fallback to the editor -- the problem wasn't as bad there, the iframe was still visible, but still kind of weirdly narrow. The selector's a little funkier though -- since there's also a `max-width: 100%`, I can't use the same `style*='width: 100%'` without it also matching with the `max-width`, so it's more specific. 
* I noticed the image blocks weren't respecting the max-width in the editor, so I fixed that.

Just let me know if you have any questions about the above, or if you notice anything else amiss! :)

dkoo (changes requested):
Since you mentioned image blocks, I did some testing in floated Group blocks and found that if you set the width of the image to something wider than the Group's max width, you get a slightly strange result in the editor:

Screen Shot 2022-04-05 at 1 03 35 PM

It looks fine on the front-end, though, with the image maxing out at the Group's max-width.

laurelfulford:
Thanks @dkoo!

I was actually testing image blocks on their own, since they can be aligned left and right without a group block. I did find a similar problem there, though, and tried to fix both that and the nested-in-group-one to the best of my abilities. The fix isn't responsive, but it's hard to get that 1:1 with the front-end -- I think this is about as close as I can get it at the moment (but do let me know if you spot anything glaring!).

dkoo (approved):
This fixes the overlap I saw earlier, and I didn't see any glaringly bad behavior in the editor on on the front-end, using images in groups or not in groups.

I did see that the selection outline for an image within a floated group looks a little off, but this seems to me like an extreme edge case and doesn't really affect the editing experience much, IMO.

laurelfulford:
Thanks for all the testing on this one, @dkoo!

PR #1752, opened 2022-03-30
Make sure captions on full-width blocks don't get cut off

### LINKED ISSUE #1751: Full-width image captions get cut-off on smaller screens
**Describe the bug**
The full-width image caption has a max-width of 780px; because of this, it ends up getting cut off on smaller screens.

Instead, it should probably have a width of 780px, and a max-width of 90% to match other elements in the theme.

laurelfulford:
### All Submissions:

This PR adjusts the width and max-width of captions on full-width blocks, to make sure they don't get cut off on smaller screens. 

Closes #1751

1. Create a new post/page and choose the one column template.
2. Add an image block and a video block; give them both captions and make each block full width. Fill in the space between with some paragraphs to compare content widths. 
3. Publish and view on the front-end; the captions should match the width of the other content (780px wide):

image

5. Scaled down your browser window -- note that your captions hit the edges and get cut off:

image

7. Apply the PR and run `npm run build`.
8. Confirm that the captions are still the same width as the regular content on desktop, and are also the same width on mobile:

image

image

10. Switch your template to the One Column wide template, and retest the above; the captions should match the content width on desktop (1200px), and not hit the edges and matches the content width on mobile:

image

image

### Other information:

### DIFF
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -5,7 +5,7 @@
 	flex-wrap: wrap;
 	justify-content: space-between;
 	margin: auto;
-	max-width: 90%;
+	max-width: 90vw;
 	width: $size__site-main;
 
 	@include media( tablet ) {
@@ -34,7 +34,7 @@
 
 #primary {
 	margin: auto;
-	max-width: 90%;
+	max-width: 90vw;
 	width: $size__site-main;
 }
 
--- a/newspack-theme/sass/media/_captions.scss
+++ b/newspack-theme/sass/media/_captions.scss
@@ -38,7 +38,7 @@ figcaption,
 	}
 
 	.alignfull > figcaption {
-		width: min( 90%, 780px );
+		width: min( 88%, 780px );
 	}
 }
 
@@ -53,6 +53,6 @@ figcaption,
 	}
 
 	.alignfull > figcaption {
-		width: min( 90%, 1200px );
+		width: min( 88%, 1200px );
 	}
 }
--- a/newspack-theme/sass/media/_captions.scss
+++ b/newspack-theme/sass/media/_captions.scss
@@ -37,8 +37,11 @@ figcaption,
 		max-width: 780px;
 	}
 
-	.alignfull > figcaption {
-		width: min( 90%, 780px );
+	.alignfull,
+	.alignwide {
+		> figcaption {
+			width: min( 90vw, 780px );
+		}
 	}
 }
 
@@ -52,7 +55,10 @@ figcaption,
 		}
 	}
 
-	.alignfull > figcaption {
-		width: min( 90%, 1200px );
+	.alignfull,
+	.alignwide {
+		> figcaption {
+			width: min( 90vw, 1200px );
+		}
 	}
 }
--- a/newspack-theme/sass/media/_captions.scss
+++ b/newspack-theme/sass/media/_captions.scss
@@ -38,7 +38,7 @@ figcaption,
 	}
 
 	.alignfull > figcaption {
-		width: min( 88vw, 780px );
+		width: min( 90%, 780px );
 	}
 }
 
@@ -53,6 +53,6 @@ figcaption,
 	}
 
 	.alignfull > figcaption {
-		width: min( 88vw, 1200px );
+		width: min( 90%, 1200px );
 	}
 }
--- a/newspack-theme/sass/media/_captions.scss
+++ b/newspack-theme/sass/media/_captions.scss
@@ -36,6 +36,10 @@ figcaption,
 	figcaption {
 		max-width: 780px;
 	}
+
+	.alignfull > figcaption {
+		width: min( 90%, 780px );
+	}
 }
 
 .newspack-front-page,
@@ -46,5 +50,9 @@ figcaption,
 		figcaption {
 			max-width: 1200px;
 		}
+	}
+
+	.alignfull > figcaption {
+		width: min( 90%, 1200px );
 	}
 }
--- a/newspack-theme/sass/media/_captions.scss
+++ b/newspack-theme/sass/media/_captions.scss
@@ -38,7 +38,7 @@ figcaption,
 	}
 
 	.alignfull > figcaption {
-		width: min( 87%, 780px );
+		width: min( 88vw, 780px );
 	}
 }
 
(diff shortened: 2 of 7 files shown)


dkoo (changes requested):
I ran into a few instances where the caption still isn't the right width:

1. Full-width blocks at tablet breakpoints (front-end):

Screen Shot 2022-03-30 at 2 12 32 PM

2. Wide-width blocks at tablet breakpoints (front-end):

Screen Shot 2022-03-30 at 2 16 16 PM

3. Wide-width blocks in the editor:

Screen Shot 2022-03-30 at 2 17 23 PM
 
4. Center-aligned blocks, and blocks with no alignment, in the editor:

Screen Shot 2022-03-30 at 2 19 12 PM

laurelfulford:
Thanks @dkoo!

For the full width issues, it looks like the culprit is `100vw` vs `100%` and the scrollbars -- I had my scrollbars set to always on and had offset the caption size to account for that without realizing it. It really doesn't work! I'll have to revisit. 

I didn't actually make any changes to other caption alignments or the editor styles, but those aren't great 😕 Thanks for flagging them! I might just focus on this alignfull issue first and circle back to those in separate PRs, though -- while they don't look 100%, at least the caption is still readable! :)

laurelfulford:
Ok, I think I've landed on a fix for this -- the downside is that it's a bit more to test 😬 Some details about the issue and approach I went with.

### The issue:

The theme is set up so all content has a max-width of 90% on smaller screens. 

The alignfull blocks use 100vw to make sure they fill the available screen size (they're inside of narrower containers, so we can't use percentages for these and actually have them match the browser window width consistently).

Unfortunately there's a slight discrepancy between 100% and 100vw when scrollbars are visible (when turned on on Mac under Settings > General, or best I can tell, most/all the time on Windows): 100vw includes the width of the scrollbar, so it ends up being kind of like 100% + ~15 or so pixels, rather than just 100%. This is why the 90% caption inside of the 100vw block did not line up with the 90% content inside of a 100% div -- it ended up being a hair wider.

### The fix:

After messing around with this a bit, I landed on changing the `max-width` value to `90vw` theme-wide, at least for the 'structure' elements (basically containers used to set a max-width in the header, content and footer). It gives a more consistent foundation across the content/blocks, and only changes the appearance of the site in really specific circumstances (if viewing the site with sidebars on a mobile-sized window, the content is a hair wider than it would be on devices without the scrollbars -- think a max width of about 92% rather than 90%). 

The main downside is that things should probably still be tested more generally -- I've listed the specific things I changed and where they need to be tested below.

### The alternative fix:

The other approach I considered was reworking styles for blocks with "external" captions, like image, video, audio, embed... so the containing `alignfull` element didn't have the 100vw, just the image, video, audio, embed did. The main issue I felt with this one was that the styles could get convoluted between these blocks and other block that support `alignfull` like group, cover and gallery. There's also not like a fixed set of blocks this could include -- new blocks can be added in Gutenberg, or introduced by plugins, and we'd potentially have to keep updating this approach.

### New things to test (beyond captions)

The original alignfull captions should be fixed with these changes, as well as the alignwide and editor issues you reported. On top of that, here are the specific spots I made changes to the max-widths:

1. `.wrapper` -- this class is used in each level of the header and footer to maintain a consistent width on smaller screen sizes. I think clicking around the site and adjusting the browser width back and forth a few times should work for testing:

Examples of where the `.wrapper` appears in the header and footer: 

image

image

2. `#primary` -- this is like `.wrapper` but for the content area, and is used on all pages:

image

3. `.tec-wrapper` -- this is similar to `.wrapper` and `#primary`, except used on the Events Calendar template markup. It can be viewed by picking the 'Default Events Template' under the Events Calendar Settings > Display, and viewing any events landing page or single event**:

image

_** Note that the single events don't have 1:1 width with the other theme templates on desktop-sized screens-- it uses the default max-width from the plugin rather than the 1200px the theme uses, and I didn't want to make it narrower (crowding the event details) or wider (making the line length even longer!)._ 

4. `.featured-image-above figcaption` -- this sets a max-width of the caption when using the featured image above placement:

image


5. `.featured-image-behind .entry-header` -- this sets a max-width for all the header content when using the featured image behind placement. This change actually moves the post title to the left, but to line it up with the left-edge of the header/content on larger screens, instead of having it float slightly off -- so I think it's actually an improvement:

Before:

image

After:

image

The last three are a little convoluted, the max-width is moved from the `.wrapper` to its parent `.middle-header-contain` and then back again, depending on what the 'featured image beside' settings are paired with. I added some steps to test to help narrow down what changed, but let me know if anything is unclear!

6. `min-width: 782px — .h-sub.single-featured-image-beside .middle-header-contain` -- this selector's a bit of a mess, so here are specific steps to test what it changes:
    * Under Customizer > Header Settings > Subpage Header, enable the Simplified Subpage Header
    * Create or navigate to a post with the single featured image beside setting.
    * Confirm that the header's max-width links up with other elements when moving from desktop to tablet-sized screens.

7. `min-width: 782px — .h-stk.h-sub.single-featured-image-beside div.middle-header-contain .wrapper` -- this selector's also a bit of a mess, so here are specific steps to test what it changes:
    * Under Customizer > Header Settings > Subpage Header, enable the Simplified Subpage Header
    * Under Customizer > Header Settings > Appearance, enable the Sticky header
    * Create or navigate to a post with the single featured image beside setting.
    * Confirm that the header's max-width links up with other elements when moving from desktop to tablet-sized screens.

8. `min-width: 782px — .h-sub-cta.h-sub.single-featured-image-beside div.middle-header-contain .wrapper` -- this selector's also also a bit of a mess, so here are specific steps to test what it changes:
    * Under Customizer > Header Settings > Subpage Header, enable the Simplified Subpage Header
    * If enabled, disabled the Sticky header under Customizer > Header Settings > Appearance.
    * Under Customizer > Header Settings > Mobile CTA, check "Show Mobile CTA in Simplfied Subpage Header"
    * Create or navigate to a post with the single featured image beside setting.
    * Confirm that the header's max-width links up with other elements when moving from desktop to tablet-sized screens.

-----

Please let me know if you have any questions at all, or if this approach seems like overkill -- thanks!

dkoo (approved):
The issues I saw with the wide/full captions at tablet breakpoints look resolved to me, and all of the new test cases also looked good. 🎉

laurelfulford:
Thanks for taking on all that extra testing, @dkoo!

PR #1762, opened 2022-04-08
Add inline style to subscribe patterns

thomasguillot:
### All Submissions:

* [x]  Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

This PR adds (basic) styles for Subscribe Patterns 7 to 9 when used as an Inline Prompt (note: Pattern 8 should not really be used as an inline...)

1. Switch to this branch
2. Add patterns 7 to 9
3. Do they look OK as an inline prompt 

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_patterns.scss
+++ b/newspack-theme/sass/blocks/_patterns.scss
@@ -67,7 +67,10 @@
 		margin: -32px;
 	}
 
+	.newspack-inline-popup &,
 	.newspack-lightbox & {
+		margin: 0;
+
 		~ .popup-not-interested-form {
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
The differences seem mostly subtle to me, but I'm not sure I like the varying type treatments and placements of the suppression form. It can result in some unpredictable behavior if you add anything to the patterns, e.g. adding a single paragraph after Pattern 8:

Screen Shot 2022-04-12 at 11 26 58 AM

I'd suggest we retain the default styles for the suppression form, but everything else looks okay to me at various breakpoints.

PR #1775, opened 2022-04-20
Make sure captions can inherit text colours

### LINKED ISSUE #1220: Photo Credit: can't change colour
You can change the inline colour of a photo caption, but the credit -- coming from XXX and styled by the theme -- can't be changed. 

It should either pick up the containing block colour (in cases where it's in, say, a Group block, and you can change the block's text colour), or the adjacent caption if it has a colour applied (it will use the class `.has-inline-color`), if possible.

laurelfulford:
### All Submissions:

You can currently use the inline colour options to change the caption text colour, but because you can't see/select the credit in the editor, you can't change its colour using this approach. 

This fix isn't perfect, it just provides a work around -- it adds styles so if you wrap the image in a group block with a text colour set, the caption, credit and link will pick it up. It's a small thing, but helps make sure the text can be made legible against dark backgrounds.

When I originally wrote the issue I had it in my head that we could somehow use the `mark` tag that's added with inline colours to also change the credit -- while this could work for any of the theme defined colours (we could use `mark.has-primary-color + .image-credit`) , it won't work for custom colours. They have a class we can target (`mark.has-text-color + .image-credit`), but no way to "tell" what the colour should be -- I figured it "kind of" working like that would be confusing, but I'm open to feedback about that!

Closes #1220.

1. Add an image block with caption and a credit, and a gallery block with per-image captions and credit, and a block level caption (this is to make sure we don't mess with the white-text-on-dark-background that's used over the images).
2. Wrap in a group block and assign it a text colour. 
3. View in the editor and front-end; note that the captions remain grey (or white, in the gallery block's case!):

image

image

5. Confirm that the captions against the white background pick up the group block colour in the editor and on the front end, but that the ones overlaid on images in the gallery are still white:

image

image

### Other information:

### DIFF
--- a/newspack-theme/sass/media/_captions.scss
+++ b/newspack-theme/sass/media/_captions.scss
@@ -16,6 +16,13 @@ figcaption,
 	text-align: left;
 }
 
+.has-text-color {
+	figcaption,
+	.wp-caption-text {
+		color: inherit;
+	}
+}
+
 figcaption,
 .wp-caption-text,
 .amp-image-lightbox-caption {
--- a/newspack-theme/sass/style-editor-base.scss
+++ b/newspack-theme/sass/style-editor-base.scss
@@ -130,6 +130,13 @@ figcaption,
 	padding-right: 0;
 }
 
+.has-text-color {
+	figcaption,
+	.wp-caption-text {
+		color: inherit;
+	}
+}
+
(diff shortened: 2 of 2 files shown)


adekbadek (approved):

laurelfulford:
Thanks Adam!

PR #1792, opened 2022-05-03
fix(sponsors): ensure sponsor bios appear above Related Posts module

dkoo:
### All Submissions:

Ensures that the native sponsor bio appears above the Jetpack Related Posts module in article content. Instead of appending the sponsor bio after the content, in place of the Author Bio, this change injects the sponsor bio to the end of the post content using a `the_content` filter callback.

This change means that the sponsor bio will now appear above the tag footer, pagination links, the Newspack entry footer, the `article-2` sidebar, and other elements that normally appear after post content in Newspack sites. @laurelfulford, is this okay in your mind?

Closes https://github.com/Automattic/newspack-sponsors/issues/61.

1. Enable Jetpack's Related Posts module and apply a native sponsor to a post.
2. On `master`, observe that the sponsor bio appears below the Related Posts module.
3. Check out this branch and confirm that the sponsor bio now appears above Related Posts.

### Other information:

### DIFF
--- a/newspack-theme/inc/newspack-sponsors.php
+++ b/newspack-theme/inc/newspack-sponsors.php
@@ -251,6 +251,8 @@ endif;
 if ( ! function_exists( 'newspack_sponsor_footer_bio' ) ) :
 	/**
 	 * Outputs the 'bio' for the sponsor.
+	 * Appends it to post content to ensure that it appears before other injected modules,
+	 * such as Jetpack's Related Posts module.
 	 */
 	function newspack_sponsor_footer_bio( $sponsors = null, $id = null, $scope = 'native', $type = 'post' ) {
 		$sponsors = newspack_get_all_sponsors(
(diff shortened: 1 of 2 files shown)


laurelfulford (approved):
This works as described for me! Checked non-sponsored, sponsored and underwritten content, and the bio only appeared 'in' the content when the post was sponsored. 

> and other elements that normally appear after post content in Newspack sites. @laurelfulford, is this okay in your mind?

I think this new behaviour makes more sense than how it currently works, to kind of 'bookend' the sponsored content -- it helps mark a clearer end.

The only think that comes to mind for me is that we've been asked about doing this for regular bios, too (well, reordering pretty much every post footer element has come up at one point or another) 😅 So I'm a little worried we could see a spike in that request, but it definitely doesn't outweigh the benefits of making this change. Worst case, if we get a good push for moving the regular author bio, we can look into adding an option. I also feel there's a good argument for continuing to show Related Posts before bios on regular non-sponsored posts to encourage people to stay on the site. 

All that to say, :shipit:

dkoo:
> The only think that comes to mind for me is that we've been asked about doing this for regular bios, too (well, reordering pretty much every post footer element has come up at one point or another) 😅

This change could serve as a template for how we might do that. Because the Related Posts module is appended to post content via a `the_content` filter, anything we want to move above it would need to be done the same way with a priority lower than `40`. We can control the order by tweaking that priority number on each filter callback.

laurelfulford:
> This change could serve as a template for how we might do that. 

Absolutely! I'd gotten stuck on the idea of moving the Related Posts themselves, rather than the Bio, and we've been blocked by the Related Posts shortcode not being AMP-compatible -- this is a good alternative! 😄

PR #1808, opened 2022-05-19
Dropdown menu accessibility improvements

### LINKED ISSUE #1756: Accessibility improvements for dropdown menus
The dropdown menu markup and attributes require some improvements to make them easier and logical to use when navigating the site via keyboard: 

The dropdown menus themselves:

* The dropdowns are always visible to screen reader users, and always receive keyboard focus.
    * To fix: Hide these elements from all users when the associated menu is collapsed.
* These elements control expandable content, but this is not indicated to screen reader users.
    * To fix: Add an `aria-controls` attribute to the elements that control the expandable content and set the values of those attributes to equal the `id`s of the tags that wrap the expanded content. Also add an `aria-expanded="true"` attribute to those same control elements when content is expanded, and `aria-expanded="false"` otherwise.
* These navigation menus expand automatically on focus.
    * To fix: Ensure that these menus are only expanded when users trigger an element to expand them (like by clicking the button). The menus can also be visible on hover, just not on focus. 

The `button`s next to each link with a dropdown menu:

* Each of these buttons has a negative `tabindex` attribute -- this causes them to not to be accessible when tabbing through the page. 
    * To fix: Remove the `tabindex` attributes from these elements so that they can be focused by the keyboard when navigating with the `Tab` and `Shift + Tab` keys. Note that this change will require other changes to make the menus open when the button is clicked. 
* The buttons have the ARIA attribute `role="presentation`, which is unnecessary and can cause issues; this should be removed.

Part of the challenge with some of these changes is making sure they work with AMP.

laurelfulford:
### All Submissions:

This PR makes some accessibility improvements to the theme's dropdown menus that came out of a third-party accessibility audit of a publisher's site:

* **Problem:** the dropdowns are always visible to screen readers because it's not truly hidden when closed, it's just pushed to the left -- because of this it can still receive keyboard focus.
**Solution:** hide these elements using `display: none` when the dropdown isn't open.

* **Problem:** elements menu control expandable content (the dropdowns), but this isn't indicated in a way that screen readers can understand.
**Solution:** add `aria-controls` attribute sto the elements that control the dropdowns, and add unique IDs to the dropdowns. The value of each `aria-controls` attribute should match the ID of the menu it opens . Also add an `aria-expanded="true"` attribute to those same control elements when the dropdown is open, and `aria-expanded="false"` when it's closed.

* **Problem:** dropdowns expand automatically when their parent menu item is focused on, so you can't "skip" them when navigating the site via keyboard.
**Solution:** use the buttons that are already included in the menu markup to control whether or not the menu is visible, rather than just showing it on focus (the dropdown can still be visible on hover, though!). 

* **Problem:** each of the buttons in the menu has a negative `tabindex` attribute -- this causes them to not to be accessible when tabbing through the page; have the ARIA attribute `role="presentation"`, which is unnecessary and can cause issues.
**Solution:** remove the `tabindex` attributes from these elements so that they can be focused by the keyboard when navigating with the Tab and Shift + Tab keys, and remove the `role="presentation"` attribute.

Closes #1756

1. Set up a site with multi-level menus in the Primary and Secondary menu locations. 
2. View the front end of the site and navigate around with your keyboard -- note that when you hit a menu item with a dropdown, it will open on focus, and that you have to tab through each level. 
4. Repeat step 2; note that now on menu items with dropdowns, you can tab to the link and arrow individually; when you tab past a menu items with a dropdown, it doesn't expand automatically.
5. Tab to one of the down arrows and hit enter; confirm that the menu can be opened and closed this way, and that the arrow flips to reflect its state. Travel through different levels of the menu confirming this behaviour:

menu-tabbing 2022-05-18 17_40_22

6. Open the element inspector and confirm that each of the arrow buttons has:
     * an `aria-controls="submenu-####"` value, where the number matches the ID `menu-item-####` value of the `li` directly above it 
     * an `aria-expanded="false"` value that switches to `true` when you tab to the button and hit 'Enter`
     * an `aria-haspopup="true"`
7. In the element inspector, confirm that each dropdown's `ul` has an attribute `id="submenu-####"`, where the full value matches the `aria-controls` value of the button before it. 
8. Try using your mouse to navigate the menu and make sure the dropdowns open on hover, and that you can't click on the down arrows with the mouse to open the dropdowns (this was the best way I could think of to avoid accidentally mixing and matching menu open states)
9. Turn off AMP (or on AMP, depending on what your test site's default is) and repeat steps 4-8 and confirm things still work as expected. 

Note: the approach I used to add IDs to the submenus may be sub-optimal -- it seems kind of 'heavy duty' for me, but at the same time it was the one way I could figure out to pass the parent menu ID, but I feel like it's more functional than "correct". So any feedback there on top of with how this all works is hugely appreciated! 

### Other information:

### DIFF
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -412,29 +412,6 @@ function newspack_get_discussion_data() {
 }
 
 /**
- * WCAG 2.0 Attributes for Dropdown Menus
- *
- * Adjustments to menu attributes tot support WCAG 2.0 recommendations
- * for flyout and dropdown menus.
- *
- * @ref https://www.w3.org/WAI/tutorials/menus/flyout/
- */
-/*
-function newspack_nav_menu_link_attributes( $atts, $item, $args, $depth ) {
-
-	// Add [aria-haspopup] and [aria-expanded] to menu items that have children
-	$item_has_children = in_array( 'menu-item-has-children', $item->classes );
-	if ( $item_has_children ) {
-		$atts['aria-haspopup'] = 'true';
-		$atts['aria-expanded'] = 'false';
-	}
-
-	return $atts;
-}
-add_filter( 'nav_menu_link_attributes', 'newspack_nav_menu_link_attributes', 10, 4 );
-*/
-
-/**
  * Add a dropdown icon to top-level menu items.
  *
  * @param string $output Nav menu item start element.
@@ -459,7 +436,7 @@ function newspack_add_dropdown_icons( $output, $item, $depth, $args ) {
 		$toggle_id = "toggle_" . $item->ID ;
 
 		$output .= sprintf(
-			'<button aria-controls="submenu-'. $item->ID . '" aria-expanded="false" class="submenu-expand" [class]="' . $toggle_id . ' ? \'submenu-expand open-dropdown\' : \'submenu-expand\'" [aria-expanded]="' . $toggle_id . ' ? \'true\' : \'false\'" on="tap:AMP.setState( { ' . $toggle_id . ' : !' . $toggle_id . ' } )">%s</button>',
+			'<button aria-controls="submenu-'. $item->ID . '" aria-expanded="false" class="submenu-expand" [class]="' . $toggle_id . ' ? \'submenu-expand open-dropdown\' : \'submenu-expand\'" [aria-expanded]="' . $toggle_id . ' ? \'true\' : \'false\'" on="tap:AMP.setState( { ' . $toggle_id . ' : !' . $toggle_id . ' } )" aria-haspopup="true">%s</button>',
 			$icon
 		);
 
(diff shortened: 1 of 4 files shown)


adekbadek (approved):
> Note: the approach I used to add IDs to the submenus may be sub-optimal -- it seems kind of 'heavy duty' for me, but at the same time it was the one way I could figure out to pass the parent menu ID, but I feel like it's more functional than "correct". So any feedback there on top of with how this all works is hugely appreciated!

Looks ok to me! It would be more elegant if this was a part of a class, then it could use a class variable instead of a real global – but it gets the job done and seems not to cause any issues.

laurelfulford:
Thanks @adekbadek!

> It would be more elegant if this was a part of a class, then it could use a class variable instead of a real global

TIL -- I knew I must be missing something! 😄 

I took a swing at this in [f88c028](https://github.com/Automattic/newspack-theme/pull/1808/commits/f88c028af7883c49ec1e5dab918b68c4253e74dc) -- would you mind taking a second look before I commit, just to make sure I didn't do anything silly? Thank you!

laurelfulford:
Awesome! Thanks for the guidance, @adekbadek!

PR #1826, opened 2022-06-01
Add blank template to theme

### LINKED ISSUE #1819: Add Blank template
ActiveCampaign includes a page scraper option that would allow publishers to create a page with a Homepage Posts block (so that content is dynamically updated) and AC will scrape the content on a schedule. Allows publishers to essentially setup curated newsletters without having to manually create them. 

But ActiveCampaign requires that the page be fully blank -- no header or footer. [Example on private page](https://evanstonnow.com/newsletter-scrape/).

1. Add a "Blank" template option OR
2. Add a toggle to disable header, menus and footer.

laurelfulford:
### All Submissions:

This PR adds a blank template to the theme. It doesn't have a header or footer, any hooks around the header or footer (like after header and before footer), and it doesn't have any widget areas. 

Some general notes:

* This template is only available for pages, not other post types. 
* This template uses the same layout base as the one-column template: the content is 780px wide, which will work well with paragraphs of text; wide and full aligned blocks can be used to make the content wider (for things like information tables). 
* This template displays the page title (though it can be hidden in the page's settings). 
* This template includes all of the featured image settings (though you can also opt to hide them). 

I made the above decisions based on the templates use case, consistency with how the other templates work, and flexibility. I imagine as we auto-create pages with it -- like we do for the "Support Our Publication" page -- we'll likely want to change certain settings (hide the title and featured image, and turn off campaigns).

Closes #1819

2. Create a new page; confirm you have a template called "Blank template" and apply it to the new page. 
3. Add some content to the page, and publish.
4. Confirm that the page doesn't have a header or footer on the front-end. 
5. Cycle through the different featured image placements; confirm they look similar to how they look with the One Column template.

<details><summary><strong>Screenshots</strong></summary>

Small featured image:
image

Large featured image:
image

Featured image behind:
image

Featured image beside:
image

Featured image above:
image

</details>

6. Hide the page title and featured image; publish and view on the front-end -- regular paragraphs shouldn't totally hit the top of the screen but should be close-ish:

image

7. Edit the page and wrap all of the content in a full-width group block and give it a background colour; publish.
8. Confirm that the group block touches the top of the browser window; if you open it in an incognito window, it should also touch the bottom (by default, the theme's edit link will be in the way when you're logged in; it was also left for consistency, but I'm open to feedback about that!):

image

### Other information:

### DIFF
--- a/newspack-theme/blank.php
+++ b/newspack-theme/blank.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Template Name: Blank Template
+ * Template Name: Blank template
  *
  * @link https://developer.wordpress.org/themes/basics/template-files/#template-partials
  *
--- a/newspack-theme/blank.php
+++ b/newspack-theme/blank.php
@@ -27,13 +27,36 @@
 		<section id="primary" class="content-area">
 			<main id="main" class="site-main">
 				<?php
+
 				/* Start the Loop */
 				while ( have_posts() ) :
 					the_post();
+
+					// Template part for large featured images.
+					if ( in_array( newspack_featured_image_position(), array( 'large', 'behind', 'beside', 'above' ) ) ) :
+						get_template_part( 'template-parts/post/large-featured-image' );
+					else :
 					?>
+						<header class="entry-header">
+							<?php get_template_part( 'template-parts/header/entry', 'header' ); ?>
+						</header>
+					<?php endif; ?>
 
 					<div class="main-content">
-						<?php get_template_part( 'template-parts/content/content', 'page' ); ?>
+
+						<?php
+						// Place smaller featured images inside of 'content' area.
+						if ( 'small' === newspack_featured_image_position() ) {
+							newspack_post_thumbnail();
+						}
+
+						get_template_part( 'template-parts/content/content', 'page' );
+
+						// If comments are open or we have at least one comment, load up the comment template.
+						if ( comments_open() || get_comments_number() ) {
+							newspack_comments_template();
+						}
+						?>
 					</div>
 
 				<?php endwhile; ?>
--- a/newspack-theme/blank.php
+++ b/newspack-theme/blank.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Template Name: No Header or Footer
+ * Template Name: Blank Template
  *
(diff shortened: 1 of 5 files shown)


dkoo (changes requested):
> Confirm that the group block touches the top of the browser window; if you open it in an incognito window, it should also touch the bottom

This is the only thing that didn't work for me—there's still a gap at the bottom of the page even while not logged in. Other than that, looks good!

laurelfulford:
Thanks @dkoo! 

> This is the only thing that didn't work for me—there's still a gap at the bottom of the page even while not logged in. Other than that, looks good!

🤔 The only way I could recreate this was if the page content was shorter than the browser window -- and in that case, I'm not sure there's a good way to work around it (that won't cause extra scrolling on smaller screeens).

Does this line up with your experience, too? If no, could you please share the content you're testing with and I can give it a spin? Thank you!

dkoo (approved):
I figured out the issue, and it doesn't have anything to do with these changes—it's just that the test case with the single Group block and no footer underneath the content causes the issue to be visible. Basically, my site had some inline prompts that weren't being displayed due to segmentation, so they were completely hidden. However, [this CSS](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/sass/blocks/_blocks.scss#L6) is still applied to the hidden `aside` element which appears at the very end of the content, resulting in a small gap at the bottom of the page.

This should ideally be fixed by either excluding `aside` elements from that `*` selector, or by overwriting the CSS with a more specific selector in the Campaigns plugin.

laurelfulford:
Great catch -- thanks @dkoo! 

> This should ideally be fixed by either excluding aside elements from that * selector, or by overwriting the CSS with a more specific selector in the Campaigns plugin.

I might be kind of split between the two -- the extra spacing is the theme's doing, and it does make sense to have it when an inline campaign is visible. It just doesn't really work when it's not.

Would it work to give the `<aside>` a class or `data-` attribute to "say" what state the campaign is in, and then update the theme's CSS to remove the space when it's not visible? As is, it's not possible to "tell" if the campaign is visible by CSS alone -- or even if this `<aside>` is wrapped around a campaign or some other unrelated content -- but a class or other attribute could help there!

Just let me know what you think and I can spin up some issues -- thanks!

PR #1827, opened 2022-06-01
Remove FSE-related blocks from editor

### LINKED ISSUE #1823: Blocks: Remove/hide more 'confusing' FSE blocks
As FSE features grow, so do the number of related blocks. Some of these blocks are pretty confusing when they appear with Newspack blocks -- some specific recent examples are the Author Bio (which even uses the same icon as ours, but is meant to appear in a Query block/pick up the current author, whereas ours lets you pick an author, and has Co-Author Plus support) and the Read More link. 

We should audit all of the blocks and hide the more confusing ones from this theme. They will make much more sense in the future block-based version of the Newspack theme, but right now, they cause more confusion than help. 

Note that there are some blocks that plugins like Newsletters use that the theme doesn't necessarily need to be available, but that also aren't really as confusing as others (like Site Logo).

Edited to add: the Site Logo, site title, etc. could also be helpful when paired with a [blank template](https://github.com/Automattic/newspack-theme/issues/1819).

laurelfulford:
### All Submissions:

This PR removes most of the full-side editing related blocks from the editor. This is to help reduce confusion between Newspack-related functionality and these blocks, and to also remove some unnecessary noise from the editor. 

Related: pamTN9-4uP-p2

Closes #1823.

2. Edit a post or page and try to add any of the following blocks, or find them in the block picker:

* Login/Out
* Post Comments Form
* Comments Query Loop
* Query Loop
* Posts List
* Post Title
* Post Featured Image
* Post Excerpt
* Post Content
* Post Tags
* Post Categories
* Post Date
* Post Author
* Previous and Next Post links
* Read More
* Avatar
* Post Author Biography
* Archive title
* Term Description

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -591,6 +591,9 @@ function newspack_enqueue_scripts() {
 		newspack_get_post_toggle_post_types()
 	);
 	wp_enqueue_script( 'newspack-post-meta-toggles' );
+
+	// Remove FSE-related Gutenberg blocks.
+	wp_enqueue_script( 'newspack-hide-fse-blocks', get_theme_file_uri( '/js/dist/editor-remove-blocks.js' ), array( 'wp-blocks', 'wp-dom-ready', 'wp-edit-post' ), $theme_version, true );
 }
 add_action( 'enqueue_block_editor_assets', 'newspack_enqueue_scripts' );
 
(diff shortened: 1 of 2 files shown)


laurelfulford:
Just a note that I'm also going to explore doing this but from a 'turning off in the UI by default' approach, which will not remove the code or make the blocks not work, it will just disable them visually by default.

laurelfulford:
Opted to stick with this approach for now, so marking "ready for review".

miguelpeixe (approved):
Confirmed this unregisters the blocks!

I wonder if the `query-loop` block should be unregistered by `blocks` or `plugin`, given it's a "duplicate" of Homepage Posts block. Not necessarily related to `theme`. If it's the case, can be tackled separately.

PR #1828, opened 2022-06-02
fix: update color and font code to support navigation block

### LINKED ISSUE #1825: Navigation Block: Custom colours not working
**Describe the bug**
The Navigation block now lets you change the colour of the menu and backgrounds; in the editor, the primary and secondary colours are previewing the defaults though (blue and grey), and the greys aren't actually applied to the dropdowns on the front-end.

image

laurelfulford:
### All Submissions:

The Navigation Block wasn't correctly inheriting some of the theme's baked in colours into the dropdowns. This PR tweaks the colour styles so the dropdowns pick them up.

Note: there is an option to separately change the submenu colour in the Navigation block, but I've had mixed results in getting that to work that I can't quite pin down: the CSS classes aren't always added to the dropdowns, so it's not that the styles aren't working so much that the classes aren't applied. It seems to be falling back to the menu's main colour when the classes aren't applied -- I'm going to keep trying to figure out reproduce-able steps for this issue so I can report it (or perhaps confirm that it's just something weird in my local environment). If you do run into issues with the submenu styles when testing this PR, you can confirm whether or not you're running into the same problem by inspecting the dropdown menu, and seeing if the `li` in the menu have CSS classes. This is what it looks like when it's working: 

image

Closes #1825.

1. Add a few navigation blocks to a page, each with at least one dropdown; apply the following settings to each menu:
* One without a background colour. 
* One with a background colour using one of the custom colours (the primary or secondary colours).
* One with a background colour using one of the shades of grey (colours 5, 6, or 7 in the palette).
3. Publish and view the front end. 
4. Open the submenus; note that the one with the grey background has transparent dropdowns:

image

5. Also note that the block uses the body font, compared to the theme's menus using the heading font, and the links all have underlines.
6. Apply the PR and run `npm run build`
7. Confirm that the dropdowns now pick up the grey background, and the menus use the heading font and are no longer underlined:

image

8. Confirm that your other dropdowns display as expected (picking up the parent menu's background colour, or are white if no parent background colour is set).

### Other information:


adekbadek (approved):
👐

PR #1834, opened 2022-06-03
Search: make sure the search forms have a predictable unique ID

### LINKED ISSUE #1832: Search: Focus is not always set in the search input when opened
**Describe the bug**
When you open the search in the theme's header, focus should move to the input. However, this doesn't always work with the AMP version of the code. 

The toggle was written in such a way to "expect" the ID `search-form-2` (assuming it's the second form on the page's markup, after the one in the mobile menu); it looks like expecting that specific ID is not dependable, and it can be a higher number. 

This needs to be updated to behave more dependably, without relying on the autogenerated ID that comes from `wp_unique_id`, which can be used by more than the search.

laurelfulford:
### All Submissions:

The theme originally used `wp_unique_id()` to generate a number for a unique ID for each search form the theme output. 

For the search in the header, the ID is required by AMP to set focus on the search input when the search is opened. 

We can predict the search number ID when on the Newspack theme is active (it's the second search the theme outputs, so the ID is `wp_unique_id()`. However, plugins also use `wp_unique_id()` (including the AMP plugin!), so it's not always possible to predict what ID number the search in the header will end up with. 

This PR adds a function to the theme to generate an ID just for the search forms, rather than using a more universal WordPress one, since we (unfortunately!) require the ID over, say, proximity to select the right input.

Closes #1832.

1. On your test site viewing an AMP version of the site, open the search field; odds are, on open the search field does not receive focus (if it does, the ID will be using the `-2`; in that case you can recreate this issue on sites like YubaNet and MTFP). 
2. Apply the PR. 
3. Repeat step one; confirm that the search field gets focus when the search is opened (if this already worked for you, just confirm that it still works!). 

### Other information:

### DIFF
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -695,6 +695,16 @@ function newspack_should_display_updated_date() {
 }
 
 /**
+ * Create a predictable unique ID for the search forms.
+ *
+ * @param string $prefix Text to prepend the ID with.
+ */
+function newspack_search_id( $prefix = '' ) {
+	static $id_counter = 0;
+	return $prefix . ( string ) ++$id_counter;
+}
+
+/**
  * Check whether there's a Post Summary, and return it.
  */
 function newspack_has_post_summary() {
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
🙌

laurelfulford:
Thanks @adekbadek!

PR #1840, opened 2022-06-07
Add checkbox style

thomasguillot:
### All Submissions:

The theme currently doesn't have custom style for checkboxes, this PR adds a simple checkbox style.

Note: I saw WooCommerce has some custom styles already (#ship-to-different-address) but I wasn't able to find out how to test it

1. Switch to this branch
2. Navigate to a screen that has a checkbox (e.g. yoururl.com/product/donate/)
3. Update button's background colour and play with text colour contrast (black/white)

### Other information:

### DIFF
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -565,8 +565,8 @@ function newspack_get_color_contrast( $hex ) {
  */
 function newspack_color_with_contrast( $color ) {
 	$contrast = newspack_get_color_contrast( $color );
-	if ( '#000' === $contrast ) {
-		return '#5a5a5a';
+	if ( 'black' === $contrast ) {
+		return 'dimgray';
 	}
 	return $color;
 }
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -223,7 +223,7 @@ function newspack_body_classes( $classes ) {
 			$classes[] = 'ad-above-footer';
 		}
 	}
-	
+
 	// Add a class for the footer logo size.
 	$footer_logo_size = get_theme_mod( 'footer_logo_size', 'medium' );
 	if ( 'medium' !== $footer_logo_size ) {
@@ -553,10 +553,10 @@ function newspack_get_color_contrast( $hex ) {
 	}
 	if ( $contrast_ratio > 5 ) {
 		// If contrast is more than 5, return black color
-		return '#000';
+		return 'black';
 	} else {
 		// if not, return white color.
-		return '#fff';
+		return 'white';
 	}
 }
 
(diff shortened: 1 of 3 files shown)


laurelfulford (changes requested):
Overall this worked well in my testing! The colours were picked up as expected, and the contrast worked well. 

The only issue I can see with this change is that it changes the `input['checkbox']` from being inline to being block with `display: grid`. We can fix where this affects things in the theme, like the comment form:

image

... but it will also affect markup from any plugins sites use, which could be a pain to correct. Off the top of my head, MC4WP "assumes" that the checkbox is inline so we would see some wider varied breakage from that ([example at the bottom of this page](https://wisconsinwatch.org/)).

Switching the checkbox to `display: inline-grid` seems to help in my testing -- will that cause other issues I'm not thinking of?

> Note: I saw WooCommerce has some custom styles already (#ship-to-different-address) but I wasn't able to find out how to test it

This appears on the checkout page -- in my testing, it looked and acted exactly the same with and without this PR applied: 

image
image

newspack-theme/inc/template-functions.php
```diff
@@ -553,10 +553,10 @@ function newspack_get_color_contrast( $hex ) {
 	}
 	if ( $contrast_ratio > 5 ) {
 		// If contrast is more than 5, return black color
-		return '#000';
+		return 'black';
```
laurelfulford, line 556: Apologies, I missed this the first time though! 

The output of this function is also used to figure out whether to switch the primary or secondary colour to a medium grey when they don't have sufficient contrast against white: 

https://github.com/Automattic/newspack-theme/blob/984644407faa7b30acc38752f2cfbac971cb4f3b/newspack-theme/inc/template-functions.php#L590

`newspack_color_with_contrast()` should also be updated to check for `black`.

thomasguillot: Note: To find the closest match, I used this brilliant tool: https://stubbart.com/computer_consulting/color_themes/color_closest.html

laurelfulford (approved):
This looks good, @thomasguillot!

> which is more generic and offers a slightly better contrast

Iiiii have to ask about this because I feel like I'm missing something 😅  `dimgray` comes back as `#696969` for me, or a contrast ratio of 5.48:1 against white, compared to 6.89:1 with `#5a5a5a` -- so I think the contrast is less? It's not a problem since both pass WCAG AA, I'm just wondering if I've misunderstood this bit -- thanks!

(Totally good to merge, though!)

thomasguillot:
> Iiiii have to ask about this because I feel like I'm missing something 😅 `dimgray` comes back as `#696969` for me, or a contrast ratio of 5.48:1 against white, compared to 6.89:1 with `#5a5a5a` -- so I think the contrast is less? It's not a problem since both pass WCAG AA, I'm just wondering if I've misunderstood this bit -- thanks!

🤦 

My bad! You're totally right.

PR #1875, opened 2022-07-12
feat: colors' css variables and action hook for mobile toggle

miguelpeixe:
### All Submissions:

This PR introduces root CSS variables for the theme's colors and 2 action hooks placed before and after the header mobile menu toggle.

Further details and instructions at https://github.com/Automattic/newspack-plugin/pull/1754

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -910,26 +910,54 @@ function newspack_colors_css_wrap() {
 add_action( 'wp_head', 'newspack_colors_css_wrap' );
 
 /**
+ * Get theme colors' values.
+ *
+ * @return string[] Array of colors.
+ */
+function newspack_get_colors() {
+	$colors              = [];
+	$colors['primary']   = newspack_get_primary_color();
+	$colors['secondary'] = newspack_get_secondary_color();
+	$colors['cta']       = get_theme_mod( 'header_cta_hex', newspack_get_mobile_cta_color() );
+
+	if ( true === get_theme_mod( 'header_solid_background', false ) ) {
+		$colors['header'] = $colors['primary'];
+	}
+
+	if ( 'default' !== get_theme_mod( 'theme_colors', 'default' ) ) {
+		$colors['primary']   = get_theme_mod( 'primary_color_hex', $colors['primary'] );
+		$colors['secondary'] = get_theme_mod( 'secondary_color_hex', $colors['secondary'] );
+
+		if ( 'default' !== get_theme_mod( 'header_color', 'default' ) ) {
+			$colors['header']       = get_theme_mod( 'header_color_hex', '#666666' );
+			$colors['primary_menu'] = get_theme_mod( 'header_primary_menu_color_hex', '' );
+		} else {
+			$colors['header'] = $colors['primary'];
+		}
+
+		if ( 'default' !== get_theme_mod( 'footer_color', 'default' ) ) {
+			$colors['footer'] = get_theme_mod( 'footer_color_hex', '' );
+		}
+	}
+
+	// Set color contrasts.
+	foreach ( $colors as $color_key => $color_value ) {
+		$colors[ $color_key . '_contrast' ] = newspack_get_color_contrast( $color_value );
+	}
+
+	return $colors;
+}
+
+/**
  * Add CSS variables to theme's colors.
  */
 function newspack_colors_css_variables() {
-	$primary_color   = newspack_get_primary_color();
-	$secondary_color = newspack_get_secondary_color();
-	$cta_color       = get_theme_mod( 'header_cta_hex', newspack_get_mobile_cta_color() );
-	$header_color    = $primary_color;
-
-	$colors = [
-		'primary'   => $primary_color,
-		'secondary' => $secondary_color,
-		'cta'       => $cta_color,
-		'header'    => $header_color,
-	];
-
+	$colors = newspack_get_colors();
 	?>
 	<style type="text/css" id="newspack-theme-colors-variables">
 		:root {
 			<?php foreach ( $colors as $color_key => $color_value ) : ?>
-				--newspack-<?php echo esc_attr( $color_key ); ?>-color: <?php echo esc_attr( $color_value ); ?>;
+				--newspack-<?php echo esc_attr( str_replace( '_', '-', $color_key ) ); ?>-color: <?php echo esc_attr( $color_value ); ?>;
 			<?php endforeach; ?>
 		}
 	</style>
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -910,6 +910,34 @@ function newspack_colors_css_wrap() {
 add_action( 'wp_head', 'newspack_colors_css_wrap' );
 
 /**
+ * Add CSS variables to theme's colors.
+ */
+function newspack_colors_css_variables() {
+	$primary_color   = newspack_get_primary_color();
+	$secondary_color = newspack_get_secondary_color();
+	$cta_color       = get_theme_mod( 'header_cta_hex', newspack_get_mobile_cta_color() );
+	$header_color    = $primary_color;
+
+	$colors = [
+		'primary'   => $primary_color,
+		'secondary' => $secondary_color,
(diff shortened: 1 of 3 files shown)


laurelfulford (approved):
Testing this PR with https://github.com/Automattic/newspack-plugin/pull/1754, it's functionally correct -- the button does appear and displays on both desktop and mobile.

I did notice some visual issues (with alignment, spacing, and contrast), but I don't think finalizing that kind of stuff is the purpose of this PR 🙂 Based on that assumption, I've marked this PR as approved.

Please let me know if I have that wrong, and I can do a more thorough visual review! 

I'm also happy to spin up an issue for addressing the styles down the road -- just let me know, thanks @miguelpeixe!

miguelpeixe:
Thank you, @laurelfulford!

I've pushed another change on 31034add1d76b9cfff9fafaa5249dc7ab3866d17 to introduce global CSS variables for the theme's colors. This will help reader activation and other projects to more easily integrate with the theme's UI.

If you can give it a test, it should display the correct applied colors as variables on DevTools:

image

laurelfulford:
Thanks @miguelpeixe! This works as described, but I have some feedback about the approach: 

I'm probably overthinking this one, but I think it might be easier in the long run to handle any theme-related styles for the reader activation stuff in the theme, rather than the plugin. This is because when we release a new theme, things like the primary and secondary colour settings or header colour settings won't exist as they do now -- they'll either be provided by Global Styles, or come from a block setting, rather than a universal spot we can pass like this. If all of this colour code is in the theme, at least there's nothing to override/untangle when this functionality is used with a different theme. (There's no reason not to use variables though!)

In a similar vein, we might be able to make the theme styles more general -- like make sure anything in the header is picking up the correct colours without having to know the classes -- to help reduce the amount of activation-specific CSS we need!

If we do go ahead with this approach, it might be helpful to pass the [contrasting colours](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/inc/color-patterns.php#L36-L40) as well. These colours are either black or white, depending on what would contrast against one of the custom colours -- so one example would be if the header colour is very light, its contrasting colour would be black, and would be used for the header text; if the header colour is very dark, that colour would be white. 

Also, it's an extreme edge case, but the colour styles are only added to a site when something has been customized -- so if you're using the [default colours, default ad colours and default mobile CTA colours](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/functions.php#L966-L968), these variables won't be available. This won't affect any "real" sites, but it could affect demos/our own sites -- it might be worth adding some defaults in one of the theme's Sass files. 

Just let me know if you have questions about either of these!

miguelpeixe:
Thank you for your feedback on the approach!

> when we release a new theme, things like the primary and secondary colour settings or header colour settings won't exist as they do now

If the new theme is going to provide variables such as `--wp--preset--color--primary` from `theme.json`, I believe this would be the most straightforward standard to implement across our entire codebase. If we adopt this approach, updating the codebase to the new variable name would be much quicker than refactoring to stop using PHP variables, javascript, or any other approach that we are currently using for the same goal. WDYT?

> If we do go ahead with this approach, it might be helpful to pass the [contrasting colours](https://github.com/Automattic/newspack-theme/blob/master/newspack-theme/inc/color-patterns.php#L36-L40) as well.

Good call! I'll update to include these variants.

> Also, it's an extreme edge case, but the colour styles are only added to a site when something has been customized

I didn't realize that! I'll also change the approach to have a separate action hook rendering the variables.

laurelfulford:
> If the new theme is going to provide variables such as `--wp--preset--color--primary` from theme.json, I believe this would be the most straightforward standard to implement across our entire codebase.

Ah right, I think that should work! 🙂 The only "gotcha" I think just might be the weird few things the theme does with colours (eg. half use the primary colour as the default header colour; others use shades of grey), but approaching this in more of a future looking way would be best in the long term -- we can always tackle those weird things as they come up! 

For this, it may not make sense to add the theme contrast colours after all? I'm not exactly sure how those would be replaced dynamically in a future FSE theme... though that can also be something we can tackle when it actually becomes a problem. It's a ways in the future at the moment.

miguelpeixe:
> For this, it may not make sense to add the theme contrast colours after all?

It may be the case to adjust/include the existing color variables as they are required. AFAIK, FSE does all that automatically, so we'll soon not have to worry about exposing theme colors at all. It'll be about choosing which and where to use them.

miguelpeixe:
2ea6497cadc30f4dc86534becad4cdde406fd670 moves the color variables to its own hook to not be conditioned to existing customizations.

laurelfulford:
Thanks for updating this, @miguelpeixe! 

I'm not sure it will fully work as expected -- when I change the theme colours, I still end up with mostly the defaults (in my settings, primary, secondary, mobile CTA and header have been changed, but only the mobile CTA variable is different than the default in the CSS variables):

image

Also, the header colour isn't always the primary colour -- it's the default for some of the themes when you turn on the solid background, but then it can be changed, but it doesn't look like that will work with the current set up.

Would it work to set a default for these variables -- even in one of the theme's SCSS files -- and then override them using the original custom colours code as needed?

miguelpeixe:
Thanks, @laurelfulford!

I didn't properly set up the variables used to generate the CSS variables, I was copying a subset of the definition of the variables. For better maintenance, 24e2a3c83341e917c29f4c7547314a999a64d83d isolates the definition of colors to be used by the `color-patterns.php` and by `newspack_colors_css_variables()`.

The CSS variables should now reflect the patterns exactly.

laurelfulford (approved):
Thanks @miguelpeixe! What's here looks good, and is a nice refactor of all that colour code! 

I did run into one issue: each of the child themes also has their own colour patterns PHP under /inc/child-color-patterns.php, and the formatting should be updated to match. Overall I don't think this is an issue that has to be resolved in this PR -- we can refactor all of those in a separate PR -- but it is going to cause one small bug:

Three of the child themes don't use the primary colour as their header background default, they use different shades of grey -- specifically: 

* [Newspack Joseph](https://github.com/Automattic/newspack-theme/blob/master/newspack-joseph/inc/child-color-patterns.php#L13) uses `#111`
* [Newspack Katharine](https://github.com/Automattic/newspack-theme/blob/master/newspack-katharine/inc/child-color-patterns.php#L13) uses `#333`
* [Newspack Sacha](https://github.com/Automattic/newspack-theme/blob/master/newspack-sacha/inc/child-color-patterns.php#L13) uses `#333`

(Originally there wasn't an option to control the header colour at all -- it just either used the primary colour for half the themes, and grey for the other half. This is a weird artifact of that). 

If the header with these themes is set to have a solid background but not a custom colour, the CSS variables will be incorrect -- they will be for the primary colour and its contrasting text colour, rather than the actual header colour and its contrasting text colour. 

In the unlikely chance a site has one of these themes, a solid header and no custom header colour set, the issue could be worked around by setting the header background colour to the same grey as the default. 

Based on the above, I still think the issue can be addressed in a separate PR to help keep things moving. So I'm marking this one as 'approved' -- if you have any hesitation about moving forward with it as is, though, I'm happy to re-review!

miguelpeixe:
Thank you, @laurelfulford!

> we can refactor all of those in a separate PR -- but it is going to cause one small bug:

Where will this bug be visible? I noticed that it'll generate an inconsistency between the created CSS variable and the actual applied color for the child themes.

Although the CSS variables are created for all colors, they are currently only implemented for reader activation buttons/links using the primary color.

laurelfulford:
> Where will this bug be visible? I noticed that it'll generate an inconsistency between the created CSS variable and the actual applied color for the child themes.

It would only be visible if: 

1. You're using Newspack Joseph, Katharine or Sacha
2. You have a solid header background set, but no colour picked so it's using the default dark grey
3. You have a really light primary colour
4. You use the CSS variable generated for the header text colour

The issue would be the CSS variable for the header text colour would be dark (to contrast against the light primary colour), but it should actually be light, to contrast against the default header colour. It's a VERY specific set of circumstances, and pretty unlikely to actually crop up in the wild 🙂 If it did, setting the header background to be a dark grey would fix it!

PR #1918, opened 2022-08-29
Bundle child theme fonts in each theme

### LINKED ISSUE #1856: Fonts: Move default Google Fonts into themes
Currently most of the Newspack themes (excluding the parent themes) load Google Fonts from fonts.google.com.

Copies of the default font options should be moved to the themes instead, to comply with the GDPR. 

See: https://make.wordpress.org/themes/2022/06/18/complying-with-gdpr-when-using-google-fonts

laurelfulford:
### All Submissions:

This PR updates the default fonts used by each of the Newspack child themes, so they're bundled in the theme rather than being loaded from Google Fonts.

I also updated where the editor style sheets are generated to (into the root of each theme rather than into the /styles directory). The sole reason for doing this was so that the `@font-face` path would work for both the front-end and editor styles without having to duplicate that code. 

Closes #1856. 

2. Navigate to Customizer > Typography, and remove any custom fonts. 
3. Cycle through each theme and check the front-end and in the editor, and confirm that the original font still loads for each theme. If you have some/all of the fonts installed locally, they can mess with the testing because if the font doesn't load from the theme it may load from your computer instead -- you can either delete the fonts from your computer to test, or use a service like Browserstack so you're not loading local fonts.

Each theme uses the following fonts:

Newspack Joseph - Old Standard TT for headings, and EB Garamond for the body:
image

Newspack Katharine - Barlow for headings and body:
image

Newspack Nelson - Montserrat for headings, sans-serif system font stack for the body:
image

Newspack Sacha - IBM Plex Serif for headings, serif system font stack for the body:
image

Newspack Scott - Fira Sans Condensed for headings, serif system font stack for the body:
image

### Other information:


adekbadek (approved):
🙌

laurelfulford:
Thanks @adekbadek!

PR #1922, opened 2022-08-31
feat: ensure footer stays at the bottom of the screen

adekbadek:
### All Submissions:

Ensures the footer sticks to the bottom of the viewport. This is a small change in code, but might have big implications for site's styling, so we should proceed with caution. Feedback very much welcome from @laurelfulford :) 

Example with Nelson theme, before:

image

after:

image

1. On `master`, publish a page with just one line of content and using one column template
2. View the front-end, observe the footer is hanging mid-page
3. Switch to this branch, rebuild, observe the footer sticks to the bottom of the viewport

### Other information:

### DIFF
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -10,15 +10,15 @@
 }
 
 .admin-bar #page {
-  min-height: calc( 100vh - 46px );
+	min-height: calc( 100vh - 46px );
 
-  @include utilities.media( tablet ) {
-    min-height: calc( 100vh - 32px );
-  }
+	@include utilities.media( tablet ) {
+		min-height: calc( 100vh - 32px );
+	}
 }
 
 #content {
-	flex: 1;
+	flex: 99999; // Make it take as much space as possible, relative to other flex-grow elements.
 }
 
 .wrapper {
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -7,6 +7,12 @@
 	display: flex;
 	flex-direction: column;
 	min-height: 100vh;
+	> *:not( #content ) {
+		flex: 0;
+	}
+}
+#content {
+	flex: 1;
 }
 
 .admin-bar #page {
@@ -15,10 +21,6 @@
 	@include utilities.media( tablet ) {
 		min-height: calc( 100vh - 32px );
 	}
-}
-
-#content {
-	flex: 99999; // Make it take as much space as possible, relative to other flex-grow elements.
 }
 
 .wrapper {
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -8,6 +8,15 @@
 	flex-direction: column;
 	min-height: 100vh;
 }
+
+.admin-bar #page {
+  min-height: calc( 100vh - 46px );
+
+  @include utilities.media( tablet ) {
+    min-height: calc( 100vh - 32px );
+  }
+}
+
 #content {
 	flex: 1;
 }
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -3,6 +3,15 @@
 
 /** === Layout === */
 
+#page {
+	display: flex;
+	flex-direction: column;
+	min-height: 100vh;
+}
+#content {
+	flex: 1;
+}
+
 .wrapper {
 	display: flex;
 	flex-wrap: wrap;


dkoo (changes requested):
On the main theme, I'm seeing this:

Screen Shot 2022-08-31 at 11 31 41 AM

I wonder if there's a potentially less-far-reaching solution for this? I feel like setting flex properties on the main containers could have unintended consequences.

laurelfulford (approved):
@dkoo and I hammered out the issue he found in Slack (p1661968485400859-slack-newspack-product); it turns out it was caused by an ad space, and was not the result of this PR so that should be safe to ignore! 

Changes like this do make me a little nervous so I appreciate the caution all around! 🙂 

I tested this PR with the following: 

* Tested different templates (default, one column, one column wide, blank), header settings and featured image settings
* Tested with some of our publisher's CSS that changed post layouts significantly (Berkeleyside, MTFP, OnlySky, Bethesda)
* Tested with different campaign placements

Overall, the only issue I ran into was a slight vertical scroll on short pages caused by either:

* The admin toolbar being visible; I added a code suggestion with a CSS fix, but this definitely isn't a blocker!
* An above-header campaign being visible -- unlike the admin toolbar, there's no way to know how tall a campaign will be, so no easy CSS tweak to work around. That being said, on pages with little content I think it's less weird than the footer appearing mid-window, so I think it's fine. 

I also did some digging and confirmed the direct decedents of `#page` in a site's markup would be:

* Skip Link (.skip-link)
* Above Header Widget (.above-header-widgets)
* Header (#masthead)
* Breadcrumbs (.site-breadcrumb)
* Below header ad (.global_below_header)
* Below Header widget (.below-header-widgets)
* Content (#content)
* Footer (#colophon)

Elements like the above header ad and all hamburger menus are outside of the #page, and the above footer and mobile sticky ad locations are inside of the #content element, so they shouldn't be affected by this change.

Based on that, I do feel like there's minimum risk of a site being negatively affected by this change, but it's hard to say for certain with all the different customizations/plugins/settings we have out there 😬 I'm making as approved based on testing, but if there's anything we can do to help catch issues when rolling this out -- maybe cloning sites to test the change ahead of time, or just being extra careful about reported issues and proactively spot-checking sites -- I think that could help!

adekbadek:
Thank you both for digging into this!

Seems like the issue is not necessarily siblings of `#content`, but siblings of `#content` which also have a `flex-grow` property. Example of which is [`.newspack_global_ad`](https://github.com/Automattic/newspack-ads/blob/dc783ca1153ff4d3899d2ceb79d2278c50fcef8d/src/frontend/style.scss#L19). 

This may be fixed by setting a super-high `flex-grow` on the `#content` element (10c6252c1f1d069f34cd7b0b8b2f128a33f22d80). WDYT?

laurelfulford:
The `99999` does the trick, though I'm not sure if that could have unintended consequences? (Part of that is just that I find `flex` values frustrating in how they're applied sometimes! I always think I have it down, until I don't 😅 ). 

What do you think about overriding the `flex-grow` value of everything that's a sibling of `#content` and changing it to `0`? A `flex-grow` value on anything that's a child of `#page` won't do anything until this change is applied -- so it shouldn't cause any issues, and that way we're not allocating any `flex-grow` to anything but the `#content`.

(As an aside, do you happen to know why `.newspack_global_ad` has flex properties? The element itself gets `display: flex` from the ads plugin, but unless its actually nested in something that is also `display: flex`, the `flex: 1 1 auto` doesn't do anything. If we don't want to do anything too general, we could also just change the `flex-grow` to `0` when the ad is a direct child of `#page`, but I'm also wondering if that's safe to remove outright?)

adekbadek:
> What do you think about overriding the flex-grow value of everything that's a sibling of #content and changing it to 0? 

Sounds interesting! Done ind4a7edf5b56c9170b8628eced95b8ee614f2a8c3

> As an aside, do you happen to know why .newspack_global_ad has flex properties? 

Nope, didn't have time to dig into that. You're right that it doesn't make much sense if the wrapper is no `flex`. @miguelpeixe – do you have an idea why `.newspack_global_ad` gets `flex: 1`?

laurelfulford (approved):
Thanks @adekbadek! I think I personally prefer this approach -- it's more exact (all the "extra space" is going to the `#content`, instead of a large fraction, which seems more predictable), and the `#page > *:not( #content )` selector should be strong enough to override most normal styles. 

(If something rolls in with a `flex-grow` value set with `!important`, that's another story, but in that case I'd also have a lot of questions about that element's styles, and why they are what they are 😅 ).

Retested with the ad spaces with and without AMP, and comparing against master the spacing looks the same.

Re-approving! 🚢

adekbadek:
@dkoo – could you have another 👀 ?

dkoo (approved):
This seems to be rendering as expected, now. 👍

kmwilkerson:
@adekbadek Will this cause any issues with the mobile sticky footer ad?

yogeshbeniwal:
@kmwilkerson No it doesn't create any issue so far.

adekbadek:
@kmwilkerson I've tested it locally and it does not.

PR #1924, opened 2022-09-07
Footer: add option to toggle off branding

### LINKED ISSUE #1185: Footer: Add option to toggle branding
Right now, the footer branding will appear when you have footer widgets; in some cases, publishers haven't wanted that layout, though, and it needs to be  hidden with Custom CSS.  

It would be handy to have the option to hide this branding in the Customizer.

laurelfulford:
### All Submissions:

The theme uses the following behaviour for the footer branding and social links menu:

* If you don't have any footer widgets assigned, the footer logo does not display. The social links menu, if present, will appear in line with the website copyright at the very bottom.
* If you do have footer widgets assigned, the site logo (or footer logo) will appear at the top of the footer with the social links menu next to it. 

This doesn't work well with different footer layouts, especially with widget blocks -- it's not uncommon to hide the footer branding section with CSS, and then re-add the logo, social links, or both as blocks so they can be laid out differently. 

This PR adds a toggle to hide the "footer branding"; it overrides the default behaviour, rather than replacing it. If I could go back in time, I would've taken this approach to start, but unfortunately, that's hard to do now without affecting live sites. So: compromise! 🙂 

Closes #1185

1. Apply the PR and run `npm run build.`
2. Navigate to Customizer > Footer Settings and confirm you now have an option to toggle off footer branding; it should be checked by default. 

image

3. On your test site, if you don't already have the following, add a site logo, a social links menu, and at least one widget in the Footer area.
4. Confirm that your site has a logo in the footer, and the social links next to it:

image

5. Uncheck "Show footer branding" and save the Customizer settings.
6. Confirm that the footer logo is gone, and that the social links menu now appears at the very bottom in the right corner:

image

7. Re-check "Show footer branding".
8. Remove the site logo and confirm that the footer branding section doesn't display, and that the social links are in the bottom right corner.
9. Re-add the logo, and remove the footer widgets; confirm that the logo does not display, and the social links appear in the bottom right corner. 
10. Remove the site logo, and add a logo to the footer logo space instead; re-add the footer widgets and confirm that the branding appears at the top of the footer with the social links, until you uncheck "Show footer branding" again; at that point it should disappear.

### Other information:

### DIFF
--- a/newspack-theme/footer.php
+++ b/newspack-theme/footer.php
@@ -60,7 +60,8 @@ if ( '' !== get_theme_mod( 'newspack_footer_logo', '' ) && 0 !== get_theme_mod( 
 					the_privacy_policy_link( '', '' );
 				}
 
-				if ( ( ! is_active_sidebar( 'footer-1' ) || ! has_custom_logo() ) || ! $show_footer_branding ) {
+
+				if ( ( ! is_active_sidebar( 'footer-1' ) || ! ( has_custom_logo() || $has_footer_logo ) ) || ! $show_footer_branding ) {
 					newspack_social_menu_footer();
 				}
 				?>
--- a/newspack-theme/footer.php
+++ b/newspack-theme/footer.php
@@ -9,6 +9,12 @@
  * @package Newspack
  */
 
+
+$show_footer_branding = get_theme_mod( 'footer_show_branding', true );
+$has_footer_logo = false;
+if ( '' !== get_theme_mod( 'newspack_footer_logo', '' ) && 0 !== get_theme_mod( 'newspack_footer_logo', '' ) ) {
+	$has_footer_logo = true;
+}
 ?>
 
 	<?php if ( is_active_sidebar( 'footer-3' ) ) : ?>
@@ -54,7 +60,7 @@
 					the_privacy_policy_link( '', '' );
 				}
 
-				if ( ! is_active_sidebar( 'footer-1' ) || ( ! has_custom_logo() ) ) {
+				if ( ! is_active_sidebar( 'footer-1' ) || ! $show_footer_branding || ! ( has_custom_logo() && $has_footer_logo ) ) {
 					newspack_social_menu_footer();
 				}
 				?>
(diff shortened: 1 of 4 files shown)


adekbadek (commented):
Two things I've found confusing:

1. When there are no footer widgets, and the footer logo is set, the footer logo is not displayed. Is that intentional?
2. If there are footer widgets, and the footer logo is set, the social link menu is rendered in two places. Is that intentional?

laurelfulford:
Thanks @adekbadek! 

> When there are no footer widgets, and the footer logo is set, the footer logo is not displayed. Is that intentional?

This is intentional, but it's also definitely confusing 😞 It's an artifact of how the theme was designed, with the idea that publications would only want very simple footer content (like just adding some text, a set of links, rather than wanting to control if/where a logo and social links would appear, which ended up being the case!). This has since been complicated by the introduction of widget blocks, which makes it much easier to create 

I'm not sure the best way to get rid of this behaviour without it causing unintended outcomes on live sites. Something like a third 'default' option could work (so you can have the branding always on, always off, or dependent on the footer widgets), but that still kind of leaves the out-of-the-box behaviour as weird. Definitely open to ideas on this! 

> If there are footer widgets, and the footer logo is set, the social link menu is rendered in two places. Is that intentional?

That is not! This should be fixed in [3a00367](https://github.com/Automattic/newspack-theme/pull/1924/commits/3a00367f87573a6f76b3bdd5cf5d7d68714a841f).

Just let me know if you have any other questions, or ideas! I was trying to move a common CSS tweak into the theme, but I realize it could cause a confusing behaviour to get even more confusing!

adekbadek:
Thanks for clarifying that! I agree that changing the footer & widgets behaviour might have unwanted consequences on the live sites.

adekbadek (approved):

laurelfulford:
Thanks for the review @adekbadek! Stuff like this reminds me to try not to get clever with any settings anymore 😅

JustinFreid:
How about an option to overwrite the default Newspack branding?

adekbadek:
If you're asking about the "Proudly powered by Newspack…" line, you can always hide it with custom CSS.

JustinFreid:
> If you're asking about the "Proudly powered by Newspack…" line, you can always hide it with custom CSS.

Yes, but it is preferable to have to footer text be editable entirely within the Theme Customizer instead. You can populate it by default with the existing text.

Issue #1128, opened 2020-10-28
Hide 'blank' avatars


laurelfulford:
Right now in the theme you can't hide the avatar only when it's empty.

It seems like the 'blank' avatar should not show at all, but instead, it will load a blank image the same size as is being passed to the avatar.

If possible, avatars that are not set -- at least when the 'blank' option is picked under WP Admin > Settings > Discussion -- should not display anything, and should not create a 'space' -- instead, they should look like avatars are turned off.

PR #1942, opened 2022-09-27
Print styles: correct sticky header, Safari page width

### LINKED ISSUE #1936: Print styles: sticky headers and two column templates
**Describe the bug**
A couple more issues have recently come up with regards to print styles:

1. The sticky header can appear further down the page if you print the page after scrolling.
2. The two column template prints the content wider than the one column template, even though the content is the same width online.

laurelfulford:
### All Submissions:

This PR further tweaks the print styles by: 
* "Unsticking" the sticky header when printing -- in some browsers, like Chrome, if you printed the page after scrolling a bit, the header would appear mid-page. 
* Swapping out the page margin approach -- `@page` isn't supported by Safari, so it was causing the right edge of the page to be cut off in some cases. 

Closes #1936

1. Set your header to sticky, and create a post and use the two column template.
2. In Chrome or Safari, scroll partway down the page, and then print/view the print preview -- the header will appear mid-page, instead of at the top (for some reason, I wasn't able to recreate this in Firefox):

image

3. In Safari, review the rest of the print preview -- note that the edge is getting cut off on the right side. This doesn't happen with the one column templates: 

image

5. Repeat steps 2 and 3, and confirm that these issues are now resolved. 

### Other information:

### DIFF
--- a/newspack-theme/sass/print.scss
+++ b/newspack-theme/sass/print.scss
@@ -18,9 +18,9 @@ Andreas Hecht in https://www.jotform.com/blog/css-perfect-print-stylesheet-98272
 @media print {
 	/* Margins */
 	#page {
-		max-width: 96%;
+		max-width: 94%;
 		margin: auto;
-		width: 96%;
+		width: 94%;
 	}
 
 	#primary,
--- a/newspack-theme/sass/print.scss
+++ b/newspack-theme/sass/print.scss
@@ -150,6 +150,7 @@ Andreas Hecht in https://www.jotform.com/blog/css-perfect-print-stylesheet-98272
 	.highlight-menu-contain,
 	.mobile-menu-toggle,
 	.desktop-menu-toggle,
+	.newspack-reader__account-link__mobile,
 	.jp-relatedposts-i2,
 	.social-navigation,
 	.subpage-toggle-contain,
--- a/newspack-theme/sass/print.scss
+++ b/newspack-theme/sass/print.scss
@@ -17,9 +17,14 @@ Andreas Hecht in https://www.jotform.com/blog/css-perfect-print-stylesheet-98272
 
 @media print {
 	/* Margins */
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
I can submit a new issue but perhaps this can be done in this PR, since it's related to print styles:

1. RAS' "My Account" link should also be hidden:

image

2. If there's no author image, there should be no empty space (there's none in non-print styles):

image

3. Not sure if we can do anything about it, but Chrome adds these page headers and footer, which may obscure content:

image

laurelfulford:
Thanks @adekbadek!

> RAS' "My Account" link should also be hidden

Oh, good catch! Pushed up a fix for this in [b0bd142](https://github.com/Automattic/newspack-theme/pull/1942/commits/b0bd142e78aea22d3b0633a1b8906239b19e2f44).

> If there's no author image, there should be no empty space (there's none in non-print styles)

Just to confirm for this one, is the setup that the author doesn't have an avatar set, and the Gravatar fallback is set to blank? My experience is [in that case is that there's a gap in the non-print styles as well](https://github.com/Automattic/newspack-theme/issues/1128), but I might be misunderstanding! 

> Not sure if we can do anything about it, but Chrome adds these page headers and footer, which may obscure content:

I could recreate this with and without the headers and footers -- it looks like it's related to the flex styles on `#page`, weirdly enough! I overrode those for the print styles in [a3810b2](https://github.com/Automattic/newspack-theme/pull/1942/commits/a3810b23410ed8e2b9910b6b315f2a717cdffe67), but please let me know if it's still happening for you (I tried different lengths of content, but I'm a bit paranoid that it just _seems_ fixed).

adekbadek (approved):
All three issues are fixed, thanks!

laurelfulford:
Awesome -- thanks @adekbadek! 🙌

PR #1959, opened 2022-10-23
fix: Correct search block styles

### LINKED ISSUE #1947: Search Block: Display errors in WP 6.1 beta 3
**Describe the bug**
The search block's appearance is changed in WP 6.1, and it has the following display issues in the Newspack theme:

image

It has:
* An extra border around the search form
* When the button is placed 'inside', it's white text on a white background so it's not visible.

A block with the same settings in WP 6.0 looks like this: 

image

This specifically is the search block with the button placed inside and using the magnifying glass icon, but all the other settings (button outside; text instead of icon), should be tested.

laurelfulford:
### All Submissions:

This PR fixes some display issues with the search block, both for the upcoming 6.1 release, and issues that exist with 6.0.3.

Closes #1947

1. Add multiple copies of the search block to the editor, using the different button placements and colours (alternatively, [you can copy-paste this code into the editor](https://gist.github.com/laurelfulford/771e5f6942feace9911b3fa453b26e66) to get all of the combinations).
2. View in the editor and on the front-end and note the issues (you can also see screenshots in [this issue](https://github.com/Automattic/newspack-theme/issues/1947#issuecomment-1288164066)).
* When the button is placed inside, there's an extra border around the form.
* The buttons aren't using the correct colours - by default they should use your primary colour like other buttons, but they should also let you change the button colour, and it's not always being applied.
4. In the editor and on the front-end, confirm that:
* The button, by default, uses your secondary colour. The only exception is when the button is placed inside the search and uses the icon -- in that case, the button should have a transparent background with a black icon. (This is so the block looks like the search field in the theme by default). 
* When the button is inside of the search, you don't have an extra border anymore.

image
image

5. Upgrade to the 6.1 RC and confirm things still look correct (currently the search doesn't look 1:1 between the current and future release -- you can see screenshots of the pre-fixed styles [here](https://github.com/Automattic/newspack-theme/issues/1947#issuecomment-1288164583)).

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -26,7 +26,6 @@ function newspack_custom_colors_css() {
 			*[class^="wp-block-"].is-style-solid-color,
 			*[class^="wp-block-"].is-style-solid-color.has-primary-background-color,
 			.is-style-outline .wp-block-button__link.has-primary-background-color:not( :hover ),
-			.wp-block-search__button-outside .wp-block-search__button,
 			.wp-block-file .wp-block-file__button,
 			.comment .comment-author .post-author-badge,
 			.woocommerce .onsale,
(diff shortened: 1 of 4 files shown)


adekbadek (approved):
👍

laurelfulford:
🎉

PR #1975, opened 2022-11-10
fix: update newspack-scripts

adekbadek:
Update `newspack-scripts`.


dkoo:
@adekbadek is this still needed? Looks like the only changes are to package-lock.js.

adekbadek:
Not needed, the [dependabot beat me to it](https://github.com/Automattic/newspack-theme/commit/e8f996c85360caa289ba11327f7965a9d6c5c68f).

PR #1983, opened 2022-11-19
Update primary domain

gamebits:
Updates `newspack.pub` references to `newspack.com`. Should not be merged and deployed until Newspack's primary domain is updated. See Slack: p1668808309416999-slack-G012YAG1WSU


laurelfulford (approved):
This looks good to me, and I confirmed the translations are loading the correct new URL.

Marking as approved for when it's safe to merge later on!

laurelfulford:
Also I just noticed that the style.css files still point to newspack._blog_ -- it's not visible on the front-end like this link is, and it redirects to the .com address, so I think it can be updated in a separate PR 🙂

laurelfulford:
Merging since .com is live now, and .pub and .blog are redirecting to it 🎉

PR #569, opened 2019-11-11
Fix issue with guest author image fallback

laurelfulford:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

This PR fixes an issue where, if you use Co-Authors Plus, set up a Guest Author, but don't set an Avatar, it uses a post's featured image as the fallback.

I set it up so that it will fall back to use whatever the WordPress site is using for the Gravatar default, but it can also be set up to output nothing. 

Closes #567.

### How to test the changes in this Pull Request:

1. Add a Guest Author using the Co-Author Plus plugin; make sure to give them a short bio (so they show up at the bottom of single posts), but no avatar.
2. Add this author to a post on their own; make sure this post has a featured image.
3. Add this author to a post with another Guest author and a regular WordPress user for comparison. Make sure this post has a featured image.
4. View the byline on the single post in each case; note that the Guest Author without an image uses the featured image:

image

image

5. View the author bio in each case:

image

image

6. View the guest author's author archive:

image

7. Apply the PR.

8. View the byline on single posts again:

image

image

9. View the author bio on single posts again:

image

image

10. View the guest author archive again:

image

11. View an archive for a regular WP user and a Guest Author with an avatar, and confirm they're both still displaying as expected.

12. Deactivate the Co-Author Plus plugin and confirm the authors are still displaying correctly.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #2008, opened 2022-12-21
fix: correct issues with users and guest authors sharing IDs

laurelfulford:
### All Submissions:

This PR fixes a couple issues that arise when a regular user and a CoAuthors Plus guest user share the same ID:
* The 'regular user' ends up with the incorrect avatar ([caused by an old fix to make sure CoAuthors didn't pick up post featured images as a fallback](https://github.com/Automattic/newspack-theme/pull/569))
* The guest user ends up with the wrong name/description on the Author Archives page.

This PR also fixes an issue where guest author information was not displaying on the archive pages. 

Closes 1200133283606042/1203581044315819

1. Start with a test site where at least one WP User shares an ID with a CoAuthors Plus Guest Author. If you don't already have a site with this, you can recreate it by:
   * Spin up a new Newspack site (opting NOT to install the demo content)
   * Add two Guest Authors, both with names, bios, but no avatars for now. To make testing easier, prefix the bios with the user name.  
   * Checking the ID of the first Guest Author; and add enough WP Users to get up to that ID number -- for me, it was ID 13. Make sure at least this 'ID twin' WP User has a bio that starts with their user name.
2.  Set up a few posts:
    * Posts just by your "ID twin" Guest Author
    * Posts just by your "ID twin" WP User
    * Posts by both together
    * Posts by your "ID twin" Guest Author and another random user
    * Posts by the Guest Author that doesn't have an ID twin. 
3. Make sure your "ID Twin" WP User has an avatar (either through Gravatar or Simple Avatars), but your "ID Twin" Guest Author does not. 
4. Visit the post by the "ID twin" WP User; note that rather than using its avatar, it's using the 'fallback' option -- this is because it's picking up the Guest Author's avatar:

_WP User's settings:_ 
image

_WP User's byline:_
image
  
6. Visit the archive of the "ID twin" guest author; note that it's using the WP User's name and description.

_Note the URL points to guest author 'Jane Doe' but the name/description does not -- the posts should match the right person though:_
image

7. Visit the regular Guest Author that doesn't have an "ID twin"; notice that its not displaying any information:

image

9. Apply the PR.
10. Visit the post by the "ID twin" WP User; confirm that its using the correct avatar.

image

11. Visit the archive of the "ID twin" guest author, and confirm that it's displaying the correct name and description.

image

12. Visit the regular Guest Author that doesn't have an "ID twin" and confirm their information is displaying:

image

13. Test adding/removing avatars from the Guest Authors (both with ID twins and without), and confirm that they correctly fall back to the placeholder image set under Settings > Discussion, and not featured images. 

14. Swap to Newspack Sacha, and comfirm that the Author Bios at the bottom of posts are using the correct images as well -- this is the only child theme that has a unique template part for this section.

(Note: the avatar issue still needs to be addressed separately in the blocks). 

### Other information:

### DIFF
--- a/newspack-theme/archive.php
+++ b/newspack-theme/archive.php
@@ -49,28 +49,22 @@ $queried             = get_queried_object();
 				}
 				?>
 
+				<?php the_archive_title( '<h1 class="page-title">', '</h1>' ); ?>
+
+				<?php do_action( 'newspack_theme_below_archive_title' ); ?>
+
 				<?php
-				if ( is_author() && 'guest-author' === get_post_type( $queried->ID ) ) {
-					printf(
-						'<h1 class="page-title"><span class="page-subtitle">%1$s</span> <span class="page-description">%2$s</span></h1>',
-						esc_html__( 'Author Archives:', 'newspack' ),
-						esc_html( $queried->display_name )
-					);
-				} else {
-					the_archive_title( '<h1 class="page-title">', '</h1>' );
-				}
-
-				do_action( 'newspack_theme_below_archive_title' );
-
 				if ( ( is_category() || is_tag() ) && ! empty( $native_sponsors ) ) :
 					// Get description for native archive sponsors.
 					newspack_sponsor_archive_description( $native_sponsors );
-				elseif ( is_author() && 'guest-author' === get_post_type( $queried->ID ) && '' !== $queried->description ) :
-					echo '<div class="taxonomy-description">' . wp_kses_post( wpautop( $queried->description ) ) . '</div>';
 				elseif ( '' !== get_the_archive_description() ) :
-					echo '<div class="taxonomy-description">' . wp_kses_post( wpautop( get_the_archive_description() ) ) . '</div>';
-				endif;
+					?>
+					<div class="taxonomy-description">
+						<?php echo wp_kses_post( wpautop( get_the_archive_description() ) ); ?>
+					</div>
+				<?php endif; ?>
 
+				<?php
 				if ( ( is_category() || is_tag() ) && ! empty( $underwriter_sponsors ) ) {
 					// Get info for underwriter archive sponsors.
 					newspack_sponsored_underwriters_info( $underwriter_sponsors );
--- a/newspack-theme/archive.php
+++ b/newspack-theme/archive.php
@@ -17,8 +17,6 @@ if ( function_exists( 'newspack_get_all_sponsors' ) ) {
 
 $feature_latest_post = get_theme_mod( 'archive_feature_latest_post', true );
 $show_excerpt        = get_theme_mod( 'archive_show_excerpt', false );
-$queried             = get_queried_object();
-
 ?>
 
 	<section id="primary" class="content-area">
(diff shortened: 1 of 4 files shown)


newspack-theme/archive.php
```diff
@@ -61,22 +49,28 @@
 				}
 				?>
 
-				<?php the_archive_title( '<h1 class="page-title">', '</h1>' ); ?>
+				<?php
+				if ( is_author() && 'guest-author' === get_post_type( $queried->ID ) ) {
```
leogermani, line 53: This check is ambiguous, it will return `true` both for the real WP user archive page and for the Guest Author archive page.

Is that the intention? Maybe we could be more explicit here, like 

```
if ( is_author() && $queried instanceof WP_User ) {
```
if we want to check if it's a real user archive.

If we want to check if it's a guest author archive:

```
if ( is_author() && $queried instanceof StdClass && isset($queried->type) && 'guest-author' ===  $queried->type) {
```
(there might be an easier way for the latter, but that's the idea)

newspack-theme/archive.php
```diff
@@ -61,22 +49,28 @@
...
+				}
 
-				<?php do_action( 'newspack_theme_below_archive_title' ); ?>
+				do_action( 'newspack_theme_below_archive_title' );
 
-				<?php
 				if ( ( is_category() || is_tag() ) && ! empty( $native_sponsors ) ) :
 					// Get description for native archive sponsors.
 					newspack_sponsor_archive_description( $native_sponsors );
+				elseif ( is_author() && 'guest-author' === get_post_type( $queried->ID ) && '' !== $queried->description ) :
```
leogermani, line 68: same as above

leogermani (changes requested):
I ran all the tests and everything you describe works well.

If we dig, there are still some bugs on how CAP handles this, but it's not our fault and I dont see it interfering with the site front end... 

we can dig more later, but I think the changes look good and the code is easier to read.

I left one comment with a test that looks ambiguous to me. I don't know if you want that check to pass in both cases or if you are trying to filter that condition only to guest authors. (if you are, it's not working)

newspack-theme/archive.php
laurelfulford: 🤦‍♀️ Thanks @leogermani! 

This was to "fix" the odd issue I saw with the archive title and description being incorrect or not displaying at all, but I can't recreate it now - that's probably why it seemed to work for me! 

It did seem very strange to me that the Guest Author information hadn't been working all this time and wasn't reported -- in reality I think it was an odd blip that happened on my test site and went away on its own.

I'm going to remove this and the description code you flagged below, and do a little more testing myself, then update this PR 🙂

laurelfulford: Ok, I've updated the PR to remove the above code for now! 

I was able to recreate the Author Title and Description being wrong by removing all the menus from the site. If this is somehow related to the theme, it _seems_ unrelated to this particular problem, so I'm going to investigate it separately!

newspack-theme/archive.php
laurelfulford: With the above, should be resolved in [2afbbba](https://github.com/Automattic/newspack-theme/pull/2008/commits/2afbbba5ea88bacaf79c4f0fdede01a9274656b3)!

leogermani (approved):
Tests well in all scenarios for me

laurelfulford:
Thanks @leogermani! 🙌

PR #2012, opened 2023-01-05
fix: skip flex rule for newspack ad

miguelpeixe:
### All Submissions:

#1922 introduced a bug that breaks the fixed height option for the Below Header placement. This PR ensures to not apply the `flex: 0;` rule for `.newspack_global_ad`.

1. While on master, configure an ad unit and enable fixed height for "Below Header"
2. Visit the frontend and confirm you get CLS
3. Switch to this branch, build, refresh, and confirm the fixed height is applied and there's no CLS

### Other information:

### DIFF
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -7,7 +7,7 @@
 	display: flex;
 	flex-direction: column;
 	min-height: 100vh;
-	> *:not( #content ) {
+	> *:not( #content ):not( .newspack_global_ad ) {
 		flex: 0;
 	}
 }
--- a/newspack-theme/sass/layout/_layout.scss
+++ b/newspack-theme/sass/layout/_layout.scss
@@ -7,8 +7,8 @@
 	display: flex;
 	flex-direction: column;
 	min-height: 100vh;
-	> *:not( #content ):not( .newspack_global_ad ) {
-		flex: 0;
+	> *:not( #content ) {
+		flex-grow: 0;
 	}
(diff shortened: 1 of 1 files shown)


laurelfulford (changes requested):
This works well overall, except the weird edge case we were seeing before: if the page content is super short, the ad ends up splitting the space between it and the content, even if the ad is shorter than that 😕 

It's definitely an extreme example, but here's a screenshot of what I mean: AMP is disabled in this screenshot (since that's the only case I could recreate the original issue with `master`), and the ad should only max out at 250px high: 

image

(It looks like this works fine with AMP because the height is set inside of the `newspack_global_ad` container instead of on it -- TIL!). 

Messing around with it, I think updating the styles to this should do the trick without affecting anything else: 

```
> *:not( #content ) {
     flex-grow: 0;
}
```

As I understand it, switching it away from the `flex` shorthand lets us update the `flex-grow` property without affecting the `flex-shrink` or `flex-basis` attributes -- I think the flex-basis of `0` is what's causing the container to collapse.

Let me know what you think, and if you have any questions, especially if this doesn't make sense/work as advertised -- I'm really good at flexbox except when I'm not! 😆

miguelpeixe:
Ah, I see! Thanks for clarifying, @laurelfulford!

Removing the shorthand property and using `flex-grow` does fix the issue for fixed-height ads while also preserving the intended flex for short pages.

laurelfulford (approved):
Thanks @miguelpeixe! This looks good to me! :shipit:

Issue #1991, opened 2022-12-02
Buttons do not have an accessible name


deLisle:
**Describe the bug**
PageSpeed and Lighthouse complain on the desktop version on menus that have sub items. It's the arrow button that doesn't have an accessible name.

**To Reproduce**
If you run pagespeed on a site that has a menu with sub items you'll see the issue under Accessibility for the Desktop version.

**Screenshots**
<img width="931" alt="Screenshot 2022-12-02 at 09 32 06" src="https://user-images.githubusercontent.com/195763/205250960-1e9dba57-6f52-4f7b-9085-0b53f2a1b763.png">

**Expected behavior**
For it to work properly read more here: https://dequeuniversity.com/rules/axe/4.4/button-name?utm_source=lighthouse&utm_medium=lr

laurelfulford:
This should be resolved in #2025!

PR #2025, opened 2023-01-21
Update menu ARIA labels, add missing button text

### LINKED ISSUE #1963: ARIA IDs are not unique
Pagespeed and Lighthouse reports have recently changed. 

In the latest report we get, we see this issue as a problem for the SEO of my URLs. 

According to this report: The value of an ARIA ID must be unique to prevent other instances from being overlooked by assistive technologies.

This issue is related to the submenu items on the menu. 

`<html>`
`<body>`
`<!--StartFragment-->`

Failing Elements
--
nav.main-navigation > ul#menu-principal > li#menu-item-408 > ul#submenu-408
`<ul class="sub-menu" id="submenu-408">`
CineSeriesDisney+Star+HBO MaxNetflixParamount+
`<ul class="sub-menu" id="submenu-408">`

`<!--EndFragment-->`
`</body>`
`</html>`

Could you please try to solve this? 

Thanks for your help.

laurelfulford:
### All Submissions:

This PR reworks the approach I used in #1808 to add ARIA labels to the theme's dropdown menus:

I had originally used PHP and a menu walker to add the parent menu ID to each submenu to associate them with the dropdown toggle. 

This worked well with AMP because it doesn't require JavaScript. However, it did not work with how AMP handles mobile menus: the theme uses AMP's [Toolbar functionality](https://amp.dev/documentation/components/amp-sidebar#toolbar), which lets you set up the menus in the mobile `amp-sidebar`, and then use the `toolbar-target` attribute to copy the menus to the header at a certain breakpoint. The copied menus resulted in duplicate IDs that started turning up in Lighthouse reports. 

To fix, I've reworked the approach to use JavaScript. This makes it way easier to avoid the duplication, but it also means sites running AMP will need to allow two JavaScript files to have the full range of ARIA attributes in their menus.

**Note:** Due to using AMP Toolbar for the menus, the menus aren't in the header immediately on pageload when AMP is enabled. 

I ended up using `MutationObserver` to make sure the two menus with dropdowns -- the primary and secondary menus -- were actually moved before firing the JavaScript when AMP is enabled. 

Another approach I considered was getting rid of the AMP Toolbar code all together, and just duplicating the menus in header and mobile sidebar code. I went with the current approach because it changed fewer files, but I'm happy to revisit removing the Toolbar code if that seems like a better approach!

This PR also adds the text labels to the dropdown button toggles. 

Closes #1963 and #1991.

<details>
<summary><strong>1. Confirm Lighthouse errors are gone</strong></summary>

1. Set up a site with two menus - one in the primary menu location and one in the secondary menu location. 
2. Make sure both menus have at least one dropdown that's more than one level deep. 
3. Run a Lighthouse test on the desktop version of your site -- you should get reports for the following issues under the Accessibility section:

Duplicate ARIA IDs:
image

Buttons without accessible names:
image

5. Rerun the Lighthouse test -- both errors should now be gone. But now we have to make sure everything else still works! 
</details>

<details>
<summary><strong>2. Re-test original functionality - without new JavaScript files</strong></summary>

1. Start by testing the AMP version of your page; first we're going to test how things look without the JavaScript -- it will be missing one attribute-ID pair, but otherwise should function the same. 
2. Open the element inspector and view the code for one of your dropdowns; confirm that the button toggle (the arrow) now has the text label "Open dropdown menu", wrapped in a span with the CSS class `screen-reader-text`:

image

3. With the element inspector still open, use the keyboard to tab through the page until you're focused on the same dropdown menu toggle:

image

4. Click 'Return' to open the menu.
5. Confirm that the following happens:
    * That the dropdown menu opens.
    * That the button's screen reader text label changed to "Close dropdown menu"
    * That the `aria-expanded` attribute on the button changed from `false` to `true`. 
6. Confirm that clicking 'Return' again while focused on the dropdown toggle resets things -- the menu closes, and the button's text label and `aria-expanded` attribute return to "Open dropdown menu" and `false`, respectively.
7. Click 'Return' again to reopen the dropdown menu; click Tab a few more times to to tab through the dropdown menu and repeat steps 5 and 6 on any lower levels of dropdown menu, confirming that the changes only happen on the level of menu you're operating. 
8. Repeat the above in the other menu -- so if you started with the primary menu, make sure to test the secondary menu.

</details>

<details>
<summary><strong>3. Re-testing original functionality - with new JavaScript files</strong></summary>

1. In the WP Toolbar, hover over the AMP menu item, and click 'Validate URL' from the dropdown:

image

2. On the next screen, look for two JavaScript files coming from the theme -- one will be called menu-accessibility.js, and the other is inline but probably right before it -- if you click the arrow toggle right after teh 'Invalid inline script' label, you should be able to see the script's ID, `newspack-menu-accessibility-js-extra`:

image 

3. Mark both of these scripts as "Kept", and click "Update".
4. The JavaScript adds IDs to the submenus, so this is a good time to try re-running the Lighthouse Accessibility test again, to make sure you're not getting any duplicate errors. 
5. Inspect one of your dropdowns again. The button toggle should now have an `aria-controls` attribute with a value like `submenu-###`. The submenu `ul` right after it should have a matching ID. 
6. Confirm that steps 5 and 6 from the previous section still work as described (toggling open the menu with the keyboard works; the `aria-expanded` value and button text update).
7. Unassign the primary menu; confirm that the steps 4-5 from this section, and 5-6 from the previous section still work on the remaining secondary menu.
8. Reassign the primary menu.
9. Disable AMP. 
10. Repeat 5-6 from the previous section, and steps 4-5 and 7 from this section to make everything works the same without AMP (this is especially important because things like the `aria-expanded` and button text use additional JavaScript when AMP is off).
</details>

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -424,23 +424,28 @@ function newspack_scripts() {
 
 	wp_enqueue_style( 'newspack-print-style', get_template_directory_uri() . '/styles/print.css', array(), wp_get_theme()->get( 'Version' ), 'print' );
 
+	$newspack_l10n = array(
+		'open_search'         => esc_html__( 'Open Search', 'newspack' ),
+		'close_search'        => esc_html__( 'Close Search', 'newspack' ),
+		'expand_comments'     => esc_html__( 'Expand Comments', 'newspack' ),
+		'collapse_comments'   => esc_html__( 'Collapse Comments', 'newspack' ),
+		'show_order_details'  => esc_html__( 'Show details', 'newspack' ),
+		'hide_order_details'  => esc_html__( 'Hide details', 'newspack' ),
+		'open_dropdown_menu'  => esc_html__( 'Open dropdown menu', 'newspack' ),
+		'close_dropdown_menu' => esc_html__( 'Close dropdown menu', 'newspack' ),
+	);
+
 	if ( ! newspack_is_amp() ) {
 		if ( is_singular() && comments_open() && get_option( 'thread_comments' ) ) {
 			wp_enqueue_script( 'comment-reply' );
 		}
 
-		$newspack_l10n = array(
-			'open_search'        => esc_html__( 'Open Search', 'newspack' ),
-			'close_search'       => esc_html__( 'Close Search', 'newspack' ),
-			'expand_comments'    => esc_html__( 'Expand Comments', 'newspack' ),
-			'collapse_comments'  => esc_html__( 'Collapse Comments', 'newspack' ),
-			'show_order_details' => esc_html__( 'Show details', 'newspack' ),
-			'hide_order_details' => esc_html__( 'Hide details', 'newspack' ),
-		);
-
 		wp_enqueue_script( 'newspack-amp-fallback', get_theme_file_uri( '/js/dist/amp-fallback.js' ), array(), wp_get_theme()->get( 'Version' ), true );
 		wp_localize_script( 'newspack-amp-fallback', 'newspackScreenReaderText', $newspack_l10n );
 	}
+
+	wp_enqueue_script( 'newspack-menu-accessibility', get_theme_file_uri( '/js/dist/menu-accessibility.js' ), array(), wp_get_theme()->get( 'Version' ), true );
+	wp_localize_script( 'newspack-menu-accessibility', 'newspackScreenReaderText', $newspack_l10n );
 
 	if ( class_exists( '\Newspack\AMP_Enhancements' ) && \Newspack\AMP_Enhancements::is_amp_plus_configured() ) {
 		wp_enqueue_script( 'newspack-amp-plus', get_theme_file_uri( '/js/dist/amp-plus.js' ), array(), wp_get_theme()->get( 'Version' ), true );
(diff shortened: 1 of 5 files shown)


leogermani:
Thanks for such a detailed step by step guide! I was able to follow it and test everything!

The only thing that didn't work for me was this:

> Inspect one of your dropdowns again. The button toggle should now have an aria-controls attribute with a value like submenu-###. The submenu ul right after it should have a matching ID.

The property is not being added.

There's an error being fired:

```
menu-accessibility.js:24 Uncaught TypeError: Cannot read properties of null (reading 'replace')
    at updateMenu (menu-accessibility.js:24:36)
    at eval (menu-accessibility.js:62:5)
    at eval (menu-accessibility.js:64:3)
    at ./newspack-theme/js/src/menu-accessibility.js (menu-accessibility.js?ver=1.66.0:18:1)
    at menu-accessibility.js?ver=1.66.0:29:79
    at menu-accessibility.js?ver=1.66.0:34:12
```

laurelfulford:
> The only thing that didn't work for me was this:

🤦‍♀️ Thanks @leogermani! I see this, too. I didn't catch that the ID was only getting added to the parent menu item when AMP was enabled -- after all this time working with AMP, this is the first time I've noticed that! 

I've updated the code so it's using a data attribute instead, and also to make sure a menu's there before the code tries to watch it.

laurelfulford:
Thanks @leogermani!

PR #1053, opened 2020-09-02
fix: harmonise how the author link is printed regardless of truncation

philipjohn:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-theme/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Fixes a bug where the "more by [author]" link was missing the actual username when the option to truncate author bios was not on.

### How to test the changes in this Pull Request:

1. In Customiser, under "Author Bio Settings" make sure that "Display Author Bio" is checked and "Truncate Author Bio" is unchecked
2. Visit any single post authored by a user with a filled in bio
3. Observe that there is a "More by [author]" link, where [author] is the user's name
4. Observe that the link looks like this: https://newspack.test/author/ (missing the username from the end)
5. Apply this PR and observe that the link now points correctly to the author archive page (e.g. https://newspack.test/author/admin)
6. Turn the "Truncate Author Bio" setting on in customiser and check that the author link is still correct

PR #2041, opened 2023-02-18
fix: Sacha author bio link

gamebits:
Updates Sacha to use `get_the_author_meta` for the "Read more" link in the author bio; otherwise, the nicename isn't obtained and the link points to `/author/` with no slug.

### All Submissions:

Brings Sacha's `author-bio.php` template part in line with the Newspack parent theme. See https://github.com/Automattic/newspack-theme/pull/1053

1. Activate the Newspack Sacha theme.
2. Publish a post and assign it to any author (say, `newspack`) who has a bio.
3. Ensure the Customizer's "Display Author Bio" option is enabled and "Truncate Author Bio" is disabled.
4. Scroll to the bottom of a single post and click the author's "More by" link. It will point to `https://example.com/author/`
5. Switch to this branch and test again. The link will now go to `https://example.com/author/newspack/`

### Other information:

* [X] Have you written new tests for your changes, as applicable?

### DIFF
--- a/newspack-sacha/template-parts/post/author-bio.php
+++ b/newspack-sacha/template-parts/post/author-bio.php
@@ -119,7 +119,7 @@ elseif ( (bool) get_the_author_meta( 'description' ) && is_single() ) :
 		<?php else : ?>
 			<?php echo wp_kses_post( wpautop( get_the_author_meta( 'description' ) ) ); ?>
 
-			<a class="author-link" href="<?php echo esc_url( get_author_posts_url( $author->ID, $author->user_nicename ) ); ?>" rel="author">
+			<a class="author-link" href="<?php echo esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ); ?>" rel="author">
 				<?php
 					/* translators: %s is the current author's name. */
 					printf( esc_html__( 'More by %s', 'newspack' ), esc_html( get_the_author() ) );


dkoo (changes requested):
I'm not able to replicate the bug you're reporting (incorrect author URLs on `master`). But the change doesn't seem to result in any negative consequences, either.

There are a couple of additional instances of `get_author_posts_url( $author->ID, $author->user_nicename ) )` in this file on lines 51 and 61. Do these also need to be updated, or are they working as intended? cc @laurelfulford

newspack-sacha/template-parts/post/author-bio.php
```diff
@@ -119,7 +119,7 @@
 		<?php else : ?>
 			<?php echo wp_kses_post( wpautop( get_the_author_meta( 'description' ) ) ); ?>
 
-			<a class="author-link" href="<?php echo esc_url( get_author_posts_url( $author->ID, $author->user_nicename ) ); ?>" rel="author">
+			<a class="author-link" href="<?php echo esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ); ?>" rel="author">
```
leogermani, line 122: We need to pass the user nicename as the second parameter here. This is used to disambiguate the author when a Guest Author has the same ID of a real user. We can't remove that.

laurelfulford: I think this should actually be okay -- it's inside of an if/else statement checking for Co-Authors Plus, and this specific bit is only run if Co-Authors Plus is not enabled (that's also the specific case where this happens, the link works okay when that plugin is activated). 

It's repetitive, but each case (with and without Co-Authors Plus) has a case where the bio is truncated, and a case where it is not -- in Newspack Sacha, [the case where the bio is truncated matches this](https://github.com/Automattic/newspack-theme/blob/master/newspack-sacha/template-parts/post/author-bio.php#L112), and the link works.

laurelfulford (approved):
Testing this without Co-Authors Plus activated, I'm able to recreate the issue and confirm that this does fix it -- I added more detail in the code comment, but basically this line is only run when Co-Authors Plus is deactivated, so the nice name shouldn't be needed in this specific case. 

I'm marking this as approved, but tagging @leogermani & @dkoo just in case this still seems incorrect to you!

newspack-sacha/template-parts/post/author-bio.php
leogermani: ok! Thanks for diving into it!

leogermani (approved):

PR #2059, opened 2023-03-14
fix: css :not() rules with multi-selectors broken in AMP

dkoo:
### All Submissions:

Fixes several instances of a CSS selector that AMP doesn't like:

```css
.selector:not( .sub-selector-1, .sub-selector-2 )
```

should be:

```css
.selector:not( .sub-selector-1 ):not( .sub-selector-2 )
```

These selectors tend to target highly specific groups of blocks that mainly appear on homepages.

1. On a fresh test site, enable AMP and install the [bad theme version](https://github.com/Automattic/newspack-theme/releases/tag/v1.69.1).
2. View the homepage and observe that it's broken, mainly because all the CSS below the first instance of the `:not` combo-selector is not compiled by AMP.

Screen Shot 2023-03-14 at 4 38 05 PM

3. Check out this branch, rebuild CSS, confirm that the homepage renders as expected.

### Other information:

### DIFF
--- a/newspack-theme/sass/plugins/newspack-ads.scss
+++ b/newspack-theme/sass/plugins/newspack-ads.scss
@@ -125,6 +125,18 @@ body {
 			z-index: 1;
 		}
 	}
+
+	&.single-featured-image-behind:not( .newspack-front-page ) .global_below_header + .site-content {
+		margin-top: 0;
+	}
+
+	@include utilities.media( tablet ) {
+		&.single-featured-image-beside:not( .newspack-front-page )
+			.global_below_header
+			+ .site-content {
+			margin-top: 0;
+		}
+	}
 }
 
 .single-featured-image-behind .newspack_global_ad.global_below_header {
--- a/newspack-theme/sass/plugins/newspack-ads.scss
+++ b/newspack-theme/sass/plugins/newspack-ads.scss
@@ -125,20 +125,6 @@ body {
 			z-index: 1;
 		}
 	}
-
-	&.single-featured-image-behind:not(.newspack-front-page, .style-pack-style-2)
-		.global_below_header
-		+ .site-content {
-		margin-top: 0;
-	}
-
-	@include utilities.media( tablet ) {
-		&.single-featured-image-beside:not(.newspack-front-page, .style-pack-style-2)
-			.global_below_header
-			+ .site-content {
-			margin-top: 0;
-		}
-	}
 }
 
 .single-featured-image-behind .newspack_global_ad.global_below_header {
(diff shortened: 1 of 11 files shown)


newspack-theme/sass/plugins/newspack-ads.scss
```diff
@@ -125,20 +125,6 @@ body {
 			z-index: 1;
 		}
 	}
-
-	&.single-featured-image-behind:not(.newspack-front-page, .style-pack-style-2)
```
laurelfulford, line 129: Sorry, this was the result of my bad explanation! It's specifically the `.style-pack-style-2` CSS class that doesn't exist anymore -- we still need these styles for two really specific use cases (it makes sure these ad styles aren't applied when the simplified subpage header is paired with either the featured image behind or featured image beside on posts and pages). 

"Style pack 2" eventually became Newspack Nelson but it doesn't look like the styles are still needed for that (or at least I can't see any issues comparing Newspack Nelson with these settings to other themes on master -- I'm betting it was a specificity thing!).

newspack-theme/sass/plugins/newspack-ads.scss
```diff
@@ -125,20 +125,6 @@ body {
...
-		.global_below_header
-		+ .site-content {
-		margin-top: 0;
-	}
-
-	@include utilities.media( tablet ) {
-		&.single-featured-image-beside:not(.newspack-front-page, .style-pack-style-2)
```
laurelfulford, line 136: See above! Only the `.style-pack-style-2` bit isn't needed, the rest of these styles look like they are still needed!

laurelfulford (changes requested):
I found one small error that was the result of my poor explanation around styles that weren't needed --- I added code comment but let me know if I'm not making sense!

Otherwise, this looks good to me! 🙂

newspack-theme/sass/plugins/newspack-ads.scss
dkoo: 397fd68 restores these styles without the `.style-pack-style-2` selector.

newspack-theme/sass/plugins/newspack-ads.scss
dkoo: 397fd68 restores these styles without the `.style-pack-style-2` selector.

laurelfulford (approved):
🙌 Thanks @dkoo! This looks good to me!

dkoo:
As expected, Stylelint doesn't like the AMP way of doing these selectors. We should update the config in `newspack-scripts` to address.

PR #2061, opened 2023-03-15
fix: display underwriters with default placement value

dkoo:
### All Submissions:

Fixes an issue with default sponsor meta values. The default values for several meta fields is `inherit`, but the theme expects them to be something else, which causes some issues with display when defaults are used. This mainly affects the display of underwriter-type sponsors when no post-level overrides are set.

Closes `1200550061930446/1204158091646412`

1. On `master`, publish a new sponsor with a title and some content and set the scope to Underwritten Content. Leave everything else default/blank.
2. Apply the underwriter to a post and leave all Sponsor Display Overrides to their defaults ("Inherit from sponsor").
3. On the front-end observe that the underwriter info isn't displayed at all.
4. Check out this branch, confirm that now the underwriter info is displayed at the top of the post (the default placement).
5. Play with different sponsor settings in the sponsor and the corresponding override settings in the post, and ensure that the display settings are reflected on the front-end (sponsor settings are used unless a different override is set on the post).

### Other information:

### DIFF
--- a/newspack-theme/inc/newspack-sponsors.php
+++ b/newspack-theme/inc/newspack-sponsors.php
@@ -481,7 +481,7 @@ function newspack_sponsored_underwriters_info( $sponsors = null, $id = null, $sc
 							return false;
 						}
 
-						return isset( $sponsor['sponsor_underwriter_placement'] ) ? 'inherit' === $sponsor['sponsor_underwriter_placement'] : true;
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
This looks good to me, @dkoo! 🙌 

Confirmed that the updates fixed the issue, and also that the sponsor settings are still being applied properly (changing the settings on the sponsor level, and on the per-post level, both work!).

PR #2077, opened 2023-03-30
feat: add multi-branded site support

leogermani:
### All Submissions:

Adds support to the multi branded site plugin

See https://github.com/Automattic/newspack-multibranded-site/pull/5

### Other information:

### DIFF
--- a/newspack-theme/functions.php
+++ b/newspack-theme/functions.php
@@ -1352,4 +1352,6 @@ if ( class_exists( 'Tribe__Events__Main' ) ) {
 /**
  * Multi-branded plugin support
  */
-require get_template_directory() . '/inc/newspack-multibranded-site-plugin.php';
+if ( class_exists( 'Newspack_Multibranded_Site\Customizations\Theme_Colors' ) ) {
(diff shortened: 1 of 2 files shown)


newspack-theme/functions.php
```diff
@@ -1348,3 +1348,8 @@ function newspack_dequeue_mediaelement() {
 if ( class_exists( 'Tribe__Events__Main' ) ) {
 	require get_template_directory() . '/inc/the-events-calendar.php';
 }
+
+/**
+ * Multi-branded plugin support
+ */
+require get_template_directory() . '/inc/newspack-multibranded-site-plugin.php';
```
laurelfulford, line 1355: What do you think about adding a check before requiring this (similar to the other plugins above) -- that way we don't run the code if not needed! 🙂

laurelfulford (approved):
This looks good! I added a small nitpick but I don't see any issues with this PR without adding that (everything seems good when the plugin is disabled), so I don't think it's a blocker - marking as approved!

newspack-theme/functions.php
leogermani: good call

PR #2093, opened 2023-04-14
fix: correct the default color used for the highlight menu label

laurelfulford:
### All Submissions:

The "Highlight" menu displays the menu's name as a label. In the default styles, this label was set to use the primary colour, but in the custom colours code, it was set to use the secondary colour, so it only ever used the primary colour when a site used the default colours (so basically, only on demo sites).

This PR updates the default styles, and removes a duplicated instance of the menu label style in the color-patterns.php file (it was set to use the primary then secondary colour there, with the latter overriding the former).

See 1204261432690303-as-1204394333707405

1. Assign a menu to the "Highlight" menu space.
2. Switch your site to use the default colour palette; note that the menu label is blue:

image

3. Switch your site to use custom colours
4. Note that the menu label uses your secondary colour (or if your secondary colour is very light, note that it is using a mid-grey as a legible fallback):

image

image

5. Apply this PR and run `npm run build`.
6. Confirm the custom colours act the same, but if you switch to the default colours, you now get the secondary colour (also grey):

image

7. Switch to Newspack Nelson, and enable the header background colour if it isn't already. The highlight menu appears against this header, so the label should be black against lighter custom headers, and white against darker custom headers. It should use the secondary colour if there's no header background.

No background (the theme defaults to a light grey to distinguish the header from content area):

image

Dark background:
image

Light background:
image

### Other information:

### DIFF
--- a/newspack-theme/inc/color-patterns.php
+++ b/newspack-theme/inc/color-patterns.php
@@ -80,7 +80,6 @@ function newspack_custom_colors_css() {
 			.mobile-sidebar .nav1 .sub-menu > li > a,
 			.mobile-sidebar .nav1 ul.main-menu > li > a,
 			.wp-block-file .wp-block-file__button,
-			.highlight-menu .menu-label,
 			/* Header default background; default height */
 			body.h-db.h-dh .site-header .nav3 .menu-highlight a,
 			.comment .comment-author .post-author-badge,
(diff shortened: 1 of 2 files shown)


leogermani (approved):
it works like a charm!

laurelfulford:
Thanks @leogermani! 🙌

PR #2165, opened 2023-07-21
feat: add styles for the details block

laurelfulford:
### All Submissions:

This PR adds some default theme styles for the Details block coming in WordPress 6.3. 

See 1204989005688794-as-1205113460196715.

1. Test on a site running the current WP 6.3 RC -- you can use the [Beta Tester plugin](https://en-ca.wordpress.org/plugins/wordpress-beta-tester/) to set this up on any test site.
3. Add several variations of the Details block to the editor: a single block, two or more blocks back to back, blocks with backgrounds... 
4. View on the front-end and in the editor and do a visual review:
* The details summary should use the theme's header font
* The details block should be framed by a top and bottom border
* When more than one block appears in a row, the blocks should appear spaced evenly

Single block:

image

Group of blocks:

image

image

Blocks when opened:

image

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1162,6 +1162,28 @@ hr {
 	}
 }
 
+//! Details block
+.wp-block-details {
+	border: solid var( --newspack-theme-color-border );
+	border-width: 1px 0;
+	padding: 0.5rem 0.5rem 0.5rem 1.5rem;
+
+	> :not(summary) {
+		margin-bottom: 16px;
+		margin-top: 16px;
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
Probably an edge case and not sure it can even be improved! But when you apply a background color to the whole Details block, the inner blocks are very slightly off-center due to the > icon.

Screen Shot 2023-07-21 at 4 23 50 PM

dkoo (approved):
Looking good to me, now! Thanks for the revision.

laurelfulford:
Thanks @dkoo! 🙌

PR #2166, opened 2023-07-21
feat: add styles for the footnote block

laurelfulford:
### All Submissions:

This PR adds a `scroll-margin` to the new footnote block's anchors when the theme's header is set to sticky.

See 1204989005688794-as-1205113460196718

1. Test on a site running the current WP 6.3 RC -- you can use the [Beta Tester plugin](https://en-ca.wordpress.org/plugins/wordpress-beta-tester/) to set this up on any test site.
2. Set up a page with the footnote block, and add a few footnotes throughout the content.
3. Turn on the sticky header under Customizer > Header Settings > Appearance. 
4. Try clicking the footnote links at the bottom; note that when the links go to content near the top of the page, they get covered by the header.
5. Apply this PR and run `npm run build`. 
6. Repeat step 4 and confirm your links don't get covered. 
7. Test this setting at different breakpoints.
8. Navigate to Customizer > Header Setttings > Appearance and enable the short header; repeat testing the links at different breakpoints.
9. Navigate to Customizer > Header Settings > Subpage Header, and check the "Use simple header on subpages" box; repeat testing at different breakpoints.

Note: since this fix uses static CSS, the footnote may be a bit further away from the header, depending on your settings; this is because the distance is based on approximate guesses. If this is subpar -- or if we run into too many issues with live sites and the exact distance (especially in cases where they may have customized their header with CSS) we can look at a JS-based solution.

### Other information:

### DIFF
--- a/newspack-theme/sass/blocks/_blocks.scss
+++ b/newspack-theme/sass/blocks/_blocks.scss
@@ -1185,22 +1185,29 @@ hr {
 }
 
 //! Footnote styles
+.wp-block-footnotes li:focus {
+	outline: thin dotted;
+}
+
 .h-stk {
-	sup.fn a {
-		scroll-margin-top: 80px;
+	sup.fn a,
+	.wp-block-footnotes li {
+		scroll-margin-top: 115px;
 
 		@include utilities.media( tablet ) {
 			scroll-margin-top: 180px;
 		}
 	}
 
-	&.h-dh sup.fn a {
+	&.h-dh sup.fn a,
+	&.h-dh .wp-block-footnotes li {
 		@include utilities.media( tablet ) {
 			scroll-margin-top: 210px;
 		}
 	}
 
-	&.h-sub:not(.home) sup.fn a {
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
This is a good fix! I have two suggestions:

1. At the smallest viewport when going back, the link still gets slightly covered on my test site. Should we increase the margin slightly for that smallest viewport?

Screen Shot 2023-07-26 at 11 31 11 AM

2. When clicking the footnoot number in content to jump down to the footnote itself, the footnote gets covered by the sticky heading. Is it possible to add `scroll-margin-top` to that element as well?

laurelfulford:
Ohhh good catches, @dkoo!

I've upped the top margin for the regular mobile header a bit, and added some for the actual footnotes (I was testing without any content below those, and totally missed that they could also get covered up! 

Just let me know if you spot anything else weird!

dkoo (approved):
Thanks for the fixes—looking good!

laurelfulford:
Awesome - thanks!

PR #2170, opened 2023-08-07
Introduce a filter for the social icons map

philipjohn:
### All Submissions:

Introduces a new hook for filtering the social icons map used to determine which logo to show in the social links menu.

With the introduction of Mastodon and the plethora of instances that could be used, including self-hosted instances, it would good to allow plugins and/or child themes to filter the social icons map to add additional mastodon instance addresses. This filter achieves that and can be used like so:

```
/**
 * Add our mastodon server to social icon map in Newspack.
 * 
 * @param array $social_icons_map Array of matches for social icons.
 * 
 * @return array Modified array of matches for social icons.
 */
function lichfieldlive_newspack_social_icons_map( $social_icons_map ) {
	$social_icons_map['mastodon'] = array_merge(
		$social_icons_map['mastodon'],
		[
			'social.lichfieldlive.co.uk',
		]
	);
	return $social_icons_map;
}
add_filter( 'newspack_social_icons_map', 'lichfieldlive_newspack_social_icons_map' );
```

It could also be used, for example, by Mastodon plugins such as the [ActivityPub plugin](https://wordpress.org/plugins/activitypub/) to ensure the connected server is recognised. A further change to the Newspack Theme could add a compatibility layer for plugins like ActivityPub.

1. Create or edit the Social Links Menu
2. Add a custom link to a Mastodon server account that does not already exist in the Newspack Theme's list (e.g. https://social.lichfieldlive.co.uk/@news)
3. Using a small plugin, or child theme, create a function to add the social.lichfieldlive.co.uk domain to the social icons map (see the above example)
4. Observe that the new item in the Social Links Menu has the Mastodon icon correctly applied (see screenshot below)

Screenshot from 2023-08-07 10-43-42

Note that as the social icons map is a static class variable it wasn't possible to add the filter to that array directly to only focus on the Mastodon map, so instead I had to filter the whole list.

### Other information:

### DIFF
--- a/newspack-theme/classes/class-newspack-svg-icons.php
+++ b/newspack-theme/classes/class-newspack-svg-icons.php
@@ -50,7 +50,17 @@ class Newspack_SVG_Icons {
 		static $regex_map; // Only compute regex map once, for performance.
 		if ( ! isset( $regex_map ) ) {
 			$regex_map = array();
-			$map       = &self::$social_icons_map; // Use reference instead of copy, to save memory.
+
+			/**
+			 * Filters the mapping of social icons to URLs.
+			 * 
+			 * Allows plugins or child themes to insert alternative mappings for recognition in the social links menu.
+			 * 
+			 * @since 1.76.0
+			 * 
+			 * @param array $servers Social icons map.
+			 */
+			$map = apply_filters( 'newspack_social_icons_map', self::$social_icons_map );
 			foreach ( array_keys( self::$social_icons ) as $icon ) {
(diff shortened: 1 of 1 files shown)


leogermani (changes requested):
Looks good and much needed.

If I may give one suggestion. Can we make the static attribute static and create a public `get_social_icons_map` method that returns the filtered array? Doing this we'll assure we will always have the filtered values anywhere we need to use this. What do you think?

adekbadek:
@philipjohn ping :) 👆

leogermani:
Closed in favor of https://github.com/Automattic/newspack-theme/pull/2212

PR #2176, opened 2023-09-04
fix: correct group block alignment in editor preview

### LINKED ISSUE #1973: Group block: Contents don't preview centred in editor when block is full-width
**Describe the bug**
In the editor, the contents of a full-width group block are no longer previewing as centred, even though they still appear that way on the front-end. This may have been a missed issue introduced in WP 6.1; I didn't spot it during testing, but this wasn't happening before so that seems possible. It doesn't appear to affect similar blocks like the Cover block.

Edited to add: this is just an issue when the block is not populated, so it's not a really obviously-breaking issue. 

**To Reproduce**
1. Add a group block to the editor and make it full-width.
2. Note that the "content area" of the Group block is flush left, but still maintains the default 1200px max width. It will be centre aligned once you add another block to it. 

Group block, set to 'full width' but not populated:
image

Group block set to 'full width', with a paragraph block inside; it's correctly lined up to the page title in the One Column Wide template:
image

laurelfulford:
### All Submissions:

This PR fixed a preview issue with the Group block, where the placeholder in wide or full-width group blocks would not be centred. 

See 1200550061930446-as-1203439986426071; Closes #1973

1. Create a new post or page and set it to use the one column template; publish and refresh (to make sure the editor is using the correct preview behaviour). 
2. Add a group block and make it full width. 
3. Note the preview placeholder placement when this empty block is selected - it's flush left rather than being centred:

image

5. Confirm that the placeholder is now centred.

image

6. Test with the Row and Stack options, and the wide and regular width options to confirm there aren't any issues. 

### Other information:

### DIFF
--- a/newspack-theme/sass/style-editor-base.scss
+++ b/newspack-theme/sass/style-editor-base.scss
@@ -421,6 +421,14 @@ div[data-align='full'] {
 	}
 }
 
+[data-type="core/group"]{
+	&.is-selected .block-list-appender,
+	.block-list-appender:only-child {
+		margin-left: auto;
+		margin-right: auto;
+	}
+}
+
 /** === Button === */
 .wp-block-buttons {
 	--wp--style--block-gap: 10px;


dkoo (approved):

laurelfulford:
Thanks @dkoo!

PR #2183, opened 2023-10-02
feat(customizer): change default template's name

adekbadek:
### All Submissions:

As [suggested here](https://github.com/Automattic/newspack-plugin/pull/2670/files#r1341655482), the "Default" name for the default template is not too descriptive. This PR changes the name to "With sidebar". 

1. Open the Customizer, navigate to Template Settings -> Post Settings
2. Observe that the first option in the "Default Post Template" select is named "With Sidebar"

### Other information:

### DIFF
--- a/newspack-theme/inc/customizer.php
+++ b/newspack-theme/inc/customizer.php
@@ -408,7 +408,7 @@ function newspack_customize_register( $wp_customize ) {
 		array(
 			'type'    => 'radio',
 			'label'   => __( 'Colors', 'newspack' ),
-			'choices'  => array(
+			'choices' => array(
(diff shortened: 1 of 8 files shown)


laurelfulford (approved):
This works as described! 

There is still some inconsistency with how the name shows in the editor, but there doesn't seem to be a nice way to change that -- there's [a filter](https://developer.wordpress.org/reference/hooks/default_page_template_title/), but it affects whatever the default template is vs. just a certain post type/situation (so dropping it in as is renames the front page template too, even though front-page.php doesn't have a sidebar). Given that CPTs and other cases could complicate this more, I think it's probably best to stick with this and keep it simple! 🙂

PR #2191, opened 2023-10-25
feat(woocommerce): remove woocommerce templates and dequeueing

adekbadek:
### All Submissions:

Templates will be handled by Newspack Blocks (as of https://github.com/Automattic/newspack-blocks/pull/1561), dequeuing by Newspack Blocks (https://github.com/Automattic/newspack-blocks/pull/1572)

1. Switch `newspack-blocks` to `feat/dequeue-wc-styles` branch (https://github.com/Automattic/newspack-blocks/pull/1572)
1. Complete a transaction via the standard WC flow*, observe the flow uses standard WC templates (not the shortened ones)
3. Complete a transaction via the Donate block (with modal checkout), observe it uses customised templates (shortened) 

\* create a non-donation product and visit `/shop` page

### Other information:


dkoo (approved):
Works for me. One thing I encountered which may only apply to local dev environments, but I was getting this fatal error upon proceeding to the Checkout template in step 2:

```
[27-Oct-2023 11:43:55 UTC] PHP Warning:  include(/var/www/html/wp-content/themes/newspack-theme/woocommerce/checkout/form-checkout.php): Failed to open stream: No such file or directory in /var/www/html/wp-content/plugins/woocommerce/includes/wc-core-functions.php on line 345
[27-Oct-2023 11:43:55 UTC] PHP Warning:  include(): Failed opening '/var/www/html/wp-content/themes/newspack-theme/woocommerce/checkout/form-checkout.php' for inclusion (include_path='.:/usr/share/php') in /var/www/html/wp-content/plugins/woocommerce/includes/wc-core-functions.php on line 345
```

Restarting PHP in my environment fixed it, but it makes me worry that PHP could be caching include paths and might cause issues in live sites.

dkoo (changes requested):
Wait, removing the Approved status. Where are we moving the custom `form-checkout.php` and `thankyou.php` template files to after this change?

adekbadek:
These templates are in `newspack-blocks` repo, in `src/modal-checkout/templates` directory. On `master` branches they are in both Theme and Blocks, which is redundant. This PR removes them from the Theme, so Blocks only are responsible for changes to WC templates. This is because we only want to hijack the templates in the context of donations, if the site sells e.g. t-shirts, standard WC templates should be used.

dkoo (approved):
Looking good now!

PR #2201, opened 2023-11-08
fix: clears woo template cache

leogermani:
Needed after files removed in 498192a4323771803626e5695f60b79c0adaf10e

### All Submissions:

In https://github.com/Automattic/newspack-theme/pull/2191 we removed two Woo templates from the theme.

In sites using persistent cache, Woo would still try to load the templates from the theme and the checkout page would break.

This PR will clear the cache to make sure it doesn't happen

⚠️ This change should be merged into the current alpha, too. 

Here is what you need if you want to test this

1. Make sure memcached is on
2. Checkout a version of the theme before https://github.com/Automattic/newspack-theme/pull/2191
3. Add a product to the cart and visit the checkout page (the regular checkout page, not our modal checkout)
4. Checkout theme on `master`
5. Visit checkout again and confirm the page is blank and there's an error in the log complaining about the missing file
6. Checkout this branch and confirm the Checkout page is back

### Other information:

### DIFF
--- a/newspack-theme/woocommerce/templates.php
+++ b/newspack-theme/woocommerce/templates.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * This folder holds Woocommerce templates that override the default Woo templates.
+ *
+ * If you add or remove a template from this folder, it's a good idea to clear the template cache
+ * to make sure the changes take effect - and not break anything - on sites that use persistent caching
+ *
+ * @package newspack-theme
+ */
+
+/**
+ * When adding or removing a template from this folder, increase the templates_version variable below.
+ */
+$newspack_theme_woo_templates_version = 1;
+
+if ( get_option( 'newspack_theme_woo_templates_version', 0 ) !== $newspack_theme_woo_templates_version ) {
+	wc_clear_template_cache();
+	update_option( 'newspack_theme_woo_templates_version', $newspack_theme_woo_templates_version );
+}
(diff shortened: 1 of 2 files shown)


newspack-theme/woocommerce/templates.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @package newspack-theme
+ */
+
+/**
+ * When adding or removing a template from this folder, increase the templates_version variable below.
+ */
+$newspack_theme_woo_templates_version = 1;
+
+if ( get_option( 'newspack_theme_woo_templates_version', 0 ) !== $newspack_theme_woo_templates_version ) {
```
dkoo, line 16: Edge case, but this ends up throwing a fatal if WooCommerce is deactivated unless we check for the existence of the function first:

```suggestion
if ( function_exists( 'wc_clear_template_cache' ) && get_option( 'newspack_theme_woo_templates_version', 0 ) !== $newspack_theme_woo_templates_version ) {
```

dkoo (changes requested):
For some reason I couldn't replicate the bug on a test site despite having memcached running, but I did see this happen on a live site. The cache clear should fix things for other live sites going forward. Just one issue I saw, below.

newspack-theme/woocommerce/templates.php
leogermani: Good call

dkoo (approved):

PR #2215, opened 2023-12-13
Improve support for distributed guest authors

leogermani:
Handle guest authors distributed via Newspack Network and Distributor plugins

### All Submissions:

Improves the support for the distributed post authors by Newspack Network. 

Basically - remove the `a` tag for authors without a URL.

Also sends an extra information in the `newspack_author_bio_name` filter.

See https://github.com/Automattic/newspack-network/pull/25

### Other information:

### DIFF
--- a/newspack-sacha/template-parts/post/author-bio.php
+++ b/newspack-sacha/template-parts/post/author-bio.php
@@ -21,23 +21,33 @@ if ( function_exists( 'coauthors_posts_links' ) && is_single() && ! empty( get_c
 	foreach ( $authors as $author ) {
 
 		if ( '' !== $author->description ) {
-			$author_avatar = coauthors_get_avatar( $author, 80 );
+			// avatar_img_tag is a property added by Newspack Network plugin to distributed posts.
+			$author_avatar = $author->avatar_img_tag ?? coauthors_get_avatar( $author, 80 );
+			$author_url    = get_author_posts_url( $author->ID, $author->user_nicename );
 			?>
 
 			<div class="author-bio">
 				<div class="author-bio-text">
 					<div class="author-bio-header">
 						<?php if ( $author_avatar ) { ?>
-							<a class="author-link" href="<?php echo esc_url( get_author_posts_url( $author->ID, $author->user_nicename ) ); ?>" rel="author">
+							<?php if ( '#' !== $author_url ) : ?>
+								<a class="author-link" href="<?php echo esc_url( $author_url ); ?>" rel="author">
+							<?php endif; ?>
 								<?php echo wp_kses( $author_avatar, newspack_sanitize_avatars() ); ?>
-							</a>
+							<?php if ( '#' !== $author_url ) : ?>
+								</a>
+							<?php endif; ?>
 						<?php } ?>
 
 						<div>
 							<h2 class="accent-header">
-								<a class="author-link" href="<?php echo esc_url( get_author_posts_url( $author->ID, $author->user_nicename ) ); ?>" rel="author">
-									<?php echo wp_kses( apply_filters( 'newspack_author_bio_name', $author->display_name, $author->ID ), array( 'span' => array( 'class' => array() ) ) ); ?>
-								</a>
+								<?php if ( '#' !== $author_url ) : ?>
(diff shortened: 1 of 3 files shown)


miguelpeixe:
Can't this be a filter inside the Newspack Network plugin so it's contained in the same codebase? Distributor has a similar strategy to filter the author to make it appear as the site name.

miguelpeixe:
Thank you for explaining!

laurelfulford (changes requested):
This is so cool, @leogermani! 🙌 

I tested posts with single guest authors, multiple guest authors, mix of guest and regular, and a guest mapped to a regular author. The only issue I've run into is that the distributed guest author avatars don't seem to be syncing, but only for the byline at the top -- they appear as expected in the author bio at the bottom of the post. The regular author avatars work 100% as expected. 

Screen Shot 2023-12-21 at 10 29 04 AM

Screen Shot 2023-12-21 at 10 29 10 AM

Job titles, bios, and email addresses were brought over as expected!

**Styles**
There was some discussion about whether these should be styled to match the linked authors. Personally, I'm leaning towards leaving them as the default grey -- I think that could give some visual feedback that they're not the same as the other names and (hopefully!) help set expectations that they're not going to be links. I'm open to discussing that further, though! I feel like either way we land with this, there will be publishers who want it to work "the other way" (either match the regular links, or look less like the links).

**Custom CSS**
I think there's a _small_ risk existing custom CSS won't be applied as expected to these new unlinked bylines because an HTML element has been removed. I think it's super-minor though -- and it won't affect existing posts since it's a new feature, just posts going forward. I think we can tackle this as it comes up (it will likely also happen if we opt to style these to match the links).

laurelfulford (changes requested):
Thanks @leogermani! [8367482](https://github.com/Automattic/newspack-theme/pull/2215/commits/836748285ce6b0e10b3f32389592a0dd8f780d06) fixed the issue for me:

image

I have one more change that I missed in my initial review: Newspack Sacha also has its own [/author-bio.php file](https://github.com/Automattic/newspack-theme/blob/master/newspack-sacha/template-parts/post/author-bio.php) that would need the same changes as the main theme's file, just to make sure it's picking up the same updated author info. The separate file was to help make its centred layout easier to accomplish, but more than once I wish I'd just reworked the original file earlier on to avoid adding it!

leogermani:
Thanks for the review @laurelfulford 

I've updated the checks with some recent changes on the PR in the Newspack Network plugin, but it should work just the same.

And I applied the changes to the Sacha theme.

All those `if`s are not pretty, but I think it's good enough for now.

I also tested with co-authors-plus disabled in both themes just to make sure :)

laurelfulford (approved):
Thanks @leogermani! 

> All those ifs are not pretty, but I think it's good enough for now.

I thought you meant the ones that were already there 😆 These files are already full of so many `if`s and `else`s, adding more at this point can't hurt!

I retested this with Newspack and Newspack Sasha, with and without Co-Author Plus, and the single posts work as expected! 🎉 

I found one more spot that may need changes, but it's not something that's actually being addressed in this PR yet, so I'm marking this as approved in case it's just easier to do it as a separate fix: clicking around, I noticed that the 'real' authors that are brought over via the Network still have the site's name as their name on the author archives. For example, this author has the display name 'laurel' but it's still showing 'Newspack Dev' (my spoke site) on the Hub archive: 

image

Just let me know if you have any questions about that, or can't reproduce! I don't remember spotting that when I last tested, but I don't think I actually clicked over to the transferred author archive pages. (Also just let me know if I should spin up an Asana task for that with better steps to replicate!).

leogermani:
@laurelfulford Can you check if you unmarked the "Override Author Byline" option in Distributor > Settings?

image

laurelfulford:
> Can you check if you unmarked the "Override Author Byline" option in Distributor > Settings?

@leogermani I had it checked when this happened, but I also unchecked it both on the Hub and Spoke and could see the same issue 😕 I'll take some time this AM trying to recreate in a different set up, and can follow up with a separate task if I'm still seeing it with a link!

laurelfulford:
@leogermani Just wanted to follow up and confirm I CAN'T recreate the above on two JN sites set up with Hub & Spoke plus this PR, so that issue was totally user error 🙂

PR #2292, opened 2024-05-08
fix(homepage-posts): handle "time ago" format in HBP

adekbadek:
### All Submissions:

Fixes a regression introduced in https://github.com/Automattic/newspack-blocks/pull/1662 – "time ago" format is not respected by the HPB with this changes. This PR fixed that. 

1. Navigate to Customizer > Template Settings > Post Template and enable "Use "time ago" date format".
2. Add a homepage post block to a page; make sure it has at least one post new enough to fall inside of the 'time ago' range (it can also be changed in the Customizer). 
3. Observe that the "time ago" format is respected by the HPB both in the editor and on the frontend

### Other information:

---
  - https://app.asana.com/0/0/1207084341012277


laurelfulford (approved):
LGTM! 🚢 

I also tested with different languages just in case, all looks good to me!

PR #2301, opened 2024-05-20
fix: handle attachement fetch failure

adekbadek:
### All Submissions:

As seen on a live site, `wp_get_attachment_metadata` may return `false`. If it's not handled, a PHP Warning will be logged. This PR adds handling of a `false` return value from `wp_get_attachment_metadata`, and refactors code for readability.

Set a featured image on a post or page, observe it behaves as expected.

### Other information:

### DIFF
--- a/newspack-theme/inc/template-functions.php
+++ b/newspack-theme/inc/template-functions.php
@@ -31,8 +31,9 @@ if ( ! function_exists( 'newspack_featured_image_position' ) ) :
 			return $position;
 		}
 
+		$image_wide_width = 1200;
 		if ( (
-			'large' === $position && 1200 > $thumbnail_info['width'] )
+			'large' === $position && $image_wide_width > $thumbnail_info['width'] )
 			|| ! in_array( get_post_type(), newspack_get_featured_image_post_types() )
 		) {
 			$position = 'small';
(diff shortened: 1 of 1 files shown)


newspack-theme/inc/template-functions.php
```diff
@@ -17,31 +17,24 @@ function newspack_featured_image_position() {
...
 
-		if ( ( 'large' === $position && 1200 > $img_width ) || ! in_array( get_post_type(), newspack_get_featured_image_post_types() ) ) {
+		// Get thumbnail
+		$thumbnail_info = wp_get_attachment_metadata( get_post_thumbnail_id() );
+		if ( $thumbnail_info === false ) {
+			return $position;
+		}
+
+		if ( (
+			'large' === $position && 1200 > $thumbnail_info['width'] )
```
chickenn00dle, line 35: Nit: Since we are refactoring, maybe we can move this 1200 into a constant. It would make it easier to track down in the future and would provide better context around what the number represents.

chickenn00dle (approved):
Looks good. Left a small nitpick, but feel free to ignore it.

newspack-theme/inc/template-functions.php
adekbadek: Sure, done in eb0176af3d1553043f2e4f3bad38c3626f1eeb4a

PR #2304, opened 2024-05-23
fix: dont update option on every request

claudiulodro:
### All Submissions:

Fixes an issue I noticed while debugging something. The template check never succeeds, so the template cache is cleared on every page load and the template version option is updated on every page load. This is because the option value is the string `"1"` but the check does a strict compare with integer `1`.

Screenshot 2024-05-22 at 8 58 59 PM

I have this as a hotfix because removing this helps performance and it's a tiny change.

1. Install Query Monitor. Go to any page. In the DB section filter on UPDATE and see the option is updated on every page load.

Screenshot 2024-05-22 at 9 42 40 PM

2. Apply this patch. Refresh the page. See the option is not updated on every page load any more.

### Other information:

### DIFF
--- a/newspack-theme/woocommerce/templates.php
+++ b/newspack-theme/woocommerce/templates.php
@@ -13,7 +13,7 @@
  */
 $newspack_theme_woo_templates_version = 1;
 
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Great find!

PR #2318, opened 2024-06-14
fix: update dependencies to avoid conflict in WP 6.6

dkoo:
### All Submissions:

Refactoring required for changes in https://github.com/Automattic/newspack-scripts/pull/209. Note that this PR will not pass CI tests and should NOT be merged until https://github.com/Automattic/newspack-scripts/pull/209 is released as an NPM package, so this repo can install it as a dependency.

Follow instructions in https://github.com/Automattic/newspack-scripts/pull/209.

### Other information:


package.json
```diff
@@ -1,52 +1,57 @@
...
+		"build": "run-p \"build:*\"",
+		"watch": "run-p \"watch:*\"",
+		"lint": "npm run lint:scss && npm run lint:js",
+		"lint:js": "newspack-scripts wp-scripts lint-js '**/{src,includes}/**/*.{js,jsx,ts,tsx}'",
+		"lint:js:staged": "newspack-scripts wp-scripts lint-js --ext .js,.jsx,.ts,.tsx",
+		"format:js": "newspack-scripts wp-scripts format '**/{src,includes}/**/*.{js,jsx,ts,tsx}'",
+		"lint:scss": "newspack-scripts wp-scripts lint-style '**/{src,includes}/**/*.scss' --customSyntax postcss-scss",
+		"lint:scss:staged": "newspack-scripts wp-scripts lint-style --customSyntax postcss-scss",
+		"format:scss": "newspack-scripts wp-scripts lint-style '**/{src,includes}/**/*.scss' --customSyntax postcss-scss --fix",
+		"release": "npm run build && npm run semantic-release",
```
chickenn00dle, line 33: Running `npm run release` returns the following output after the build process completes:

```
[newspack-scripts@1.0.0] Unknown script "proxy".
```

package.json
```diff
@@ -1,52 +1,57 @@
...
+	},
+	"homepage": "https://github.com/WordPress/newspack#readme",
+	"browserslist": [
+		"extends @wordpress/browserslist-config"
+	],
+	"lint-staged": {
+		"*.scss": "npm run lint:scss:staged",
+		"*.js": "npm run lint:js:staged"
+	},
+	"scripts": {
```
chickenn00dle, line 16: There are a bunch of npm scripts included in package.json that are not explicitly mentioned in the testing steps. For example, `npm run format:scss` which right now does seem to format sass files, but exits with an error code 2 and no useful messaging. Are these meant to be removed/cleaned up?

chickenn00dle (changes requested):
Thanks for this @dkoo!

I left a few comments where I ran into some issues. In addition I had one other note worth bringing up that is likely unrelated to this PR, but is still worth mentioning:

I noticed a bunch of JS errors in console while smoke testing in customizer. This happen in both `trunk` as well as the updated branches, and for both the current WP and 6.6

Screenshot 2024-06-20 at 13 18 26

chickenn00dle (dismissed):
Whoops. This comment was meant to be posted in the blocks PR 😨

package.json
dkoo: Whoops, this was a holdover from a prior commit. Fixed in 0f2f821c

package.json
```diff
@@ -1,52 +1,57 @@
...
+		"watch:js": "newspack-scripts wp-scripts start --output-path newspack-theme/js/dist",
+		"build:scss": "node scripts/compile-scss.js",
+		"watch:scss": "node scripts/compile-scss.js --watch",
+		"build": "run-p \"build:*\"",
+		"watch": "run-p \"watch:*\"",
+		"lint": "npm run lint:scss && npm run lint:js",
+		"lint:js": "newspack-scripts wp-scripts lint-js '**/{src,includes}/**/*.{js,jsx,ts,tsx}'",
+		"lint:js:staged": "newspack-scripts wp-scripts lint-js --ext .js,.jsx,.ts,.tsx",
+		"format:js": "newspack-scripts wp-scripts format '**/{src,includes}/**/*.{js,jsx,ts,tsx}'",
+		"lint:scss": "newspack-scripts wp-scripts lint-style '**/{src,includes}/**/*.scss' --customSyntax postcss-scss",
```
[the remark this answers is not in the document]
dkoo: You're right—committed your suggested change.

composer.lock
```diff
@@ -1,4 +1,1866 @@
 {
+<<<<<<< HEAD
```
miguelpeixe, line 2: A pending conflict resolution got its way in a commit. The lock file is currently broken:

```

  "./composer.lock" does not contain valid JSON
  Parse error on line 1:
  {<<<<<<< HEAD	"_read
  ^
  Expected one of: 'STRING', '}'
```

miguelpeixe (changes requested):

composer.lock
dkoo: Whoops! Fixed in 868e96b4, thanks for catching!

package.json
dkoo: To wrap up this thread, we've decided that `fix:js` and `format:js` should both be present for all repos, but that they do different things: `fix:js` runs ESLint autofixing (for code quality issues) whereas `format:js` runs Prettier formatting (which should only affect code style and not functionality). We're mainly concerned about`fix:js` working as the ESLint stuff is more important to our workflows, but in the future we could decide to formally update our code style to match WP's Prettier configs.

miguelpeixe:
I find it strange that when I build/watch I get a bunch of deprecation warnings, e.g.:

```
    ╷
51  │ ┌     &,
52  │ │     &:visited:not(:hover) {
53  │ │         background: #d33;
54  │ │         color: #fff;
55  │ │     }
    │ └─── nested rule
... │
58  │       margin-left: auto;
    │       ^^^^^^^^^^^^^^^^^ declaration
    ╵
    newspack-theme/sass/navigation/_menu-mobile-navigation.scss 58:2  @use
    newspack-theme/sass/navigation/_navigation.scss 15:1              @use
    newspack-theme/sass/style-base.scss 22:1                          @use
    newspack-joseph/sass/style.scss 3:1                               root stylesheet

Deprecation Warning: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls
```

But those are not caught with `npm run lint`. Is that expected?

dkoo:
@miguelpeixe I think this is because the Theme repo uniquely uses `sass` as a direct dev dependency, whereas the other repos just use whatever peer dependency `@wordpress/scripts` (and before that, Calypso Build) uses. Looks like the peer dependency specifies `^1.35.2` whereas the direct dependency in Theme is now `^1.77.8`. Looks like the deprecation warning was [just introduced](https://sass-lang.com/documentation/breaking-changes/mixed-decls/#deprecating-the-old-way) in `1.77.7`, so the linters from `@wordpress/scripts` won't show it since they're running an older version.

If it's obtrusive, we can always [silence the warnings](https://sass-lang.com/documentation/js-api/interfaces/options/#silenceDeprecations) in our Theme config. I'm fairly certain SASS wouldn't introduce a breaking change like this in a minor version, so the risk would be low.

miguelpeixe (approved):

PR #2351, opened 2024-07-26
fix: give the inline sign up form a max-width

laurelfulford:
### All Submissions:

This PR gives the inline sign-up form a max-width (along with https://github.com/Automattic/newspack-plugin/pull/3282 to add an extra CSS class).

I opted to add the styles in the theme because I'm not sure we're going to have the same setup/outcome with the block theme. Let me know if that doesn't make sense, though!

See 1205234045751551-as-1207890325042756

1. Apply this PR and https://github.com/Automattic/newspack-plugin/pull/3282 and run `npm run build`
2. Go to /my-account and confirm that the sign-up form appears roughly the same size as it does in the modal.
3. Try signing in, and confirm that the different screens (OTP, password) look okay.

image

image

image

4. Log out and confirm the sign out message uses the max-width, too. 

image

### Other information:

### DIFF
--- a/newspack-theme/sass/plugins/woocommerce.scss
+++ b/newspack-theme/sass/plugins/woocommerce.scss
@@ -969,6 +969,12 @@ table.variations {
  * Account section
  */
 .woocommerce-account {
+	.entry-content .newspack-reader-auth__inline-wrapper {
+			margin-left: auto;
+			margin-right: auto;
+			max-width: var( --newspack-ui-modal-width-s, 410px );
+	}
+
 	.woocommerce-MyAccount-navigation {
 		margin: 0 0 2rem;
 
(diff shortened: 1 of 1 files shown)


laurelfulford:
@kevinzweerink I could use your eyes on this just from a design perspective -- you can see what it looks like now by going to /my-account on one of the test sites.

chickenn00dle (approved):
Looks good @laurelfulford 

I'm going to go ahead and approve this since it works as expected, but I do think these styles might be better in `plugin` alongside the other auth styles.

That said, I'll defer to your opinion on this since you're more familiar with the styles for RAS-ACC.

laurelfulford:
Thanks @chickenn00dle!

I stuck the styles here mostly in case we wanted to do something different with the block theme 🙂 But there's other ways to do that if it's better to keep them all together. I was a little on the fence about placement so am okay with moving them! 

I'm going to spin up a fresh PR with both the plugin changes and CSS in one in the plugin and tag you for review just to confirm I haven't messed anything up! 😅

PR #2387, opened 2024-10-11
Minify styles

adekbadek:
### All Submissions:

Adds CSS minification to the created stylesheets.

Also fixes a warning triggered by Sass about rule nesting. Nested rules should be placed after other rules.

1. Run `npm run build` and observe the `<theme-slug>/style.css` files are minified
2. Smoke-check the front-end to see if it looks as expected

### Other information:

### DIFF
--- a/newspack-theme/sass/navigation/_menu-mobile-navigation.scss
+++ b/newspack-theme/sass/navigation/_menu-mobile-navigation.scss
@@ -48,15 +48,15 @@
 
 // Mobile CTA
 .site-header .mb-cta {
+	font-size: 0.7em;
+	margin-left: auto;
+	padding: #{0.6 * structure.$size__spacing-unit} #{0.5 * structure.$size__spacing-unit};
+
 	&,
 	&:visited:not(:hover) {
 		background: #d33;
 		color: #fff;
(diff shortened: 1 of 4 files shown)


dkoo (commented):
This is working as described, but while testing I noticed a warning in the SASS CLI output:

```bash
Deprecation Warning: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.

More info: https://sass-lang.com/d/legacy-js-api
```

Looking at the [migration documentation](https://sass-lang.com/documentation/breaking-changes/legacy-js-api/) and the [replacement `compile` function](https://sass-lang.com/documentation/js-api/#md:usage), it seems that Dart SASS now supports compressed/minified output natively. Would it be preferable to migrate to the new API and avoid having to add the `cssnano` dependency?

adekbadek:
@dkoo – I've tried upgrading to `compile`, but ran into errors. I've decided that since it works as it is, it's not worth investing more time into.

dkoo:
Ok, we can re-cross that bridge when `sass` introduces a new major version.

dkoo (approved):

PR #2435, opened 2025-02-10
feat: Implement corrections in theme.

Takshil-Kunadia:
### All Submissions:

This PR implements the tickest:- 
1. https://app.asana.com/0/1209292256643614/1209292256643619
2. https://app.asana.com/0/1209292256643614/1209292256643621

#### Changes:-
- Add a new global customizer setting “Corrections Location” (with options “Top” and “Bottom”) in the post_default_settings section.
- Add corrections template part to implement individual correction front-end.
- Update the single post template to read the corrections location from the above implemented theme mod.
- Add wrapper functions to Corrections class's methods to use in the theme.

NOTE: Testing this will require local set up or uploading modified newspack-plugin with the below mentioned change. I have prepared a PR to remove that code from the plugin codebase once this PR gets approved.

1. In the local setup, Open plugin's codebase and comment out [Line 55](https://github.com/Automattic/newspack-plugin/blob/trunk/includes/corrections/class-corrections.php#L55) to disable the current implementation.
2. Open the WordPress Customizer and Load a single post in the theme.
3. Open Template Settings > Post Settings > and check the Radio buttons Top & Bottom to see the impact.

### Other information:

#### Demo:-

https://github.com/user-attachments/assets/36c5149e-39d7-4a15-b780-7fe659055f29


Takshil-Kunadia:
Hi @laurelfulford @leogermani Please do have a look at this PR whenever possible. I wanted to implement JS postMessage selective refresh to this but since the "post content" is dynamic it wouldn't be as robust. Will keeping it as default refresh suffice or should I work on this?

thomasguillot:
`correction-title` should be a link that takes the reader to the corrections archive page – and to the #id of the correction.

Takshil-Kunadia:
@thomasguillot updated the PR with the change ✅ . The catch here is since we are using pagination, if the article is on other pages (page 2 or more...) This would not work. Not sure if there is any efficient way to handle this...

cc: @leogermani

newspack-theme/template-parts/post/corrections.php
```diff
@@ -0,0 +1,45 @@
...
+		$timezone           = \wp_timezone()->getName();
+		$correction_heading = sprintf(
+			'%s, %s %s %s:',
+			newspack_get_correction_type( $correction->ID ),
+			$correction_date,
+			$correction_time,
+			$timezone
+		);
+
+		if ( empty( $correction_archive_url ) ) {
```
leogermani, line 33: it's ok to hardcode the corrections post type slug up there on line 15. It's hardcoded elsewhere in the theme code anyways. No need to dynamically grab it from a post

newspack-theme/inc/template-functions.php
```diff
@@ -774,3 +774,40 @@ function newspack_inject_post_summary( $content ) {
...
+}
+
+/**
+ * Get the correction type for a post.
+ *
+ * @param int $post_id The post ID.
+ *
+ * @return string The correction type.
+ */
+function newspack_get_correction_type( $post_id ) {
```
leogermani, line 802: to be clear, this is `$correction_id` right? Let's rename it for better readability

leogermani:
> @thomasguillot updated the PR with the change ✅ . The catch here is since we are using pagination, if the article is on other pages (page 2 or more...) This would not work. Not sure if there is any efficient way to handle this...
> 
> cc: @leogermani

Yeah, I agree this doesn't work very well. Not sure if it makes sense... FTR, currently on Texas Tribune it's just a link to the archive, without any ID ([example](https://www.texastribune.org/2025/02/07/ut-dallas-student-newspaper-palestinian-protest-coverage/)).

thomasguillot:
> Yeah, I agree this doesn't work very well.

No worries, simply pointing to the archive page is fine then.

leogermani (approved):

newspack-theme/template-parts/content/content-single.php
```diff
@@ -57,6 +67,10 @@
 			)
 		);
 
+		if ( $corrections_enabled && 'bottom' === $correction_location ) {
+			get_template_part( 'template-parts/post/corrections' );
+		}
+
```
laurelfulford, line 73: I think this approach for placement could run into issues with plugins that add stuff to `the_content` -- for example, for the bottom the big one is Related Posts, which will end up between the post content and the correction: 

CleanShot 2025-02-13 at 12 40 20

Since the correction is more part of the content, what do you think about adding them to `the_content`?

newspack-theme/template-parts/post/corrections.php
```diff
@@ -0,0 +1,39 @@
...
+$corrections_archive_url = get_post_type_archive_link( 'newspack_correction' );
+?>
+
+<div class="newspack-corrections-module">
+	<?php
+	foreach ( $corrections as $correction ) :
+		$correction_content = $correction->post_content;
+		$correction_date    = \get_the_date( get_option( 'date_format' ), $correction->ID );
+		$correction_time    = \get_the_time( get_option( 'time_format' ), $correction->ID );
+		$timezone           = \wp_timezone()->getName();
```
laurelfulford, line 24: Just noting that WordPress's formatting can be a little strange for timezones, depending what you pick:

CleanShot 2025-02-13 at 13 21 13

laurelfulford (changes requested):
Thanks for your work on this, @Takshil-Kunadia! I think it's looking really good overall 🙂 

I added some feedback, some based on things I've done in the past adding options to the theme that have caused unexpected issues 😆 I'm happy to discuss any of them, or answer any questions -- just let me know!

newspack-theme/sass/site/primary/_posts-and-pages.scss
```diff
@@ -1001,3 +1001,10 @@ div.sharedaddy {
 		}
 	}
 }
+
+.newspack-corrections-module {
+	.correction {
+		color: var(--wp--preset--color--contrast-3, var(--newspack-theme-color-text-light));
+		font-size: var(--wp--preset--font-size--small, var(--newspack-theme-font-size-sm));
```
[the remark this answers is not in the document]
Takshil-Kunadia: Makes sense ☑️. I had added the default value as per the values shared in the task design [description](https://app.asana.com/0/0/1208740794953680/1209210672433460/f). But thinking about it from your perspective, it makes more sense to keep the NP colours as default.

newspack-theme/inc/customizer.php
```diff
@@ -976,6 +976,33 @@ function newspack_customize_register( $wp_customize ) {
 		)
 	);
 
+	// Add Setting and Control for Corrections.
```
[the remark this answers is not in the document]
Takshil-Kunadia: Yes a separate panel would be the better option for scalable inclusions of new correction settings/updates in the future.

newspack-theme/template-parts/content/content-single.php
Takshil-Kunadia: Yes, the current implementation appends the markup of corrections to the content. I went along with the current approach to create a clear separation between FE and processing logic, but this approach although logically sound will prove to be a hurdle in cases of plugins that work on the content like the one you shared. I believe that this logic then should be implemented in the plugin than in the theme.

newspack-theme/template-parts/post/corrections.php
Takshil-Kunadia: You're right—WordPress’s handling of timezones can be a bit quirky depending on the settings. But If we want to include timezone in the markup, to keep it true with the timings we will have to keep it as per settings of the site configured by the users. Unless we skip including timezone or switch to number based display of timezone( +5:30 ) and such.
cc: @leogermani

Takshil-Kunadia:
Thanks @laurelfulford for the feedbacks and insights! 💯

Takshil-Kunadia:
@leogermani 
After reviewing the feedback I'm in favor of keeping the corrections code in the plugin. This will let us centralize all corrections-related functionality, including adding a dedicated Customizer panel and flexible insertion via the_content.
I have raised a draft PR for it:- https://github.com/Automattic/newspack-plugin/pull/3751.

We also have the option to implement this functionality in theme, however, I am worried this will couple this functionality to the theme.

If we go ahead with either approach I will close the other PR and vice versa.

laurelfulford:
Thanks @Takshil-Kunadia! 

> After reviewing the feedback I'm in favor of keeping the corrections code in the plugin. This will let us centralize all corrections-related functionality, including adding a dedicated Customizer panel and flexible insertion via the_content.

There are a few things here -- the Customizer panel and inserting the corrections via `the_content` -- that will _only_ be needed for this classic theme, and not the block theme. I'm personally not attached to where this code lives, but if this all goes in the plugin we'll probably need to check for the kind of theme being used in a couple spots ([this function](https://developer.wordpress.org/reference/functions/wp_is_block_theme/) should do the trick!) 🙂

Takshil-Kunadia:
Thanks @laurelfulford . I agree that customizer settings and inserting the markup on `the_content` are really only needed for block themes. For sites that rely solely on the plugin, however, keeping this code within the plugin ensures that the functionality remains intact.

> but if this all goes in the plugin we'll probably need to check for the kind of theme being used in a couple spots ([this function](https://developer.wordpress.org/reference/functions/wp_is_block_theme/) should do the trick!) 🙂

Yess, this function will definitely do the trick 💯

Takshil-Kunadia:
Closing this PR in favour of https://github.com/Automattic/newspack-plugin/pull/3751 

Thank you all for your valuable feedback and time! 🙇

PR #2456, opened 2025-03-10
feat: add small image size for smaller featured images

laurelfulford:
### All Submissions:

This PR adds a 'small' image size (780px wide x infinite tall) to use for the 'Small' image placement instead of the default Large (1200px wide x infinite tall) size.

~~I can't think of a case where this should cause issues, but I'm going to stick this in draft mode for a bit and dig around in our more customized sites to make sure.~~ I checked some of our more customized sites and can't find a case where the 'Small' image size is being scaled up, but there's still a very small risk this will clash with some more custom styles out there (either ours or third-party ones). I'll make sure to document the change clearly in the release notes once it's ready!

(This PR also fixes the `set_post_image_size()` default value to something actually used in the theme, though we don't get the post thumbnail in the theme without calling a specific size!). 

See 1200550061930446-as-1209168099234266.

1. Create a post, add a featured image, and set it to use the 'Small' size. Publish.
2. Inspect the image and confirm it has `1200` in the width attribute.
3. Apply this PR, and if you're not running Jetpack's Site Accelerator on your test site, run `wp media regenerate` to make sure you have image sizes that match this updated size. 
4. Refresh your post and comfirm the width attribute on the featured image is `780`.
5. Try scaling down the webpage etc. and confirming that the `780` wide image is sufficiently wide when the default template switches to a single column.

### Other information:


adekbadek (approved):

laurelfulford:
Thanks @adekbadek!

PR #2457, opened 2025-03-11
fix(post-dates): use JS for "time ago" dates

naxoc:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?
* [ ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?

 This moves the "time ago" calculation from PHP to JS so the time ago does not get old with the cache.

See 26890605006346-as-1209535906175869

### Questions
- It uses [Intl.RelativeTimeFormat](https://caniuse.com/?search=Intl.RelativeTimeFormat) – I think the browser support is OK for that?
- Should we remove the PHP formatting to timeago too? I have not touched that code. You could say that it is a nice fallback from JS to just keep it.
- Does the selector I use in the JS cover all our cases? I tested with posts as much as I could and with comments.
- Am I adding the JS in the best way like that? Conditionally? 

1. Go to _Customizer > Template Settings > Post Settings_ and click "Use 'time ago' date format". Set the number of days to relatively low. Make sure you have posts that are older than that.
2. Create some posts or edit the publish date to be after the number of days you set in step 1.
3. Click around and see that dates are "time ago" on the posts you expect. The "Cut off for "time ago" date in days." should be respected. Lists, archives, etc.
4. Try changing the site language and see that the dates are output in that language.
5. Hover over a date and see that you get a title thingy with the actual date

### Other information:

---
  - https://app.asana.com/0/0/1209535906175869


newspack-theme/functions.php
```diff
@@ -502,6 +502,25 @@ function newspack_scripts() {
 			'fonts' => newspack_get_used_custom_fonts(),
 		]
 	);
+
+	if ( get_theme_mod( 'post_time_ago' ) ) {
+		wp_register_script( 'newspack-relative-time', get_theme_file_uri( '/js/dist/relative-time.js' ), [], wp_get_theme()->get( 'Version' ), true );
+
+		$cutoff_in_days = get_theme_mod( 'post_time_ago_cut_off', '14' );
```
adekbadek, line 509: The `'14'` default now lives in two places – here and in `inc/customizer.php`. To avoid confusion when updating it, it should rather be stored as a constant and defined in one place.

adekbadek (changes requested):
Works as described 🎉 Left one refactoring comment.

newspack-theme/functions.php
adekbadek: You're right, there's no precent for this. That's why the default is _already_ duplicated (also in `newspack-theme/inc/template-functions.php`, which I haven't noticed before :D 

I think the constants could be defined either in `newspack-theme/functions.php` or in a new file.

naxoc:
Thanks @adekbadek ! I added a constant and a small fix so the customizer is responsive to the cutoff being updated.

adekbadek (approved):

PR #2458, opened 2025-03-12
fix: update json files with MD5 hash versions

laurelfulford:
### All Submissions:

This PR fixes a 'doing it wrong' situation in the Newspack Theme translations. When I originally set up translations, I could only get the editor's JSON translations to load if I renamed all of the auto-generated MD5 hash JSON file names to match their original JS file names.

It turns out this is because I was generating the POT file with the `src` directory included.

This PR was made by making the POT file with ` wp i18n make-pot ./ ./languages/newspack-theme.pot --exclude='src'` in the /newspack-theme directory, and then generating each JSON file with `wp i18n make-json languages/xx_XX.po --no-purge`. It doesn't make changes to the translations (outside of line number changes and removing a bunch of /src cruft), but it should make the translation generation process a little smoother going forward. I've also updated our internal docs for this.

1. Apply this PR.
2. Under General > Settings, switch the site's language to "Français de Belgique" (which has "Article Subtitle" translated, making it easy to test).
3. Edit a post.
4. Confirm that Article Subtitle is translated:

CleanShot 2025-03-12 at 09 23 58

5. Spot-check the front-end and confirm standard theme strings are still translated; one good thing to check is the byline on single posts or the archive page, or the comment form, or even the Powered by Newspack in the footer.

### Other information:


adekbadek (approved):
Works as described. 

Perhaps it's as intended, but when I look at the `newspack-theme/languages` directory, I see support for six languages: `de_DE`, `es_ES`, `fr_BE`, `pl_PL`, `pt_PT`, `pt_BR`. The per-language files are different, though – for example:

- `pt_BR` has four `pt_BR-<md5>.json` files
- `pt_PT` has four `newspack-pt_PT-<md5>.json` files

Is that expected?

laurelfulford:
@adekbadek Oh god, I was so excited that the translations loaded I didn't even notice that 🤦‍♀️ 

It looks like the file prefix depended on whether or not the .po file had the line `"X-Domain: newspack\n"` or not - I manually added that to the es_ES and pt_BR files and regenerated their JSON files, and they now match the other languages.

There aren't a ton of editor strings translated for either language, but it looks like the files are working:

CleanShot 2025-03-14 at 10 10 26

CleanShot 2025-03-14 at 10 11 55

Thank you for spotting that! It was a silly error on my part. I'm going to go ahead and merge this one!

PR #2479, opened 2025-04-11
feat: add specific style for the atomic proxy bar

thomasguillot:
### All Submissions:

The Atomic Proxy Bar had been bothering me for a long time... it was invasive and visually unappealing. BUT... this PR fixes that! I did have to rely on quite a few `!important` declarations, but in this case, it feels justified.

__Before__
Before

__After__
After

1. What I did was simply adding

```html
<div id="atomic-proxy-bar" style="position: fixed;">PROXIED NEWSPACK</div>
```
to a page to simulate a staging/live environment in my local instance. You can also upload this theme to a newspack staging site if you want.

2. Check if you see the new "proxy banner"

### Other information:


adekbadek (changes requested):
I see why the default proxy bar is unappealing, but it's a visual indicator we're all used to – in and outside of the team. I think this change could create a lot of confusion, especially when someone from another team looks into an issue related to our sites. 
Also, even if we do agree it's a good idea, it should live in the internal plugin, not in this OSS repository.

PR #2486, opened 2025-04-24
fix: add custom site titles

rbcorrales:
### All Submissions:

Fixes the site title displayed for multibranded sites. It introduces a new custom site title functionality that allows filtering both the site name and URL using hooks. This is used to address an issue with handling multi-branded site title links, where the rendered URL points to the site's root instead of the brand's root.

The implementation includes:
- A new `newspack_site_title()` function that supports custom site names and URLs via filters, similar to the existing `newspack_the_custom_logo()` function.
- Improved template structure in `site-branding.php` to use this new function and avoid duplication.
- Three new filters:
  - `newspack_site_title_name`: For filtering the site name that appears within the site title section.
  - `newspack_site_title_url`: For filtering the home URL that is linked by the site title.
  - `newspack_site_title`: For filtering the whole `newspack_site_title()` function at the end.

1. Set up a multibranded site configuration:
- Go to Appearance > Customize > Site Identity.
- Enable "Display Site Title".
- Verify the site title appears correctly in the header.

2. Test the custom filters:
- Add a filter for `newspack_site_title_name` to modify the site name.
- Add a filter for `newspack_site_title_url` to modify the site URL.
- Add a filter for `newspack_site_title` to modify the entire site title HTML.
- Verify each filter works as expected.

3. Test the site title display in different contexts:
- Check the site title on the homepage.
- Check the site title on the post pages.
- Verify the site title is escaped correctly and sanitized.
- Test with special characters in the site title to ensure proper escaping.

4. Test along with the [newspack-multibranded-site](https://github.com/Automattic/newspack-multibranded-site) plugin:
- Download and activate the plugin with the corresponding changes from https://github.com/Automattic/newspack-multibranded-site/pull/73.
- Create and configure a brand.
- Add pages to the brand.
- Upon visiting them, the site title URL in the header of the branded pages should point to the root of the site brand (e.g., `localhost/my-brand` instead of just `localhost/`), just like the header logo link.

### Other information:

---
  - https://app.asana.com/0/0/1210072782655693


newspack-theme/inc/template-functions.php
```diff
@@ -630,6 +630,60 @@ function newspack_the_custom_logo() {
...
+ *
+ * @return string The site title.
+ */
+function newspack_custom_site_title() {
+	/**
+	 * Filters the site name.
+	 *
+	 * @param string $blog_name The site name.
+	 */
+	$blog_name = apply_filters( 'custom_site_title_name', get_bloginfo( 'name' ) );
```
miguelpeixe, line 646:
```suggestion
	$blog_name = apply_filters( 'newspack_custom_site_title_name', get_bloginfo( 'name' ) );
```

As the other filters, we should prefix this as well

newspack-theme/inc/template-functions.php
```diff
@@ -630,6 +630,60 @@ function newspack_the_custom_logo() {
...
+	if ( empty( $blog_name ) ) {
+		return;
+	}
+
+	/**
+	 * Filters the site URL.
+	 *
+	 * @param string $site_url The site URL.
+	 */
+	$site_url = apply_filters( 'custom_site_tile_url', home_url( '/' ) );
```
miguelpeixe, line 657:
```suggestion
	$site_url = apply_filters( 'newspack_custom_site_title_url', home_url( '/' ) );
```

Same as the other, plus a small typo

newspack-theme/inc/template-functions.php
```diff
@@ -630,6 +630,60 @@ function newspack_the_custom_logo() {
...
 }
 
+/**
+ * Customize the site title.
+ * This is the one that is displayed in the header, next to the logo, when "Display Site Title" is enabled in Appearance > Customize > Site Identity.
+ * If not enabled, this site title is still rendered, but hidden via CSS.
+ *
+ * @return string The site title.
+ */
+function newspack_custom_site_title() {
```
miguelpeixe, line 640: Non-blocking suggestion: We don't have to call this a custom site title. The `newspack_` prefix already implies that it's the title determined by Newspack. It's different from `custom_logo`, which is custom because it's editable by a user (an admin uploads an image for it).

miguelpeixe (changes requested):
This is a clean solution to make it extensible and works well! Left a few comments inline.

newspack-theme/inc/template-functions.php
rbcorrales: Thanks for the suggestion. I updated the function name and removed the `custom_` part. I did the same for the filters and prefixed them all with `newspack_`.

miguelpeixe (approved):
Thank you for the changes!

PR #2536, opened 2025-08-14
fix: do not apply link colors to buttons

adekbadek:
### All Submissions:

Fixes an issue with CSS specificity for links. 

@laurelfulford – please have a look, I don't know if this won't case unintended consequences 😬 

1. On `trunk`
1. Insert a WooCommerce "Hand-picked products" block and add at least one subscription product, but ideally a list of products
1. On the front-end, observe the link text is barely or not at all visible:

image

1. Switch to this branch, observe text is visible

### Other information:


laurelfulford (changes requested):
Thanks @adekbadek! 

The downside is that this will override anyone changing the link colours in Custom CSS because it makes the selector stronger 😕 

I think it might be better to target this kind of button more specifically, since this issue doesn't affect regular button blocks. It looks like the theme styles use `.wp-block-buttons .wp-block-button__link` to make this work with the regular button block. This instance doesn't have the `.wp-block-buttons` class, but it does have `.wp-block-button` -- what do you think about replacing [this](https://github.com/Automattic/newspack-theme/blob/ad3cd4fe8a61fa3222a815681e5fb72618101fb1/newspack-theme/sass/blocks/_blocks.scss#L703C1-L709C2):

```
//! Button
.wp-block-buttons {
	--wp--style--block-gap: 0.5rem;
	.wp-block-button__link {
		color: var(--newspack-theme-color-against-secondary);
	}
}
```

with this:

```
//! Button
.wp-block-buttons {
	--wp--style--block-gap: 0.5rem;
}

.wp-block-buttons,
.wp-block-button {
	.wp-block-button__link {
		color: var(--newspack-theme-color-against-secondary);
	}
}
```

It's a little redundant, but I'm not sure if we also have cases in the wild (in our blocks or others) where we only have the `.wp-block-buttons` class and not`.wp-block-button`. 

Let me know what you think and if that doesn't make sense!

adekbadek:
@laurelfulford – much cleaner and gets the job done! Thank you 🙇‍♂️

laurelfulford (approved):
Thanks @adekbadek! Looks good to me!

PR #2578, opened 2025-10-30
fix: correct an edge case where single collections are smooshed

laurelfulford:
### All Submissions:

This PR fixes an edge case in the Collections style that I noticed while working on another bug. I've fixed it in the theme because it's caused by some `flex` styles coming from the theme, not the Collections themselves. When a Collection has posts assigned, they 'stretch' the content area, but when it doesn't, the Collection at the top is only as wide as its contents, and the image in turn is only a percentage of that.

1. Set up a new Collection, but don't assign any posts to it.
2. View the single view of the Collection on the front end - note the teeny-tiny image:

CleanShot 2025-10-30 at 13 05 48

3. Apply this PR and run `npm run build`.
4. Refresh the single Collection and confirm that all looks well.

CleanShot 2025-10-30 at 13 05 13

### Other information:


dkoo (approved):
Couldn't replicate the bug, but the fix looks fine!

laurelfulford:
Thanks Derrick!

> Couldn't replicate the bug, but the fix looks fine!

Ah weird! It could be that the Collection title was too long? ¯\_(ツ)_/¯ At least it's a "harmless" change, it just reinforces what the layout should be doing anyway, so I'll merge and hopefully stop seeing this.

PR #2581, opened 2025-11-07
fix: move some Collection margin styles from the block to theme

laurelfulford:
### All Submissions:

Please follow the testing steps in https://github.com/Automattic/newspack-plugin/pull/4293.

1.
2.
3.

### Other information:


rbcorrales (changes requested):
Hi, please look at the feedback from:
https://github.com/Automattic/newspack-plugin/pull/4293#pullrequestreview-3436782219

Also, I'm thinking that the collection styles here are a bit redundant with the styles included along with template files in the main plugin ([archive](https://github.com/Automattic/newspack-plugin/blob/trunk/src/collections/frontend/_archive.scss) and [single](https://github.com/Automattic/newspack-plugin/blob/trunk/src/collections/frontend/_single.scss)). In the long run, these could be confusing to track, especially if the styles are scattered in multiple repos. Ideally, all styles related to these pages should be kept in a single place (here or there).

laurelfulford:
Thanks @rbcorrales! I didn't clock that [the template for this archive is coming from the Newspack Plugin](https://github.com/Automattic/newspack-plugin/blob/22ad5d98e798d715928a8bd89a81153cb42a8208/includes/templates/collections/single-newspack-collection.php#L18) rather than from the theme.

I agree if the template is in the plugin then it makes sense to include the styles there, but do we have a plan for how this will work with the Block theme? The template setup - even CSS classes and markup - won't be the same between the themes. Long-term, I'd lean towards moving the current template to the Classic theme so we can better separate the styles/markup we need for the current Classic theme vs the styles/markup we need for the future Block theme (it wouldn't be the first plugin-specific template we have there -- there's some for WooCommerce and Event Calendar. In both cases the plugins have a basic template for a fallback and the theme has one that's more custom to its design).

This is only a concern when we're ready to use the Block theme with Collections, so I'll move these styles to the plugin for now!

PR #2647, opened 2026-03-09
feat: add an avatar size control for author bios

laurelfulford:
### All Submissions:

This is a belated avatar size control for the theme to fix a wall I keep running into building some designs. It changes the avatar size in both the single post footer and on the author archive pages, in addition to CSS also controlling the image size. It sets the default to 120px, but should still display as 80px on single posts.

1. Apply the PR.
2. Make sure you have at least one author with Biographical Info populated in their profile, so their bio shows on single posts.
3. View the author profile on a single post and confirm the avatar still displays as 80px (even though it's 120px under the hood).
4. Switch your site to use Newspack Sacha and repeat step 3. 
5. View the author profile on an archive page and confirm the avatar still shows as 120px.
6. Navigate to the Customizer > Author Bio Settings and confirm you now have a control to change the author avatar size with a default of 120px. 

CleanShot 2026-03-09 at 13 34 22

7. Change the size of the avatar and save changes.
8. Repeat steps 3, 4, 5 and confirm the avatars have picked up your new image size, but they still display the same size.
9. Try changing the avatar size one more time to a value under 32 or over 512; save.
10. Refresh the Customizer; confirm the saved value is either 32 (if you went under) or 512 (if you went over).

### Other information:


rbcorrales (commented):
_**EDIT: For additional context, this comment was generated by Claude automatically while we were doing a workshop.**_

## Code review

Clean implementation that follows existing Customizer patterns well. Sanitization is solid (server-side clamp + client-side `input_attrs` + defensive `absint()` at call sites). Only `newspack-sacha` overrides `author-bio.php` among child themes, so no missed files.

### Suggestions

**1. Setting controls image resolution, not display size**
The CSS in `_posts-and-pages.scss` constrains `.author-bio .avatar` to 60px / 80px, and `_archives.scss` constrains archive avatars to 30px / 80px / 120px. So regardless of the Customizer value (32 or 512), the on-screen size stays the same. The PR description and control description acknowledge this, so it's intentional, but a user changing "Author Avatar Size" to 200px and seeing no visible change could be confusing.

Consider renaming to something like "Author Avatar Image Resolution" or adjusting the description to emphasize this is about image quality/sharpness for retina screens rather than visible size.

**2. Customizer description is a bit technical for end users**
> "This controls the requested image size; final on-screen size is still controlled by CSS."

Suggestion: *"Controls the resolution of author avatars. Higher values produce sharper images on high-DPI screens but increase page load time."*

**3. Byline avatars in `template-tags.php` not covered (intentional?)**
`template-tags.php:105` uses `coauthors_get_avatar( $author, 80 )` and `template-tags.php:157` uses `get_avatar()` with no size arg (defaults to 96px). These are likely intentionally excluded since byline avatars serve a different visual purpose, but wanted to flag in case they should also be covered.

**4. Default `120` is hardcoded in 4 places**
`customizer.php`, `archive.php`, and both `author-bio.php` files each repeat `get_theme_mod( 'author_avatar_size', 120 )`. If the default ever changes, all four need updating. A constant would prevent drift, but this is minor.

### Nits

- Inconsistent comment style: `+` in templates vs `and` in `customizer.php` ("author archive + author bio" vs "author archive and author bio")

### Verdict

Looks good overall. The main question is whether the control name/description sets the right expectations for non-technical users, since it only affects image resolution, not visible size.

laurelfulford:
Thanks for the review, @rbcorrales! 😄 It's interesting to see the focus on the UI. There's some good points but it's a weird situation where this feature is kind of threading the needle of not changing how the theme looks for existing sites, while adding some more flexibility for the author avatar sizing under the hood.

I did adjust the Customizer description in [3073781](https://github.com/Automattic/newspack-theme/pull/2647/commits/3073781cc4a6f81a6fb272b6a82f18a7fd4e0a9e) to hopefully make it a little clearer that changing the customizer value had to be accompanied with a CSS tweak to make a noticable image size difference. Just let me know if that could still use help! I avoided using the term "resolution" in case that was confusing for publishers.

I also left the default 120 value in the code since we're not going to change that -- now that the field is flexible, there will be no need to mess with the default.

Just let me know if anything doesn't sense, and thanks for the demo-review of this!

rbcorrales (approved):
Clean implementation, follows existing Customizer patterns well. Sanitization is solid (server-side clamp + `absint()` at call sites + client-side `input_attrs`). Only `newspack-sacha` overrides `author-bio.php` among child themes, so no files missed. Byline avatars in `template-tags.php` are correctly excluded since those serve a different visual purpose.

One minor thing: the default `120` is repeated in 4 `get_theme_mod()` calls across the templates. A constant would prevent drift if it ever changes, but it's consistent with how `author_bio_length` is already handled, so not a big deal.

Looks good :+1:

laurelfulford:
Thanks @rbcorrales! 🎉

PR #2649, opened 2026-03-12
fix: ensure editor fonts and widths are loaded in iframed editor

laurelfulford:
### All Submissions:

In WordPress 7.0, the Gutenberg editor will always load in an iframe. This breaks our theme's font customizations, and also makes it so the body classes we need to change the editor width aren't loaded. 

This PR takes a swing at fixing that by:
* Changing how the font, colour, and editor override CSS files are enqueued using a different hook
* Uses a hacky work-around to get specific body classes added to the editor's iframe body

1. Switch your test site to the latest WP 7.0 beta/RC.
2. Navigate to the Customizer and switch up your fonts: pick a couple fonts from Google fonts that are not installed on your computer, and set them up as your font. Set the fallback to something noticeable (like monospace).
3. Change your colours to something noticeably different.
4. Create a new post and try editing it; note that your fonts aren't applied.
5. Try switching the post template to the One Column Wide template, save, and refresh. Note your blocks aren't any wider.
6. Apply this PR and run `npm run build`. 
7. Refresh the editor and confirm the font is loaded, and your template widths are applied.
8. Confirm that the colours still work - try a group block with a background and a paragraph block with a background.
9. Toggle between the different templates (and save and refresh) and confirm blocks are wider when the One Column Wide template selected.
10. Roll back to the current WordPress release.
11. Repeat the testing steps (applying fonts, colours, and different template widths), and confirm there are no issues. 

### Other information:


adekbadek (commented):
Works as described.

adekbadek (approved):
(the previous review mentioned a change request (removed now) which upon closer inspection turned out to be not needed)

PR #2655, opened 2026-03-24
feat(yoast): use shared primary category utility from newspack-plugin

rbcorrales:
### All Submissions:

Replaces inline Yoast `WPSEO_Primary_Term` logic in the classic theme with the shared `Newspack\Primary_Category` utility from newspack-plugin (see Automattic/newspack-plugin#4563).

Changes:
- `newspack_categories()` in `template-tags.php` now calls the shared utility with a `class_exists` guard, falling back to all categories when the plugin is not active
- Body class logic in `template-functions.php` uses the shared utility for the `primary-cat-*` class
- Removed the `post_primary_category` Customizer toggle (the setting now lives in Newspack > Settings > Advanced Settings)

The plugin handles backwards compatibility: sites that had the old Customizer toggle disabled will remain disabled until an admin explicitly changes it via the new settings UI, at which point the legacy theme mod is cleaned up.

Depends on Automattic/newspack-plugin#4563.

Closes NPPD-1346.

1. Activate the classic Newspack theme with Yoast SEO and newspack-plugin (with PR #4563) active.
2. Create a post with multiple categories, set one as primary via Yoast.
3. View the post. Only the primary category should display.
4. Go to Newspack > Settings > Advanced Settings. Toggle off. View the post. All categories should display.
5. Deactivate newspack-plugin entirely. View the post. All categories should display (graceful fallback).
6. Verify the Customizer no longer shows the "Use Yoast's primary category functionality" checkbox.

### Other information:


laurelfulford (approved):
Looks good! That Customizer panel is getting empty!

PR #2667, opened 2026-03-31
fix(layout): prevent horizontal scrollbar on full-bleed image positions

wil-gerken:
### All Submissions:

### Changes proposed in this Pull Request

On Chromium browsers with classic (non-overlay) scrollbars — especially on Windows, but also on macOS when **Always show scrollbars** is enabled — posts using the **Behind Headline**, **Beside Headline**, or **Above Headline** featured image positions can display a horizontal scrollbar at the bottom of the page.

The issue is caused by the full-bleed technique used on `.featured-image-behind`, `.featured-image-beside`, and `.featured-image-above`, which combines `width: 100vw` with `margin: 0 calc(50% - 50vw)`. On browsers with classic scrollbars, `100vw` includes the scrollbar width (about 15px), causing the element to become slightly wider than the visible viewport and creating horizontal overflow.

This PR fixes the issue by adding `overflow: clip` to `.site-content` for the three affected body classes. The existing `single-feature`, `single-wide`, and front page templates use `overflow: hidden` for the same purpose, but `hidden` creates a new block formatting context that breaks `position: sticky` on descendant elements (e.g. sticky sidebar ads). `overflow: clip` contains the overflow identically without affecting sticky positioning.

Closes NPPM-2670.

**Reproduce the problem:**

1. On a post using the Behind Headline, Beside Headline, or Above Headline featured image position, open Chrome DevTools Console and run:

```js
const s = document.createElement('style');
s.textContent = `html { overflow-y: scroll !important; } ::-webkit-scrollbar { width: 15px !important; height: 15px !important; } ::-webkit-scrollbar-track { background: #f1f1f1 !important; } ::-webkit-scrollbar-thumb { background: #888 !important; }`;
document.head.appendChild(s);
window.scrollTo({ left: 100 });
console.log('scrollX:', window.scrollX);
```

2. Without the fix, `scrollX` will be `7.5` and the page will have drifted slightly — confirming horizontal overflow.

**Verify the fix:**

3. With the fix applied, repeat the same steps. `scrollX` should be `0` (you'll need to shift + reload)
4. Verify the full-bleed image still fills edge to edge visually on all three positions (Behind, Beside, Above).  No layout regression.
5. Verify on a post using the `single-feature` or `single-wide` template that those layouts are unaffected.

### Other information:


laurelfulford (changes requested):
Hey @wil-gerken - thanks for tackling this! 

One small issue with this approach is that the `overflow: hidden` will prevent sticky sidebar ads from working when these featured image placements are paired with the default template (which has the widget sidebar). [Those ads use](https://github.com/Automattic/newspack-ads/blob/trunk/src/frontend/style.scss#L4) `position: sticky`, which works great until any parent element has certain overflow values.

You can kind of fake this out by using the default template, adding some sidebar widgets and adding this CSS to the Custom CSS:

```
#secondary .widget:last-child {
    position: sticky;
    top: 1rem;
}
```
For some reason, `clip` is treated differently than `hidden` for this (like `overflow: clip` and `overflow-x: clip`) -- the theme's approach predates that being widely used but it seems to be well supported now. That might be worth trying instead - if it doesn't work let me know, I'm happy to help brainstorm alternatives 😅

wil-gerken:
Thanks for catching this, @laurelfulford!

`overflow: clip` did the trick. It contains the horizontal overflow without breaking `position: sticky` on sidebar elements, so I’ve updated the PR to use it for the three featured image body classes.

Copilot also suggested wrapping it in `@supports (overflow: clip)` and falling back to `overflow: hidden` for browsers that don’t support `clip`, so I added that too.

And thanks for the handy tip on simulating a sidebar with a sticky widget. I’m definitely keeping that in my local instance.

laurelfulford (approved):
Looks good, @wil-gerken! And the fallback is smart 🙂

